PR #388, opened 2018-05-31
menu checks ~/version.txt for current version

tehn:
version.txt just contains a string ie 180531

### DIFF
--- a/lua/startup.lua
+++ b/lua/startup.lua
@@ -4,15 +4,6 @@ require 'menu'
 require 'math' 
 math.randomseed(os.time()) -- more random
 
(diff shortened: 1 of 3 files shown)


lua/startup.lua
```diff
@@ -4,6 +4,15 @@ require 'menu'
 require 'math' 
 math.randomseed(os.time()) -- more random
 
+local fd = io.open(os.getenv("HOME").."/version.txt","r")
+if fd then
+  io.input(fd)
+  norns.version.update = io.read()
+  io.close(fd)
+else
+  norns.version.update = "000000"
```
artfwo: this can be initialized in norns.lua too

tehn: will do

Issue #381, opened 2018-05-27
clean up sclang -> matron start notification sequence


catfact:
synchronization of sclang and matron on startup is currently really basic:

- on launch, `matron` repeatedly sends a "ready?" OSC message to sclang, and waits for a response. this happens before the interpreter is initialized or the main event loop starts running.

there are at least two problems that need addressing:

- [1] if sclang fails to start (e.g. because of a coding error in a SC class), then matron will hang forever. (quick solution: add a timeout)

- [2] if we want to recompile sclang from within lua (as `audio.restart()` does now), there is no way to actually get the notification that sclang has finished restarting. what we would want to do is to run `state.resume()` at that point.

right now i'm thinking that the best way to deal with both issues might actually be to restructure the ready-wait:
- send "ready?" from matron _after_ the lua interpreter and event loop are started (will require sending it on a new thread, or from lua)
- receipt of "ready" confirmation in matron should actually send an event to lua, triggering `state.resume()` and so on.

catfact:
PR 392 addresses this sufficiently for now.

PR #392, opened 2018-06-01
Refactor crone startup

catfact:
implemented changes described in https://github.com/monome/norns/issues/381

this makes `audio.restart()` lua method more functional, as a first step for editing SC classes. (`audio.restart()` should now re-trigger the main startup sequence, including resuming last script and sending audio I/O and effect parameters.)

it also means that if SC fails to boot, you get a timeout message but can continue to use the menu to turn on wifi.

### DIFF
--- a/lua/startup.lua
+++ b/lua/startup.lua
@@ -58,19 +58,10 @@ norns.startup_status.ok = function()
 end
 
 norns.startup_status.timeout = function()
-   print("norns.startup_status.timeout")
-   screen.clear()
-   screen.aa(1)
-   screen.line_width(1)
-   screen.level(10)
-   screen.move(10, 10)
-   screen.text("error: audio timed out")
-   screen.update()
-   --- FIXME: should probably do something different here,
-   --- but menu breaks if we don't try to load a script :/
-   norns.state.resume()
+  print("norns.startup_status.timeout")
+  norns.script.clear()
+  norns.scripterror("AUDIO ENGINE")
 end
-
 
 print("start_audio(): ")
 -- start the process of syncing with crone boot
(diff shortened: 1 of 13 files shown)


catfact:
did you try it with sclang crashing / not running?

also i'd like some more stressful tests of `audio.restart()` with these changes. (does it succesfully restore effects params? how about with different scripts? &c)

will come back in a couple hours to merge

lua/startup.lua
```diff
@@ -47,7 +47,31 @@ end
...
+   norns.state.resume()
+end
+
+norns.startup_status.timeout = function()
+   print("norns.startup_status.timeout")
+   screen.clear()
+   screen.aa(1)
+   screen.line_width(1)
+   screen.level(10)
+   screen.move(10, 10)
```
tehn: would be prettier with:

```screen.move(64,40)
screen.text_center("error: audio timed out")```

**Comment by tehn** [comment]
yes tested with crone off. works well.

**Comment by tehn** [comment]
ah, FYI, just read the intended functionality text. after the timeout we start in PLAY mode and errors just get thrown if you're running a script that needs engine params (ie, awake).

hitting KEY1 gets us into the menu. if audio fails we should clear out the current script and get into menu mode

File: lua/startup.lua [hunk 2]
```diff
@@ -47,7 +47,31 @@ end
...
 -- resume last loaded script
-norns.script.clear()
-norns.log.post("norns started")
-norns.state.resume()
+   norns.script.clear()
+   norns.log.post("norns started")
+   norns.state.resume()
+end
+
+norns.startup_status.timeout = function()
```
tehn: replace with

```
norns.startup_status.timeout = function()
  print("norns.startup_status.timeout")
  norns.script.clear()
  norns.scripterror("AUDIO ENGINE")
end

catfact: okey dokey  
see below
|
v

PR #394, opened 2018-06-02
menu: update method, multiple paramset slots

tehn:
## update method

- SYSTEM > UPDATE now simply primes an update. grabs file from USB disk, extracts it to ~/update
- on boot, ~/update is checked for any updates and then executes them if available. shut down upon completion

## parameter set slots

- BREAKING CHANGE: parameter sets moved to subfolders ie tehn/awake.pset
- hold KEY1 in PARAMETER menu to bring up load/save.
- ENC3 scrolls slot (default is 0)
- then use KEY2/KEY3 to save/load to slot positions.
- "slot" files saved with -nn extension ie awake-02.pset
- save/load status/fade shown, load dimmed for nonexistent files

- NEEDED: dust updates for default .pset files and loading within `init` of any script doing `params:read()`

## select nav behavior

- when navigating into ie `/tehn` and then backing out, the select position will resume from `/tehn`

### DIFF
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -787,14 +787,30 @@ end
 
 m.params = {}
 m.params.pos = 0
+m.params.n = 0
+m.params.loadable = true
 
 m.key[pPARAMS] = function(n,z)
   if menu.alt then
     if n==2 and z==1 then
-      params:read(norns.state.name..".pset")
+      if m.params.n == 0 then
+        params:read(norns.state.folder_name..".pset")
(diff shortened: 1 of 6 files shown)


tehn:
would probably be very smart to extract some of the logic from `menu.lua` and put it into `params.lua`

ie, have an extra arg to `params:read()` which is the "slot"/"bank" number.

tehn:
merged in update method to this PR

PR #397, opened 2018-06-03
revise update system

tehn:
- SYSTEM > UPDATE now simply primes an update. grabs file from USB disk, extracts it to ~/update
- on boot, ~/update is checked for any updates and then executes them if available. shut down upon completion


tehn:
gah included my other PR

Issue #358, opened 2018-05-22
provide more info on script errors


pq:
_From @pq on May 22, 2018 22:36_

currently (at least some) script errors provide very little information to the REPL (e.g., `### SCRIPT ERROR` and nothing more).  if we're trapping the error it'd be great to pass more back (in the repl or log).

a few observations:

* load failures do produce output
* "null pointers" don't

to reproduce, put this into an `init()` and play:

```lua
local a = nil
print(a.tostring())
```
result:

```shell
# script run
# script init
### SCRIPT ERROR:
```

_Copied from original issue: ngwese/maiden#82_

PR #398, opened 2018-06-04
report errors in script init

pq:
image

fixes: #358

(there's another `pcall` in `cleanup()` that could maybe get similar treatment, but since there's a `norns.try` right after maybe that's intended?)

/cc @tehn @catfact

### DIFF
--- a/lua/script.lua
+++ b/lua/script.lua
@@ -38,8 +38,9 @@ end
 Script.init = function()
   print("# script init")
   params.name = norns.state.name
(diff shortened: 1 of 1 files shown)


tehn:
if `cleanup()` similarly throws unhelpful errors we should correct that

pq:
> if cleanup() similarly throws unhelpful errors we should correct that

will pick that up in a follow-up 👍

PR #413, opened 2018-06-09
After update, encourage the user to wait for a reboot

nf:
If the user powers down the machine before the file system is synchronized,
then they may leave their norns in a broken state. Change the message at the
end of the update process so that it encourages the user to wait for a reboot.


lua/update.lua
```diff
@@ -58,9 +58,11 @@ function Update.run()
     screen.update()
     os.execute("$HOME/update/"..newest.."/update.sh")
     screen.clear()
-    screen.move(64,32)
     screen.level(10)
-    screen.text_center("complete. shutting down.")
+    screen.move(64,16)
+    screen.text_center("completing update.")
```
nf: either this, or 'update complete after reboot.' ?

tehn:
so, FYI. in the first update, the update procedure was overhauled for just this reason (confusion leading to user error).

what happens now:

- the SYSTEM > UPDATE action primes an update (copies it to ~/update and unpacks after validation)
- if confirmed, user is presented with "update available, reboot to apply"
- on next reboot (every boot checks) the version number is checked against available updates. if there's a valid update, THEN the update procedure happens (instead of the sparkles), and upon completion it auto-shuts down. it will shut down within 0.5 seconds, so upon saying "shutting down" there is really no confusing possibility of the user hard resetting IMO

but definitely the method that was necessary in the first update was not good, which is why i revised it right away.

i think the text is good as-is, i think the user should be alerted that the shutdown about to happen is intentional.

samdoshi:
Although this doesn't apply anymore, on Linux there is the `sync` command which flushes all writes to the filesystem, though it doesn't happen instantly:

> **Notes**
On Linux, sync is only guaranteed to schedule the dirty blocks for writing; it can actually take a short time before all the blocks are finally written. The reboot(8) and halt(8) commands take this into account by sleeping for a few seconds after calling sync(2).

https://linux.die.net/man/8/sync

simonvanderveldt:
@samdoshi since the new update procedure does a shutdown an explicit sync isn't necessary (on shutdown the volumes get unmounted at which point they also get synced :))

PR #419, opened 2018-06-10
Util.lua additions and fixes

markwheeler:
Three commits in here:

- The first adds some comments for documentation for the undocumented methods in `util.lua`. I wasn't sure how the docs were generated so this is just the lua file.

- The second commit is a suggestion to add a method to `util.lua` for generating note name strings from MIDI note numbers. Let me know if this should go elsewhere.

- The third requires a little more explaining...

It seemed that the existing `util.expexp()` method was returning invalid values outside of range. This method was pulled directly from [this snippet](https://github.com/supercollider/supercollider/blob/cca12ff02a774a9ea212e8883551d3565bb24a6f/lang/LangSource/MiscInlineMath.h#L49) of SC and matches. However, looking at [this other snippet](https://github.com/supercollider/supercollider/blob/a51dad9cb479503a6f031d425d8aa4d484474b34/SCClassLibrary/Common/Math/SimpleNumber.sc#L439) from SC there is a difference in the brackets. My third commit replaces that line of code with the line from SC's `SimpleNumber.sc`. I assume this is a SC bug that got pasted across but please check I'm not crazy here :)

### DIFF
--- a/lua/util.lua
+++ b/lua/util.lua
@@ -130,7 +130,7 @@ function util.expexp(slo, shi, dlo, dhi, f)
   elseif f >= shi then
     return dhi
   else
-    return math.pow(dhi/dlo, math.log(f/slo)) / (math.log(shi/slo)) * dlo
+    return math.pow(dhi/dlo, math.log(f/slo) / math.log(shi/slo)) * dlo
   end
 end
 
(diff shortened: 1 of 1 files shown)


markwheeler:
@tehn hmm just thinking about this more maybe it's better to split the note-to-name method now to avoid confusion later. any thoughts on if i put that method in a new `midiutil` (or `musicutil`?) class file? was thinking it'd be nice to have some methods for snapping to and generating scales etc too.

tehn:
@markwheeler sounds good to me!

`musicutil` or just `music` seems good since MIDI isn't always the end use

markwheeler:
I'll update this branch to remove the midi note method and get a new branch going with a `musicutil` class shortly 👍

markwheeler:
OK I pulled that method out, this PR now just contains the doc comments and fix to `expexp()`
Should be good to go!

Issue #412, opened 2018-06-09
Screen relative movements are inaccurate


markwheeler:
`screen.move_rel(0, 0)` causes things to move a little on both axes. It seems like all relative movements are slightly offset.
Similar error in `screen.line_rel()`

tehn:
@markwheeler 

indeed there's a bug: https://github.com/monome/norns/blob/master/matron/src/hardware/screen.c#L255

this command shouldn't have the 0.5,0.5 offset. and many other similar bugs can be fixed.

so the issue is the "center" of the pixel is offset by 0.5,0.5. if you use 0,0 it's the "corner" of the pixel and the color will be darker as a result of anti-aliasing an off-center coordinate.

so it seems when typing up the lib i blindly applied this rule to a bunch of stuff that wasn't supposed to have it applied. particularly the _rel commands.

markwheeler:
Ah yep I see. Looks like this applies to `rect` and `arc` also. I would expect `screen.rect(0,0,1,1)` to draw the top leftmost pixel but I have to offset by -0.5 on both axis to get it there.

Issue #410, opened 2018-06-08
Cairo line drawing is limited to integers


markwheeler:
It looks like the `_screen_line` and `_screen_line_rel` methods in `weaver.c` only allow integer inputs but I think Cairo is happy to take a double. This results in unnecessary jitter when moving/rotating lines on an anti-aliased screen. Is there a reason or could we change to a `luaL_checknumber()` there?

Issue #395, opened 2018-06-02
cairo line_width not working?


tehn:
```
screen.line_width(4)
screen.line_rel(20,20)
```

should give a fat line. it doesn't. this is broken somewhere in the C implementation.

Issue #416, opened 2018-06-09
levels screen right output vu showing left output level instead


j-flee:
i don't know if there was a specific reason for keeping the output vu meters the same, but just something very minor i noticed while panning something completely right and not seeing it at all on the levels vu bar. while investigating i noticed that in norns/lua/menu.lua (line 468) uses n = m.mix.out1 for the right vu instead of m.mix.out2.

tehn:
fixed #423

PR #423, opened 2018-06-12
various screen fixes

tehn:
fixes https://github.com/monome/norns/issues/412

fixes https://github.com/monome/norns/issues/410

fixes https://github.com/monome/norns/issues/395

also adds `screen.pixel(x,y)` (lua)

### DIFF
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -457,53 +457,43 @@ m.redraw[pMIX] = function()
   local x = -40
   screen.level(2)
   n = (60+mix:get("output"))/60*48
-  screen.rect(x+42,56,2,-n)
+  screen.rect(x+42.5,56.5,2,-n)
   screen.stroke()
 
   screen.level(15)
   n = m.mix.out1/64*48
-  screen.rect(x+48,56,2,-n)
+  screen.rect(x+48.5,56.5,2,-n)
   screen.stroke()
 
   n = m.mix.out2/64*48
-  screen.rect(x+54,56,2,-n)
+  screen.rect(x+54.5,56.5,2,-n)
   screen.stroke()
 
   screen.level(2)
   n = (60+mix:get("input"))/60*48
-  screen.rect(x+64,56,2,-n)
+  screen.rect(x+64.5,56.5,2,-n)
   screen.stroke()
 
   screen.level(15)
   n = m.mix.in1/64*48
-  screen.rect(x+70,56,2,-n)
+  screen.rect(x+70.5,56.5,2,-n)
   screen.stroke()
   n = m.mix.in2/64*48
-  screen.rect(x+76,56,2,-n)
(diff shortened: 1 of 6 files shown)


tehn:
added one more fix to menu display

fixes https://github.com/monome/norns/issues/416

tehn:
so i think i got everything. unfortunately you don't have hardware at the moment to check!

markwheeler:
Ha yep, actually I don't have the build environment setup anyway (only been doing lua and sc stuff so far).

Looking at the code the fixes look like they make sense.

However I would suggest also removing the half pixel offset from `screen_arc`, `screen_rect`, `screen_move` and `screen_line`. The offset only really makes sense if you are drawing with an odd-width stroke which seems like a confusing assumption to make. Thoughts?

tehn:
FYI the dev chain is all built in on norns. to recompile:

```
cd norns
git pull
./waf
./stop.sh
./start.sh
```

and you'll be up and running.

perhaps i'm misunderstanding cairo--- i added the offsets after some googling as a result of initial weird rendering issues.

is pixel 0,0 translated as the area between (-0.5,-0.5) and (0.5,0.5) or is it (0,0) to (1,1)?

looking for references now...

markwheeler:
I think this explains it: https://www.cairographics.org/FAQ/#sharp_lines

tehn:
thanks for the link, that was the one i read way in the past and had forgotten.

so you imagine it's best to let users figure out the offset business on their own? as the current hack isn't a real solution clearly.

markwheeler:
Oh I didn't answer your earlier question: When Cairo draws a 1px rect at '0,0' it is exactly filling the first pixel, which could be described as the area from 0-1 on a graph.

Yes I think if we stick to the default cairo way of doing things we'll at least be in alignment with how most screen drawing systems work. And yep I'm sure some folks will draw 1 pixel lines and wonder why they're blurry but at least when they end up at that link or similar it'll get them on track!

Of course this change will throw off some of the current UI, we'll have to do a pass to see what needs tweaking.

tehn:
checking the UI now. `rect` is really off with integers using `stroke` so i'm offsetting the LEVELS page with the bars. other scripts with normals lines seem fine ie awake and playfair

markwheeler:
Hmm I would have expected any on-pixel line drawing to now need to be shifted on the script side (eg, in awake) in order to stay sharp. Think I need to test this on device! :D

tehn:
> Hmm I would have expected any on-pixel line drawing to now need to be shifted on the script side (eg, in awake) in order to stay sharp. Think I need to test this on device! :D

i expected it as well, but playfair and awake look identical to as before.

tehn:
scratch that... _almost_ identical. there's some off-by-one pixel tuning now but it's minor

tehn:
i think this is good to merge if you (or anyone) would like to approve the PR

PR #424, opened 2018-06-13
arc support

scanner-darkly:
arc support. i mostly copied the grid implementation, trying to re-use monome naming/functions were possible. i did have to introduce arc specific methods for LED rendering as it's done differently for arcs.

there is a couple of issues with this implementation:
- it will return 0 for number of encoders - from what i can tell libmonome doesn't currently support it, i'll add it once i have a chance to add it to libmonome
- it breaks (error messages in REPL suggest some memory corruption, and then at some point it hangs) if both grid and arc are connected and LEDs are set on both (or if refresh is called). they seem to work separately fine, or if you only update one or the other, so i think it should still be okay to merge this, and i'll fix it in a separate PR.

didn't want to create a global variable `a` so used `arc` instead.

### DIFF
--- a/lua/monome.lua
+++ b/lua/monome.lua
@@ -1,0 +1,23 @@
+-- monome device manager
+
+norns.monome = {}
+
+
+norns.monome.add = function(id, serial, name, dev)
+  -- TODO is there a better way to determine?
+  if (string.find(name, "arc") == nil) then
+    -- print("grid detected")
+    norns.grid.add(id,serial,name,dev)
+  else
+    -- print("arc detected")
+    norns.arc.add(id,serial,name,dev)
+  end
+end
+
+
+norns.monome.remove = function(id)
+  -- call both since we don't know if it's a grid or an arc
+  -- grid/arc functions will check for existence
+  norns.grid.remove(id)
+  norns.arc.remove(id)
+end
--- a/matron/src/device/device_monome.c
+++ b/matron/src/device/device_monome.c
@@ -18,9 +18,7 @@
 //-- static functions
 static void dev_monome_handle_press(const monome_event_t *e, void *p);
 static void dev_monome_handle_lift(const monome_event_t *e, void *p);
-
-static void dev_monome_handle_press(const monome_event_t *e, void *p);
-static void dev_monome_handle_lift(const monome_event_t *e, void *p);
+static void dev_arc_handle_enc(const monome_event_t *e, void *p);
 
 //-------------------------
 //--- monome device class
@@ -44,6 +42,7 @@ int dev_monome_init(void *self) {
 
     monome_register_handler(m, MONOME_BUTTON_DOWN, dev_monome_handle_press, md);
     monome_register_handler(m, MONOME_BUTTON_UP, dev_monome_handle_lift, md);
+    monome_register_handler(m, MONOME_ENCODER_DELTA, dev_arc_handle_enc, md);
 
     // drop the name set by udev and use libmonome-provided name
     free(base->name);
@@ -87,6 +86,22 @@ void dev_monome_all_led(struct dev_monome *md, uint8_t val) {
     }
 }
 
+// set a given arc LED value
+void dev_arc_set_led(struct dev_monome *md,
+                        uint8_t enc, uint8_t led, uint8_t val) {
+    /* fprintf(stderr, "dev_monome_set_led: %d %s %d %d %d\n", */
+    /*             md->dev.id, md->dev.serial, enc, led, val); */
+    md->data[enc][led] = val;
+    md->dirty[enc] = true;
+    md->arc = true;
+}
+
+// set all arc LEDs to value
+void dev_arc_all_led(struct dev_monome *md, uint8_t val) {
+    dev_monome_all_led(md, val);
+    md->arc = true;
+}
+
 // transmit all dirty quads
 void dev_monome_refresh(struct dev_monome *md) {
     static const int quad_xoff[4] = {0, 8, 0, 8};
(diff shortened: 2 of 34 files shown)


artfwo:
the global arc variable could be entirely dropped in this case (implying that arc add/remove is handled by the script).

matron/src/device/device_monome.c
```diff
@@ -137,6 +174,11 @@ int dev_monome_grid_cols(struct dev_monome *md) {
     return monome_get_cols(md->m);
 }
 
+int dev_monome_arc_encs(struct dev_monome *md) {
+    // FIXME returning rows, libmonome currently doesn't support returning encoders
+    return monome_get_rows(md->m);
```
artfwo: if this is broken, consider returning a constant or removing the method completely.

matron/src/device/device_monome.c
```diff
@@ -87,17 +86,39 @@ void dev_monome_all_led(struct dev_monome *md, uint8_t val) {
...
+    /*             md->dev.id, md->dev.serial, enc, led, val); */
+    md->data[enc][led] = val;
+    md->dirty[enc] = true;
+    md->arc = true;
+}
+
+// set all arc LEDs to value
+void dev_arc_all_led(struct dev_monome *md, uint8_t val) {
+    dev_monome_all_led(md, val);
+    md->arc = true;
```
artfwo: this flag shouldn't be set here or in the above method. i'd suggest removing it completely and using rows/cols for checking that we're dealing with an arc in `dev_monome_refresh` (arcs have 0x0 size in libmonome)

lua/monome.lua
```diff
@@ -0,0 +1,23 @@
...
+  else
+    -- print("arc detected")
+    norns.arc.add(id,serial,name,dev)
+  end
+end
+
+
+norns.monome.remove = function(id)
+  -- call both since we don't know if it's a grid or an arc
+  -- grid/arc functions will check for existence
```
artfwo: makes sense to check for monome device type here and call the appropriate remove handler (and remove the existence check from the handler then).

matron/src/device/device_monome.c
artfwo: as a hack, scripts can get the number of encoders from the device name (which can be done in lua).

scanner-darkly: that would be easier - is it considered a reliable enough way? not sure if it's different for DIY'd arcs. (the reason i didn't use a constant was because it would give me a compiler error about unused `md` variable, since warnings are treated as errors)

matron/src/device/device_monome.c
scanner-darkly: i'm considering implementing separate arc methods instead of re-using `dev_monome_refresh`/`dev_monome_init` etc. it wouldn't create much code duplication and it would be cleaner (and i have a feeling this would help with the issue of grid and arc not working together).

lua/monome.lua
scanner-darkly: same as above, i think i'll add `norns.arc.add` / `norns.arc.remove` instead - any objections?

scanner-darkly:
"the global arc variable could be entirely dropped in this case (implying that arc add/remove is handled by the script)." - could you elaborate on this a bit?

matron/src/device/device_monome.c
artfwo: right, just use `(void) md;` in the beginning of the function to avoid that warning. that's what we currently do to override that.

artfwo:
take a look at the midi module, which hasn't got a global "m" var, like "g" for grid. the reason for midi not having a "default" variable is that there are use-cases for several devices and device-specific mappings in scripts. midi scripts (such as earthsea) then define handlers for midi.add and midi.remove to setup the controller.

since arc is much more rare and specific than the grid (and often used as secondary controller), it could also have the requirement to explicitly define add/remove handlers in a script, eliminating the need for global default "arc" variable.

matron/src/device/device_monome.c
artfwo: this will also make things a bit messier. i believe the issue with grid/arc not working together could be solved with the current implementation. could you run gdb to see what leads to the memory errors?

matron/src/device/device_monome.c
scanner-darkly: cool, good to know!
going to remove this and move the implementation to lua as you suggested.

matron/src/device/device_monome.c
scanner-darkly: wouldn't it be cleaner with separate methods?

`void dev_arc_all_led(...)`
`void dev_grid_all_led(...)`
vs
```
void dev_monome_all_led(...)
...
if (is_arc)
    // arc implementation
else
  // grid implementation
```
they are pretty light weight wrappers, so there will be hardly any code duplication (and if there is it could be extracted into generic monome functions which both arc and grid functions would call).

regardless yeah, i'll investigate the problem some more, i have a feeling we might run into this with multiple grids as well. thanks for the suggestion, will give it a try!

scanner-darkly:
makes sense. 
one thing to consider for having a global variable is `script.lua` turns off all LEDs when doing a script cleanup, which is a good thing. is there a way to achieve the same without a global var?

artfwo:
yeah, send a led_all(0) on `arc.add` that will be called by reconnect in turn. same could be done for the grid.

scanner-darkly:
sounds good, will remove the global var.

scanner-darkly:
ran into several issues with this update. somehow arc:refresh() would cause issues even though it worked before (i had a script responding to encoders and displaying current values with LEDs). while investigating i managed to brick norns with an unclosed print statement, had to reflash the latest update and after that arc:refresh works, but somehow arcenc function is not getting called (which worked before). pretty baffled at this point and feel like i'm hitting the limits of my [very limited] knowledge of lua. 

i'm going to treat this as a universe way of saying i should finish teletype first - going to close this PR and come back to this when i have the time to properly investigate and refactor.

scanner-darkly:
also should mention - it's almost as if failures somehow accumulated between stop.sh/start.sh cycles - is there a "cache" of some sort that doesn't get flushed by stop/start cycle?

catfact:
i guess there could be some issue within `libmonome`, but not within lua or matron, no

Issue #425, opened 2018-06-13
stop the globalocalypse


catfact:
there are way too many globals and they aren't defined in any particular place. it's a recipe for disaster. not even just in user scripts but when we want to extend or maintain the system.

in the beginning, there was the `norns` table, and it was global. it had to be, so that the C API glue could see it and call its fields. 

then, there were more globals added for convenient usage in scripts and REPL. this was controvesial, but at least they were collected in one place (`globals.lua`) and commented with `ldoc.` 

now, i don't know what happened to `globals.lua` but the definition of globals has gone everywhere- many are in `startup.lua`, which makes sense:
https://github.com/monome/norns/blob/master/lua/startup.lua#L9-L31

but many are in `script.lua` (i think? can't tell! that's the whole problem in a nutshell)
https://github.com/monome/norns/blob/master/lua/startup.lua#L9-L31

and probably elsewhere. this is really bonkers because `script.lua` is otherwise structured as a module!

pq:
yeah, i see a bunch more.

typing 

```for k,v in pairs(_G) do print(k..' : '..tostring(v)) end```

in the REPL (on a fresh start), yields this list of suspects:

```
audio_output_level : function: 0x1b104
s_line_rel : function: 0x1b984
grid_rows : function: 0x1bdb0
audio_pitch_on : function: 0x1a514
_G : table: 0x53938
home_dir : /home/we
load_engine : function: 0x1b3dc
redraw : function: 0x12cf70
poll : table: 0x647b0
s_arc : function: 0x1b74c
set_aux_fx_input_level : function: 0x1afb8
s_level : function: 0x1bbdc
io : table: 0x55418
audio_monitor_level : function: 0x1b0b4
set_aux_fx_off : function: 0x1a484
gain_in : function: 0x1b534
s_move : function: 0x1bb14
request_poll_value : function: 0x1b1bc
bit32 : table: 0x54918
s_line_width : function: 0x1bb90
s_aa : function: 0x1bc28
table : table: 0x54f60
osc_send : function: 0x1aa24
set_insert_fx_on : function: 0x1a474
utf8 : table: 0x574d8
audio_input_level : function: 0x1b154
set_insert_fx_off : function: 0x1a464
grid_all_led : function: 0x1be74
gain_hp : function: 0x1b59c
rawget : function: 0x76c0be04
metro_stop : function: 0x1b38c
script_dir : /home/we/dust/scripts/
loadfile : function: 0x76c0c674
[...]

Issue #258, opened 2018-04-28
sandboxed environment for lua scripts and REPL


catfact:
i think we can and should use a dedicated table as the pseudo-global environment for executing the current script and REPL chunks. that way users can use globals without fear.

https://stackoverflow.com/questions/35910099/how-special-is-the-global-variable-g
http://lua-users.org/wiki/EnvironmentsTutorial
http://www.lua.org/manual/5.3/manual.html#pdf-_G

pq:
more food for thought: if i understand the implications, i think sandboxing could also make development of scripts with module dependencies easier[*] and lessen the need for some kind of module re-load facility in maiden (https://github.com/monome/maiden/issues/22)


[*] tl;dr: the rub is that since modules are not automatically reloaded, changes to them do not get picked up in dependent scripts/modules leading to potential confusion when making changes.

ngwese:
for user modules (those in dust) one possible way to handle this is to monkey patch the require function in the sandboxed environment allowing us to keep track of modules loaded explicitly by a user script... once we know what was loaded we can certainly clear out the package.loaded global as needed. 

system modules (part of norns) are still problematic but given that they are loaded in multiple places potentially. 

the evaluation environment for the repl commands might be the more challenging aspect since the lifetime of the repl spans multiple scripts

tehn:
primary fix https://github.com/monome/norns/pull/436

closing this, but referencing #557 #559

PR #426, opened 2018-06-13
script loading experiments (experimental)

pq:
**(experimental)**

replacing `dofile`  with `loadfile` allows us to pass an env into which script globals get loaded.  this experiment introduces a simple proxy env that is essentially a pass-through to `_G`.  minimally, this proxy gives us hooks to monitor or police global access and (re)definition.  taking it further we could consider expanding on this approach in the service of sandboxing #258.

anyway, this is more modest than that.  as a drop-in replacement for our current `dofile` approach it should “just work” and gives us the opportunity to at least report un-toward re-writes of globals (see: #425).

_note: i’m not proposing this land as is but rather see it as food for thought.  hoping it moves the conversation along._

/cc @catfact @ngwese @tehn

### DIFF
--- a/lua/script.lua
+++ b/lua/script.lua
@@ -44,6 +44,40 @@ Script.init = function()
   end
 end
 
+local function do_load(file)
+  -- print(">> loading from... "..file)
+  
+  -- carve out some space in the global env to track our state
+  _G.__norns_init_state = { }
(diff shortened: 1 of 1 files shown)


tehn:
#436 instead

PR #427, opened 2018-06-14
reset gridkey on clear

pq:
reset `gridkey` on script clear.

/cc @tehn @catfact


lua/script.lua
```diff
@@ -21,6 +21,8 @@ Script.clear = function()
     g.key = norns.none
   end
   g = nil
```
pq: also a question: why is g getting clobbered here?  one side-effect is that it's not available in the next `init()`.

PR #432, opened 2018-06-21
musicutil.lua

markwheeler:
Suggestion for a new util class for working with notes and scales.
Includes methods for generating scales, snapping notes to an array, converting note numbers to MIDI note names, converting note numbers to frequencies and vice versa.

Thoughts very welcome!

Question: I added ldoc comments but wasn't sure how to generate the actual docs and check if they were all displaying as expected.


lua/musicutil.lua
```diff
@@ -0,0 +1,186 @@
...
+-- @return Array of MIDI note numbers
+function MusicUtil.generate_scale(root_num, scale_type, octaves)
+  if type(root_num) ~= "number" or root_num < 0 or root_num > 127 then return nil end
+  scale_type = scale_type or 1
+  octaves = octaves or 1
+
+  -- Lookup by name
+  if type(scale_type) == "string" then
+    scale_type = string.lower(scale_type)
+    for i = 1, #MusicUtil.SCALES do
```
pq: i'm not positive but at a quick glance it seems you could make this a little simpler using the `pairs` function.  something like:

```
for _,scale in pairs(MusicUtil.SCALES) do 
  if string.lower(scale.name) == scale_type then
    ...
```

pq: (not at an interpreter now so can't try it, forgive me if i'm only muddying the waters!)

lua/musicutil.lua
```diff
@@ -0,0 +1,186 @@
...
+  if type(freqs) == "number" then freqs = {freqs} end
+  local output = {}
+  for i = 1, #freqs do
+    output[i] = util.clamp(math.floor(12 * math.log(freqs[i] / 440.0) / math.log(2) + 69.5), 0, 127)
+  end
+  if #output == 1 then return output[1]
+  else return output end
+end
+
+return MusicUtil
```
pq: add a trailing newline?

lua/musicutil.lua
markwheeler: From [what I've read](https://stackoverflow.com/questions/8955085/should-i-use-ipairs-or-a-for-loop) there is a performance hit for using `pairs`or `ipairs` so I've been tending to avoid them when there's a sequential index. Open to switching though.

pq: ah! interesting. would be fun to do some benchmarking. doesn’t hurt to play it safe though.

markwheeler:
Re: here or dust, or which folder, just let me know where makes the most sense 👍

lua/musicutil.lua
catfact: if i understand correctly, `#` operator has the same performance hit (requires iterating over the whole array to determine length, goes til it hits `nil`, is O(n) on table size)

markwheeler: Looking at this again I think there are some small changes I should make to avoid using `#` multiple times on the same table in a single function. Will do those changes and stick with `#` rather than `pairs`.

tehn:
this looks fantastic, thank you!

i may actually be in favor of hosting this in `dust` given it's not central to the core functionality of the menu system.

markwheeler:
Cool I'll close this PR and open a new one on dust shortly 👍

PR #434, opened 2018-06-23
Support spaces in Wi-Fi network SSIDs and passwords

Dewb:
I'm one of those unlucky people whose home wifi network has a space both in the network SSID and the password. This PR contains the following changes, which allow me to connect:
* Adds a space character to `textentry.lua`. The space is rendered as a space; a good future enhancement might be to display some kind of dim _ character to make selecting and entering it clearer.
* Quotes the network SSID and password arguments when calling `wifi.sh select`.

### DIFF
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -1032,7 +1032,7 @@ end
 
 m.wifi.passdone = function(txt)
   if txt ~= nil then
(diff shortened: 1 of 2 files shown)


Dewb:
Area of potential concern: I haven't put any thought yet into the implications of allowing spaces in all other uses of textentry.

tehn:
thank you!

PR #439, opened 2018-06-26
enable device unplug callbacks

antonhornquist:
i think this is needed for https://github.com/monome/norns/blob/master/lua/grid.lua#L27 and https://github.com/monome/norns/blob/master/lua/midi.lua#L22 to work?


pq:
this looks right to me and it sure feels like those missing calls are an oversight.

maybe @tehn or @catfact can confirm?

PR #441, opened 2018-06-27
reset midi callbacks on script clear

pq:
follow-up from https://github.com/monome/dust/pull/131#discussion_r198365764.

to be clear, i have no midi devices handy but based on a related fix for `gridkey` (https://github.com/monome/norns/commit/2e6762b409db0186156772137d1462734fc9ccd6) i'm thinking this is probably right.  would be awesome if someone could confirm. 👍 

/cc @artfwo @tehn

### DIFF
--- a/lua/script.lua
+++ b/lua/script.lua
@@ -23,6 +23,9 @@ Script.clear = function()
   g = nil
   -- reset gridkey callback
   gridkey = norns.none
+  -- reset midi callbacks
+  midi.add = norns.none
+  midi.remove = norns.none
   -- stop all timers
   metro.free_all()
   -- stop all polls and clear callbacks


antonhornquist:
lgtm.

btw, shouldn't we do this in norns core too,  so that scripts do not need to cleanup explicitly: https://github.com/monome/dust/blob/master/scripts/tehn/earthsea.lua#L383-L385

pq:
@antonhornquist: yeah, we probably should!

@artfwo: it looks like you added that bit to `earthsea`.  what do you think?

artfwo:
@pq this looks awkward, perhaps good enough for now, but not in the long term. as mentioned elsewhere, callback registration should fix these problems, is anybody working on it?

tehn:
this works.

and yes we need to figure out callback registration. not yet working on it myself... 13 more units of last shipment to build/ship :)

PR #442, opened 2018-06-28
Reset lvm

catfact:
this seems to work at first blush, but i haven't tried on hardware (only laptop) - give it a spin...

aside: seems we've gone ahead and added a preceding underscore to this function. if we want to make that a convention, let's make it consistent yeah?

meta-aside: normally i would make the feature branch on a fork. but with multiple contirbutors on the same feature maybe this is more convenient (upstream feature branch)


tehn:
rad rad rad, works!

PR #448, opened 2018-07-04
implement LVM reset

tehn:
refined the menu/script system to reset the LVM on new script loading. simplified the loading procedure, now focussed in `script.lua`

one bug i could use help with: the device_monitor needs some sort of facility to re-report currently connected devices (ie grids, midi) as the LVM forgot what's available. my naive attempts failed ie `device_monitor_scan()` after `_reset_lvm()`

otherwise working well, i need to test further but the code is much more followable given there isn't so much cleanup.


lua/engine.lua
```diff
@@ -84,10 +84,8 @@ Engine.load = function(name, callback)
   if type(callback) == 'function' then
     norns.report.did_engine_load = function()
       --print("Engine: norns.report.did_engine_load callback")
-      local status = norns.try(callback,"init")
-      norns.init_done(status)
+      callback()
```
pq: curious why not `norns.try` here?

lua/menu.lua
```diff
@@ -262,7 +249,7 @@ norns.key = function(n, z)
       t:start()
     elseif z == 0 and pending == true then
       menu.alt = false
-      if menu.mode == true and menu.scripterror == false then
+      if menu.mode == true and norns.script.errorstate == false then
```
pq: maybe:

`if menu.mode and not norns.script.errorstate then`

?

lua/script.lua
```diff
@@ -3,106 +3,69 @@
...
-  init = norns.none
-  -- clear last run
-  norns.state.script = ''
-  norns.state.name = 'none'
-  -- clear params
-  params:clear()
-end
+Script.file = ''
+Script.name = 'none' 
+Script.errorstate = false
```
pq: a thought.

if `errormsg` were set to `nil` when not in an error state, you could drop this flag and just check for `nil` instead.

not sure that's better though| 🤔

lua/script.lua
```diff
@@ -3,106 +3,69 @@
...
-      norns.state.folder_name = string.gsub(filename,'.lua','') -- store name
-      norns.state.name = norns.state.folder_name:match("[^/]*$") -- strip path from name
-      norns.state.save() -- remember this script for next launch
-      norns.script.redraw = redraw -- store redraw function for context switching
-      redraw = norns.none -- block redraw until Script.init
-      Script.run() -- load engine then run script-specified init function
-    end
+    Script.file = filename
   end
+  if pcall(cleanup) then print("# cleanup")
```
pq: maybe `norns.try` or something similar here?  notice that `norns.try` uses `xpcall` which gives us better error tracing:

https://github.com/monome/norns/blob/4e908e3edeb1ce5f49efc6d072aadbc0dd6fe1d6/lua/norns.lua#L160

lua/script.lua
```diff
@@ -121,5 +84,52 @@ Script.metadata = function(filename)
...
+
+-- read system.lua and set parameters back to stored vals
+function Script.resume()
+  local f = io.open(data_dir..'system.lua')
+  if f ~= nil then
+    io.close(f)
+    dofile(data_dir..'system.lua')
+  end
+
+  if Script.file== nil or Script.file== '' then
```
pq: (nit) add some spaces:

    if Script.file == nil or Script.file == '' then

also, we should really have a `string.isEmpty(..)` function somewhere?

lua/menu.lua
```diff
@@ -223,28 +220,18 @@ norns.menu.get_enc = function() return menu.penc end
...
-
+-- init menu after script load
 norns.init_done = function(status)
   menu.set_page(pHOME)
-  if status == true then
-    menu.scripterror = false
+  if norns.script.errorstate == true then
     m.params.pos = 0
-    menu.set_mode(false)
+    menu.mode = true
```
pq: aside: maybe a more descriptive name?  from a quick scan, this looks odd next to `menu.set_mode(menu.mode)`.  what does setting mode to `true` mean?

lua/engine.lua
tehn: because the init function (`Script.init`) called higher up has its own `try`

tehn:
closing this but there were some non-lvm-reset related optimizations that i'll carry over later.

PR #451, opened 2018-07-05
save norns state on sleep

### LINKED ISSUE #369: menu > sleep needs to perform more cleanup
- [ ] needs to write paramsets to disk
- [ ] some engines may need cleanup 

this is recreatable with `hello_ack`

tehn:
partial fix #369


tehn:
added tape stop @antonhornquist

PR #453, opened 2018-07-06
PARAMETER menu midi cc mapping

tehn:
- allows mapping from any midi device's cc inputs.
- there is no differentiation between devices (TODO, with the general device manager)
- "pmap" files are read/write for persistence, per script, like a "pset" file
- if a cc is assigned to multiple params it will only affect one, the first indexed
- only CONTROL type params mappable (because they have `set_raw` for scaling)

howto:

- with midi controller connected
- load script (ie earthsea)
- go to PARAMETERS
- hold KEY1 (alt menu)
- ENC2 scroll down to `midi-cc mapping`, use ENC3 to enable
- you're now back on the param list, but showing cc assignment numbers.
- use ENC3 to change cc number _or_
- push KEY3 to toggle LEARN, whereupon you can tweak the midi cc to auto-assign

### DIFF
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -515,7 +515,7 @@ m.redraw[pMIX] = function()
       screen.text_center("RECORDING")
       screen.move(90,48)
       local min = math.floor(tape.time / 60)
-      local sec = tape.time % 60 
+      local sec = tape.time % 60
       screen.text_center(string.format("%02d:%02d",min,sec))
     end
   elseif tape.mode == tPLAY then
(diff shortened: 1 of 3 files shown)


tehn:
i'm going to merge this as i'm sure nobody wants to actually look at the swamp which is `menu.lua`

lua/menu.lua
```diff
@@ -779,33 +781,44 @@ m.params = {}
...
-      else
-        params:write(norns.state.folder_name.."-"..string.format("%02d",m.params.n)..".pset") 
+    if n==3 and z==1 then
+      if m.params.altpos == 1 then
+        if m.params.n == 0 then
+          params:read(norns.state.folder_name..".pset")
+        else
+          params:read(norns.state.folder_name.."-"..string.format("%02d",m.params.n)..".pset")
+        end
+        m.params.action = 15
```
pq: aside: perhaps consider some constants for these magic numbers?

pq: makes more sense.  thanks!

PR #464, opened 2018-07-07
[experimental] supernova

catfact:
working / good news:

- supernova seems to build and run ok on the norns.

- anecdotally, it seems a little more efficient overall and also better at handling time desync and xruns. 

needs work / bad news:

- supernova doesn't automagically create  multiple DSP threads. to perform parallel execution, crone / engines need to create a `ParGroup` and add synths to it. i'll open a separate PR on `dust` showing the appropriate change to `Engine_PolySub`.

- currently in this PR, include paths for `nova-tt` and `nova-simd` are really stupid (see note.) i wonder if we should add a package for supernova sources, add them as submodule to norns repo, or something else. (they are required to build ugens.)


wscript
```diff
@@ -54,6 +54,8 @@ def configure(conf):
             '/usr/include/SuperCollider/common',
             '/usr/local/include/SuperCollider/plugin_interface',
             '/usr/local/include/SuperCollider/common',
+	    '/home/we/supercollider/external_libraries/nova-simd',
```
catfact: this include path is stupid.

sc/wscript
```diff
@@ -6,12 +6,13 @@ def build_ugen(bld, target, source):
         target=target,
         use=['SUPERCOLLIDER'],
         install_path='${PREFIX}/lib/SuperCollider/plugins',
-        cflags=['-O3', '-Wall'])
+        cflags=['-O3', '-Wall'],
+	cxxflags=['-DSUPERNOVA', '-DNOVA_SIMD'])
```
artfwo: woops, looks like a TAB sneaked in here

wscript
artfwo: ugh, supercollider doesn't install these files indeed. i think submodules will work great here, these libs are header-only.

artfwo:
build_ugen() could have another function defined in wscript, e.g. `build_supernova_ugen()` to build both scsynth and supernova plugins, unless we explicitly want to use only supernova.

artfwo:
it also looks like `nova_tt` is not required to build ugens.

simonvanderveldt:
Is there no `-dev` package for supercollider/supernova?

catfact:
@simonvanderveldt  nope...

i'm going to go ahead and merge this if there are no objections, along with 
https://github.com/monome/dust/pull/171

simonvanderveldt:
While submodules can be useful/convenient, it does mean a change in how we've worked so far and an additional step the users will need to do when updating (or we'll need to do if we start doing git based updates in maiden).
So, I'm not sure if it's a good move. @nf's suggestion also gets us the headers without these downsides (but at the downside of there being more stuff in this repo).

catfact:
@simonvanderveldt @nf @artfwo i removed the submodules and added supernova headers manually. i this ok? (supernova headers haven't changed in a year, so seems safe enough)

tehn:
seems good to me, thank you for adding them

nf:
This looks good. One q: what's `sc/external_libraries/nova-simd/testsuite`? Looks like unit tests that we don't need.

Issue #376, opened 2018-05-25
disable acceleration (or reduce it) within menu


tehn:
way too twitchy

nf:
I'm happy to take a crack at fixing this.

pq:
iirc should make for an easy tweak in `menu.lua`.

https://github.com/monome/norns/blob/7c553e4f0f8dcb2fd35b031f1db43aa07826b5ee/lua/menu.lua#L305-L308

nf:
Thanks @pq, that was the spot.

I disabled acceleration and it did the trick for me. I'm sure further tweaks should be made, but this is a step in the right direction.

antonhornquist:
@nf fix is mot merged

tehn:
https://github.com/monome/norns/pull/480

tehn:
acceleration should be enabled for LEVELS screen (so all faders will react the same)

PR #467, opened 2018-07-10
Disable acceleration within menu

nf:
This is a definite improvement, if not a fine solution.
Tested navigating through very long and very short menus alike,
and it seems good.

Issue #376

### DIFF
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -302,7 +302,7 @@ menu.set_mode = function(mode)
     screen.line_width(1)
     menu.set_page(menu.page)
     norns.encoders.callback = menu.enc
-    norns.encoders.set_accel(0,true)
+    norns.encoders.set_accel(0,false)
     norns.encoders.set_sens(1,1)
     norns.encoders.set_sens(2,0.5)
     norns.encoders.set_sens(3,0.5)


antonhornquist:
an alternative for the long-parameters-list use case is to add some kind of grouping of parameters and the possibility to jump to the next / previous group, and have acceleration disabled

antonhornquist:
come to think of it, separators imply grouping, so that could probably be used for jumps

antonhornquist:
merg this? @tehn @catfact

catfact:
sure

antonhornquist:
i tried this fix out again, more thoroughly. noted that this not only disables acceleration for enc2 == menu-navigation-cursor-encoder (as i first thought) but all encoders thus affecting main volume knob and parameter value changes.

so with this parameters would be changed with acceleration in script mode and without in menu, which is inconsistent.

afaict the idea of #376 is to address twitchy-ness in navigating the menu cursor?

i suggest to close this pr and keep issue #376 open

nf:
@antonhornquist if you have suggestions for improving a pull request you needn't close it; that just prevents me from fixing it.

I modified the change to only affect encoder 2 and sent #480.

PR #478, opened 2018-08-01
screen.pixel fix

markwheeler:
Looks like screen.pixel wasn't drawing a single pixel, this fixes that.
I couldn't find any scripts currently using this function but if there are some they will need tweaking.

### DIFF
--- a/lua/screen.lua
+++ b/lua/screen.lua
@@ -174,11 +174,11 @@ Screen.font_face = function(index) s_font_face(index) end
 -- @param size in pixel height
 Screen.font_size = function(size) s_font_size(size) end
 
--- draw single pixel (requires integer x/y, line_width == 1, stroke afterwards)
+--- draw single pixel (requires integer x/y, fill afterwards)
 -- @param x position
 -- @param y position
-Screen.pixel = function(x,y)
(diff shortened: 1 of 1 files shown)


lua/screen.lua
```diff
@@ -174,11 +174,11 @@ Screen.font_face = function(index) s_font_face(index) end
...
 Screen.font_size = function(size) s_font_size(size) end
 
--- draw single pixel (requires integer x/y, line_width == 1, stroke afterwards)
+--- draw single pixel (requires integer x/y, fill afterwards)
 -- @param x position
 -- @param y position
-Screen.pixel = function(x,y)
-  s_rect(x-0.25,y-0.25,0.5,0.5)
+Screen.pixel = function(x, y)
+  s_rect(x, y, 1, 1)
```
catfact: haven't tried this but if x,y really need to be int maybe it would be appropriate to "cast" them"?
`s_rect(int(x), int(y), 1, 1)`

markwheeler: Well actually they can be floats you just won't get a sharp pixel (might be useful for animation though). I can remove that comment if it's confusing?

PR #480, opened 2018-08-06
Disable acceleration within menu

nf:
This is a revised version of #467 that only disables acceleration for encoder 2.

Tested navigating through very long and very short menus alike, and it seems good.

Issue #376


catfact:
@tehn @nf 

i'll go ahead and merge this. my only hesitation was an assumption that @tehn had something less straightforward in mind, based on the fact that this was ever a GH issue in the first palce.

nf:
Thanks. We can always change/improve it later if it's not sufficient. :-)

PR #500, opened 2018-08-15
fix menu with no script loaded, remove default script, optimize

tehn:
minor fixes and optimization

- prevent error on trying to exit PARAMs
- prevent entering PLAY mode with no script loaded

these are boring micro-fixes so i'm going to admin-merge


nf:
Looks good.

PR #526, opened 2018-08-30
ParamSet:add now uses table argument

antonhornquist:
small addition to make it possible to specify action directly in `params:add`

### DIFF
--- a/lua/paramset.lua
+++ b/lua/paramset.lua
@@ -37,35 +37,39 @@ function ParamSet:add_separator()
 end
 
 --- add generic parameter
-function ParamSet:add(param)
+function ParamSet:add(args)
+  local param = args.param -- param is mandatory
   table.insert(self.params, param)
   self.count = self.count + 1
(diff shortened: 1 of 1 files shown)


antonhornquist:
this is about the under-the-hood function `add()`

was:

```
paramset.add(control.new("filter", cs.HZ, formatter))
paramset.set_action("filter", some_function)
```

proposed:

```
paramset.add{param=control.new("filter", cs.HZ, formatter), action=some_function}
```

background: i'm writing scripts that reuse a `Control` instance outside of `ParamSet`, hence my use of `add()`.

`add_control()`, `add_number()`, etc functions are not affected (deliberately to retain backwards compatibility), but imo we could consider adding an `action` argument to those to not have to repeat code as much. Ie.:

```
paramset.add_control("filter",cs.HZ)
paramset.set_action("filter", some_function)
```

```
paramset.add_control("filter",cs.HZ, some_function)
```

Issue #527, opened 2018-08-30
device monitor - unnecessary check on usb subsystem for midi


TheTechnobear:
because norns uses alsa raw open (snd_rawmidi_open) the dependency on monitor is to /dev/snd/midiCxDx.

there is no requirement in monitor to check that this is a usb subsystem for alsa to work.
so, we might as well check against this path in scan as  we are already doing check already in device_monitor.c in other places

I accept , that the Norns hardware is only extendable via USB,  but for those using the software on other platforms this is an unnecessary restriction and stops midi devices being detected on (e.g.) gpio.
and I dont see any real advantage for the Norns hardware to do this additional check.

thoughts?

catfact:
(sorry wrong button.)

i'm not sure what you're suggesting. we have the usb subsystem monitor to support hotplugging of USB MIDI devices. i suppose you wouldn't need it if you had GPIO midi but of course the norns hardware doesn't.

TheTechnobear:
dev_monitor_scan() 
device_monitor.c : 154

catfact:
ok thanks, so IIUC you just want to remove these tests:
https://github.com/monome/norns/blob/master/matron/src/device/device_monitor.c#L153
https://github.com/monome/norns/blob/master/matron/src/device/device_monitor.c#L250

these of course affect all devices, not just midi. they are not unnecessary _if_ you have PCI or other subsystems that collide with the naming rules. (which was the case when this module was written, on x86 systems.)

i agree that it might be fine to remove these entirely on norns or other stripped down hardware.

TheTechnobear:
I’d say that’s more a distribution thing, rather than architecture - but sure paths should not be hardcoded. 
Really the midi detection stuff should all be done thru the alsa api , not filenames.  
And then the udev should just be used as a ‘watch’ for new devices. 
(And this is only applicable to usb since that’s the only thing we are supporting hotswap for )

artfwo:
We used ALSA originally for midi detection, but decided to stick with udev for having a unified device detection mechanism for grid, midi, and hid. The i/o was also switched to alsa rawmidi interface so we could handle midi in lua, and since rawmidi requires `/dev/midi` paths there's a path finder in device_monitor too.

TheTechnobear:
@artfwo ok, fair enough... I should have raised the alsa api as a separate discussion point.

my focus really on using udev /dev_monitor usb as its the only hot-swapping we have.
 
[...]

PR #532, opened 2018-08-31
device monitor return non-usb midi devices

TheTechnobear:
see https://github.com/monome/norns/issues/527

expose non-usb midi devices e.g. gpio

### DIFF
--- a/matron/src/device/device_monitor.c
+++ b/matron/src/device/device_monitor.c
@@ -150,9 +150,7 @@ int dev_monitor_scan(void) {
             dev = udev_device_new_from_syspath(udev, path);
 
             if (dev != NULL) {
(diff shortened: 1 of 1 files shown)


catfact:
i still don't like this:

- it means that when running the code on a laptop, we get a lot of noise and confusion as norns attempts to use all the other TTY and HID devices. the subsystem check is there for the simple reason that we would like to ignore these and have a similar experience when testing norns code on laptop as on norns HW.

- the change has no application to the norns hardware. i'm also failing to see how it has much application on other, hypothetical hardware - the entire point of `device_monitor.c` is to _monitor_ hotplugged devices;  if you are using some other hardware with persistent GPIO midi, why do you need to detect it?

PR #533, opened 2018-08-31
fix tape play

### LINKED ISSUE #531: TAPE broke
regression!

TAPE record now only makes blank tapes.
TAPE plays not-blank tapes as silence.

antonhornquist:
fixes #531

### DIFF
--- a/sc/core/Crone.sc
+++ b/sc/core/Crone.sc
@@ -242,10 +242,18 @@ Crone {
 				player.stop;
 			};
 			fork {
+				// TODO: old school SoundFile.cue SynthDef that actually works
+				SynthDef('cronetape', { | out, amp = 1, bufnum, sustain, ar = 0, dr = 0.01 gate = 1 |
+					Out.ar(out, VDiskIn.ar(2, bufnum, BufRateScale.kr(bufnum) )
+					* Linen.kr(gate, ar, 1, dr, 2)
+					* EnvGen.kr(Env.linen(ar, sustain - ar - dr max: 0 ,dr),1, doneAction: 2) * amp)
+				}).add;
+
 				playerFile = filename;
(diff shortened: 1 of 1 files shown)


antonhornquist:
i suspect issue is in supecollider core, the default SynthDef for `SoundFile.cue` did not works on the device. i replaced it with an earlier one, and this should work.

https://github.com/supercollider/supercollider/commit/d41d61170285b6f984b935854a29815d8a74401e

https://github.com/supercollider/supercollider/pull/3713

antonhornquist:
please test on your side.

tehn:
will do.

nf:
Great catch @antonhornquist.

(for my own edification) I'm trying to understand what changed, and can't see how the upstream changes could have broken for us.

The only external change I see is the args to the instrument went from

`| out, amp = 1, bufnum, sustain, ar = 0, dr = 0.01 gate = 1 |`

to

`|out, amp=1, bufnum, sustainTime, atk=0, rel=0, gate=1|`

But we don't seem to use any of these args? Unless there's some acton at a distance that I'm not seeing? Would love to figure out the root cause here.

tehn:
tested on hardware, works!

antonhornquist:
@nf the ugen graph is different too. 

Compare the custom SynthDef I added here with the core one at SCClassLibrary/Common/Files/SoundFile.sc

antonhornquist:
@nf compare the files on the device. My own 3.9.3 SuperCollider installation has the working ugen in core. Not sure we’re running 3.9.3 on device...

This has now been acknowledged as a bug in 3.10 beta with - I believe - a fix upcomimg in 3.10. 

Anyhow, our norns tape logic will probably be rewritten and improved in the near future. :)

tehn:
@antonhornquist wondering if the updated sc broke this again?

Issue #399, opened 2018-06-05
sclang crash on system time sync


tehn:
sclang crashes when time/date change by a substantial amount

- this happens when connecting to a wifi network which syncs time/date

samdoshi:
What are you using to set the time on network connection? `systemd-timesyncd` or a full NTP client?

Does the hardware have an RTC?

tehn:
the hardware does not have an RTC. it didn't occur to me that this would ever be an issue unfortunately.

samdoshi:
Okay. Apparently you can use [`fake-hwclock`](https://packages.debian.org/buster/fake-hwclock) to stop the time coming back up as the UNIX epoch (1st Jan 1970),  instead it will come up with the time set the same as when Norns was last powered down.

Something like `chrony` (an NTP client) will slew the clock rather than step it, but that's more designed to handle differences of a few seconds.

I guess the question is what counts as a "substantial amount"? If the issue is that it crashes when going from 1st Jan 1970 to the current date, but not if jump is only a few weeks/months then `fake-hwclock` might be good enough.

Otherwise someone will need to fire up `gdb` and figure out why `sclang` is crashing.

This might be handy for setting up a reproducible crash on a desktop machine:

https://serverfault.com/questions/138325/faking-the-date-for-a-specific-shell-session

catfact:
@artfwo has already done a lot of work on this including with `fake-hwclock`. He's out for the week at Linux audio conference. Maybe will have more to add. oddly, spoofing the clock doesn't crash sclang (necessarily) but does mess up various ugens in various ways

samdoshi:
Bother. Tried using `libfaketime` to crash `sclang`:

```
$  FAKETIME_STOP_AFTER_SECONDS=10 faketime "1970-01-01" sclang 
```

(skipped some output)

```
*** Welcome to SuperCollider 3.9.3. *** For help type ctrl-c ctrl-h (Emacs) or :SChelp (vim) or ctrl-U (sced/gedit).
qt5ct: palette support is disabled
sc3> thisThread.seconds
-> 7.212063747
sc3> thisThread.seconds
-> 1528231717.044
sc3> 
Exiting sclang (ctrl-D)
sc3> main: waiting for input thread to join...
main: quitting...
cleaning up OSC
```

The 2 calls to `thisThread.seconds` are a few seconds apart. No crash.

catfact:
did you have a server running though? the crash is in scsynth, not sclang

[...]

PR #554, opened 2018-09-20
audio reset within menu

tehn:
issues systemctl messages for crone.

auto-disables crone when turning on wifi in router mode.

LOG removed from SYSTEM menu.

this is a short-term fix.

### DIFF
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -30,7 +30,8 @@ local pWIFI = 8
 local pSYNC = 9
 local pUPDATE =10
 local pLOG = 12
-local pSLEEP = 13
+local pRESET = 13
+local pSLEEP = 14
 
 -- page pointer
 local m = {}
@@ -613,7 +614,7 @@ m.redraw[pHOME] = function()
   screen.level(15)
   local line = string.upper(norns.state.name)
   --if(menu.scripterror and state.script ~= '') then
-  if(menu.scripterror) then
+  if(menu.scripterror and menu.errormsg ~= 'NO SCRIPT') then
     line = line .. " (error: " .. menu.errormsg .. ")"
   end
   screen.text(line)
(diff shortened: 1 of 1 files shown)


tehn:
while there isn't a proper way to see if NTP caught up, and you have to manually re-enable audio, this is a semi-effective fix for #399

tehn:
pushing this through. comments still welcome.

nf:
👍 

While ideally we'd like to the point where this is unnecessary, having a RESET option in the SYSTEM menu seems like a sound short-term usability improvement.

catfact:
this implementation gives me the creeps - a superuser OS call smack in the middle of the menu spaghetti. we have an `audio` lua module, an `audio.reset` method, it retriggers a matron/crone handshake and makes sure we get engine reports &c! it was a lot of work! can we use that? if it doesn't work correctly can we fix it?

tehn:
sure thing, i'll fix it

tehn:
unfortunately `restart_audio()` (lib recompile) does not reset whatever mechanism within sc that gets jammed/backlogged on a time shift. my steps to reproduce:

- boot norns, running `awake`, connect via `screen`, adjust time with `sudo date --set="1:00:00"`
- connect to wifi router
- connect to norns via ssh, run `htop`
- check for timeshift via screen (or just listen for `awake` to stop)
- run `maiden` and type `restart_audio()` into REPL
- `htop` still peaked. it takes about 2 minutes for sc to process the time shift (as of right now) which is followed by a huge bang of backed-up note triggers

i'm wondering if once sc gets into this timeshift state if it can even process incoming commands, ie, if we wanted to issue a graceful shutdown.

the better approach would likely be to shut down sc prior to activating a wifi router connection, monitor the date, and then possibly even just self-restart matron (as to avoid writing a secondary handshake negotiation).

will keep working on it but very open to suggestions and ideas.

artfwo:
@tehn does restarting the server (s.reboot) help?

tehn:
on timeshift, sc is totally 100% hosed.

to reproduce:

- boot norns, `screen` in. `sudo date --set="1:00:00"` then `./stop.sh` then `./matron.sh > /dev/null &` then `sclang` (so we have a sc REPL)
- turn on `awake`, activate wifi
- on time shift, sc (in `screen`) is stuck. commands can be typed in but they are not processed... enter just gives a new line

PR #565, opened 2018-09-26
MIDI channel pressure

markwheeler:
Added support for 'channel pressure' to midi.lua
Also suggesting to rename polyphonic 'aftertouch' -> 'key pressure' for clarity.

This ref seems to use this as preferred naming but I can revert if need be: https://www.midi.org/specifications-old/item/table-1-summary-of-midi-message

Note: I don't have a device which supports polyphonic key pressure to double check it but it should be fine :)


tehn:
updated study 4 to match, thank you!

PR #571, opened 2018-10-01
param ids

### LINKED ISSUE #517: identifiers for params
introduce param IDs - short, not necessarily human-readable identifiers that will be used for:

- keys in psets
- osc addresses
- param.set, e.g. `param.set('1_filter', 500)` instead of `param.set('1: filter', 500)`

param names will remain and will be used for display in the UI.

artfwo:
this adds ids to parameters. we want different ids for params to be able to address it with OSC or save in a file ('freq_mod_env1' vs 'Frequency Mod (Env-1)'). fixes #517

### DIFF
--- a/lua/params/control.lua
+++ b/lua/params/control.lua
@@ -12,10 +12,11 @@ local tCONTROL = 3
 -- @param name of param
 -- @param controlspec
 -- @param formatter
-function Control.new(name, controlspec, formatter)
+function Control.new(id, name, controlspec, formatter)
   local p = setmetatable({}, Control)
   p.t = tCONTROL
   if not controlspec then controlspec = ControlSpec.UNIPOLAR end
+  p.id = id
   p.name = name
   p.controlspec = controlspec
   p.formatter = formatter
--- a/lua/paramset.lua
+++ b/lua/paramset.lua
@@ -76,8 +76,8 @@ function ParamSet:add_taper(id, name, min, max, default, k, units)
 end
 
 --- add trigger
-function ParamSet:add_trigger(name)
-  self:add { param=trigger.new(name) }
+function ParamSet:add_trigger(id, name)
+  self:add { param=trigger.new(id, name) }
 end
 
 --- print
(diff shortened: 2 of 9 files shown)


lua/params/control.lua
```diff
@@ -12,10 +12,11 @@ local tCONTROL = 3
 -- @param name of param
 -- @param controlspec
 -- @param formatter
-function Control.new(name, controlspec, formatter)
+function Control.new(id, name, controlspec, formatter)
```
antonhornquist: would it not make more sense to handle `id` outside of param classes - that is, in the paramset class instead, since ìd`s constitute keys of the paramset ?

lua/paramset.lua
```diff
@@ -43,40 +43,41 @@ function ParamSet:add(args)
   local param = args.param -- param is mandatory
   table.insert(self.params, param)
   self.count = self.count + 1
-  self.lookup[param.name] = self.count
+  self.lookup[param.id] = self.count
```
antonhornquist: alternative: `args.id` used instead of `param.id` here

lua/paramset.lua
```diff
@@ -43,40 +43,41 @@ function ParamSet:add(args)
...
+function ParamSet:add_option(id, name, options, default)
+  p = option.new(id, name, options, default)
+  self:add { param=option.new(id, name, options, default) }
 end
 
 --- add control
-function ParamSet:add_control(name, controlspec, formatter)
-  self:add { param=control.new(name, controlspec, formatter) }
+function ParamSet:add_control(id, name, controlspec, formatter)
+  self:add { param=control.new(id, name, controlspec, formatter) }
```
antonhornquist: alternative would be `self:add { id=id, param=control.new(name, controlspec, formatter)`

lua/params/control.lua
artfwo: it makes sense, but i'd also like to keep the id in params so we can pass a param instance around and avoid looking up the id when we need it.

lua/paramset.lua
artfwo: also agreed, but i'd rather make this change in a different PR.

PR #572, opened 2018-10-01
metro syntax change

tehn:
re: github.com/monome/norns/issues/570

this is a breaking change:

- `metro:init()` removed as it was unused (simply resets metro to defaults)
- `metro.alloc` is now `metro.init`
- `callback` renamed `event`
- table syntax allowed in `init`

ie

```
m = metro.init()
n = metro.init{event = function(x) print(x) end}
o = metro.init{time = 2, count = 10, event = function(x) print(x) end}
```

and you can still

```
cb = function(x) print(x) end
q = metro(cb, 1, -1)
```

if you want.

## much breakage in dust, however

- `metro.alloc` -> `metro.init`
- all instances of metro, ie `m.callback` -> `m.event`

count is -1 by default, need not be specified
time is 1 by default, need not be specified

## norns study must be updated (!)

this merge should coincide with dust fix and study fix


lua/metro.lua
```diff
@@ -14,118 +14,123 @@ Metro.metros = {}
 Metro.available = {}
 Metro.assigned = {}
 
---- assign
+
+--- init/assign
 -- "allocate" a metro (assigns unused id)
```
pq: "allocate" => initialize?

lua/metro.lua
```diff
@@ -14,118 +14,123 @@ Metro.metros = {}
...
-	if cb then m.callback = cb end
-	if time then m.time = time end
-	if count then m.count= count end
-	return m
+function Metro.init(arg, arg_time, arg_count)
+  local event = 0
+  local time = arg_time or 1
+  local count = arg_count or -1
+
+  print("what "..time)
```
pq: remove (unless this is intended)?

lua/metro.lua
```diff
@@ -14,118 +14,123 @@ Metro.metros = {}
...
-end
-
---- reset to default state
-function Metro:init()
-    self.id = id
-    self.time = 1
-    self.count = -1
-    self.callback = nil
-    self.init_stage = 1
+  local m = {}
```
pq: while you're at it, consider a table constructor?

```lua
m.props = {
  id = id,
  time = 1,
  count = -1,
  ...
}
```

lua/metro.lua
```diff
@@ -14,118 +14,123 @@ Metro.metros = {}
...
+  m.props = {}
+  m.props.id = id
+  m.props.time = 1
+  m.props.count = -1
+  m.props.event = nil
```
pq: isn't this `nil` already (or do we check for the key somewhere?)

lua/metro.lua
```diff
@@ -14,118 +14,123 @@ Metro.metros = {}
...
+  m.props.init_stage = 1
+  setmetatable(m, Metro)
+  return m
 end
 
 --- start a metro
 -- @param time - (optional) time period between ticks (seconds.) by default, re-use the last period
 -- @param count - (optional) number of ticks. infinite by default
 -- @param stage - (optional) initial stage number (1-based.) 1 by default
 function Metro:start(time, count, stage)
```
pq: perhaps consider a similar table approach here? iow:

```lua
m:start{time = 0.1, count = 1, stage = 3}
```

catfact:
@tehn do you still want to do this?

tehn:
holding off for the next breaking-change release, and need to do doc revisions which are not ready.

Issue #569, opened 2018-09-30
params:add (helper function)


tehn:
ie

```lua
params:add{control: 'freq_mod_env1', name: 'Frequency Mod (Env-1)', spec: specs.FREQ_MOD_ENV, action: engine.freqModEnv}
```

- name is derived from the value for control if not explicitly specified.
- param type `control` derived from the fact that it's the only arg of the set {control, option, number, etc} present
- throw a friendly error if the main arg isn't a table

cc @pq

pq:
as motivation, the call proposed above would replace these two-lines

```lua
params:add_control("Frequency Mod (Env-1)", specs.FREQ_MOD_ENV)
params:set_action("Frequency Mod (Env-1)", engine.freqModEnv)
```

in the case where index is OK for a name, the call could be even simpler:

```lua
paramset:add{ control: 'monitor',  spec: cs.MAIN_LEVEL, action: function(x) norns.audio.monitor_level(x) end }
```

since this is pretty common, param initialization could get a bunch simpler for the common case.

tehn:
changed spec, now explicit `type` is required.

```
paramset:add{ type="control", id="level", action=engine.level }
```

PR #574, opened 2018-10-02
paramset add table helper

tehn:
fixes https://github.com/monome/norns/issues/569

### DIFF
--- a/lua/paramset.lua
+++ b/lua/paramset.lua
@@ -66,8 +66,12 @@ function ParamSet:add(args)
       local id = args.trigger
       local name = args.name or id
       param = trigger.add(id, name)
+    else
+      print("paramset.add() error")
+      return nil
     end
   end
+
 
   table.insert(self.params, param)
   self.count = self.count + 1
--- a/lua/paramset.lua
+++ b/lua/paramset.lua
@@ -57,17 +57,17 @@ function ParamSet:add(args)
     local id = args.id
     local name = args.name or id
 
-    if args.number then
+    if args.type == "number"  then
       param = number.new(id, name, args.min, args.max, args.default)
-    elseif args.option then
+    elseif args.type == "option" then
       param = option.new(id, name, args.options, args.default)
-    elseif args.control then
+    elseif args.type == "control" then
       param = control.add(id, name, args.controlspec, args.formatter)
-    elseif args.file then
+    elseif args.type == "file" then
       param = file.add(id, name, args.path)
-    elseif args.taper then
+    elseif args.type == "taper" then
       param = taper.add(id, name, args.min, args.max, args.default, args.k, args.units)
-    elseif args.trigger then
+    elseif args.type == "trigger" then
       param = trigger.add(id, name)
     else
       print("paramset.add() error: unknown type")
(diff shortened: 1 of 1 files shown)


lua/paramset.lua
```diff
@@ -40,11 +40,43 @@ end
 
 --- add generic parameter
```
pq: perhaps some docs here?

lua/paramset.lua
```diff
@@ -40,11 +40,43 @@ end
...
 function ParamSet:add(args)
-  local param = args.param -- param is mandatory
+  local param = args.param
+  if param == nil then
+    if args.number then
+      local id = args.number
+      local name = args.name or id
+      param = number.new(id, name, args.min, args.max, args.default)
```
pq: in general, i wonder if, in addition to docs, we don't want to some validation to ensure required args are provided.

if we had something like

```lua
local function checkRequired(args, names) {
  for _,name in names do
    if args.name == nil then error('argument "'..name..'" is required')
  end
} 
```

we could do something like:

```lua
  checkRequired(args, {'min', 'max'});
```

which should report something useful like `argument "max" is required` (only better) if someone did something like:

```lua
params:add{number = 'foo', min = 0}
```

(caveat: all untested; just food for thought!)

lua/paramset.lua
```diff
@@ -40,11 +40,43 @@ end
...
+    elseif args.taper then
+      local id = args.taper
+      local name = args.name or id
+      param = taper.add(id, name, args.min, args.max, args.default, args.k, args.units)
+    elseif args.trigger then
+      local id = args.trigger
+      local name = args.name or id
+      param = trigger.add(id, name)
+    else
+      print("paramset.add() error")
```
pq: how about something like:

```lua
print('paramset.add() error: unspecified param type (expected   "option", "control", "file", ... ')
```

artfwo:
i might have missed some earlier discussions, but why is `add` taking `number`, `trigger`, etc. arguments? why not `add{type='number', id='foo'}`?

tehn:
> why not` add{type='number', id='foo'}`?

this is probably more straightforward and normal, though more verbose of course. i think the goal was to make a concise syntax. @pq @artfwo

lua/paramset.lua
tehn: this sounds nice, certainly. but most fields really aren't required... they just need sensible defaults. have you scanned through to see if there are many required fields?

pq:
> this is probably more straightforward and normal, though more verbose of course.

yeah.  i like it concise but maybe it's a bridge too far.

curious what other folks think.

/cc @markwheeler @antonhornquist @simonvanderveldt

lua/paramset.lua
```diff
@@ -39,12 +39,46 @@ function ParamSet:add_separator()
...
+      return nil
+    elseif args.id == nil then
+      print("paramset.add() error: id required")
+      return nil
+    end
+
+    local id = args.id
+    local name = args.name or id
+
+    if args.number then
```
artfwo: hmm, but shouldn't we check args.type here instead?

tehn: sorry totally messed this up

PR #576, opened 2018-10-02
param missing in paramset error message

antonhornquist:
refactored param lookup, and introduced a more descriptive error message when param is not found in paramset


pq:
nice!

PR #585, opened 2018-10-07
map incoming osc to params

### LINKED ISSUE #501: map incoming OSC to params
for now, easiest to use id numbers given the parameter names can have spaces and weird characters.

so incoming OSC

`/param/1 x`

could map to `params.set(1, x)` (?)

artfwo:
fixes #501 

incoming numeric osc param values should be within the specific parameter range (i.e. -80, 20). file and option params require string argument value.

### DIFF
--- a/lua/paramset.lua
+++ b/lua/paramset.lua
@@ -16,19 +16,22 @@ local ParamSet = {
   tCONTROL = 3,
   tFILE = 4,
   tTAPER = 5,
-  tTRIGGER = 6
+  tTRIGGER = 6,
+  sets = {}
 }
 
 ParamSet.__index = ParamSet
 
 --- constructor
 -- @param name
-function ParamSet.new(name)
+function ParamSet.new(id, name)
   local ps = setmetatable({}, ParamSet)
+  ps.id = id or ""
   ps.name = name or ""
   ps.params = {}
   ps.count = 0
   ps.lookup = {}
+  ParamSet.sets[ps.id] = ps
   return ps
 end
 
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -48,7 +48,7 @@ menu.scripterror = false
 menu.errormsg = ""
 
 -- mix paramset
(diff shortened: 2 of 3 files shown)


lua/paramset.lua
```diff
@@ -16,19 +16,22 @@ local ParamSet = {
...
 
 ParamSet.__index = ParamSet
 
 --- constructor
 -- @param name
-function ParamSet.new(name)
+function ParamSet.new(id, name)
   local ps = setmetatable({}, ParamSet)
+  ps.id = id or ""
   ps.name = name or ""
```
tehn: would this make sense to inherit `id` if `name` not specified?

ie

```
ps.name = name or id or ""
```

(does chaining `or`'s work in lua?)

lua/paramset.lua
```diff
@@ -16,19 +16,22 @@ local ParamSet = {
...
   ps.params = {}
   ps.count = 0
   ps.lookup = {}
+  ParamSet.sets[ps.id] = ps
```
tehn: seems like `id` needs to be required in the initializer then?

tehn:
does the global `params` table (defined in `startup.lua`) need a `name` and `id`?

artfwo:
@tehn the current logic allows empty IDs - the script params can be addressed by `/param/foo`, mix params can be addressed by `/params/mix/foo.

the "" id will be replaced by an empty paramset whenever a script is loaded, but additional sets can be created if we ever need to namespace parameters.

lua/paramset.lua
artfwo: i am keeping this separated like with param ids, in case we have spaces and similar characters in the name (which is currently not even used, but this PR sets it for `mix` pset to keep its params in a separate osc namescpae).

lua/paramset.lua
artfwo: we have `ps.id = id or ""`, so no, empty string remains the key for the script pset.

PR #586, opened 2018-10-07
add info to print statement on maxing out metro.alloc

jlmitch5:
I ran into an issue where I had allocated too many metros.  I wasn't sure what the statement "nothing available" meant, so I had to go look through the code.  This pr just updates what is printed to the REPL in this case to something a bit less ambiguous.

### DIFF
--- a/lua/metro.lua
+++ b/lua/metro.lua
@@ -34,7 +34,7 @@ function Metro.alloc (cb, time, count)
 	if count then m.count= count end
 	return m
     end
-    print("metro.alloc: already used max numbert of allocated script metros")
+    print("metro.alloc: already used max number of allocated script metros")
     return nil
 end
 
--- a/lua/metro.lua
+++ b/lua/metro.lua
@@ -34,7 +34,7 @@ function Metro.alloc (cb, time, count)
 	if count then m.count= count end
 	return m
     end
-    print("metro.alloc: nothing available")
+    print("metro.alloc: already used max numbert of allocated script metros")
     return nil
 end
 


tehn:
thanks @jlmitch5 !

excited to see what you made that uses 30 metros :)

catfact:
btw: number of statically allocated metro objects is totally arbitrary, and there isn't a real downside to having more.

but generally, metros are threads, and there is some performance overhead for each thread that actually runs: say some tens of microseconds for the thread itself, plus penalties for cache misses. it's maybe a good idea to structure complex / nested timings as a single (fast) metro servicing a number of soft-timers. (should we add such a structure to lua libs?)

jlmitch5:
> excited to see what you made that uses 30 metros :)

@tehn I am working on a version of boingg by @rross101 in dust that allows for the selection of different states by clicking in a particular row of grid.  My first approach was just to have anything stateful related to the state of the cycles be duplicated in an array--since each cycle had it's own dedicated metro, the duplication across the states caused many metros to be allocated.

I did a quick revision so that each state shared a metro (so back to the original number of allocations) which so far has worked.  I believe based on how they are being used that it might be possible to reduce the script further down to using only a single metro.

> it's maybe a good idea to structure complex / nested timings as a single (fast) metro servicing a number of soft-timers. (should we add such a structure to lua libs?)

@catfact while I don't think this is necessary for the scenario I ran into, I think this could be useful, for example, being able to create loops quantized (would be cool if you could set the amount of quantization) to a clock.  That being said, at least for this case, I think something could be coded up in a script and then ported into lib afterwards (though you might be thinking along the lines of a different scenario where that's not doable)

rross101:
Looking forward to seeing this.

PR #589, opened 2018-10-08
re-gen lua docs

pq:
generated like so:

```
[~/norns/repo/lua] ldoc -c ../config.ld .
```

is this right?

/cc @tehn @catfact


tehn:
thanks!

a reminder that `luarocks` isn't installed by default (we should probably add it?)

PR #611, opened 2018-10-22
cleanup menu.lua, move mix

tehn:
move mix paramset into its own file, makes menu.lua less daunting


lua/mix.lua
```diff
@@ -0,0 +1,152 @@
+local cs = require 'controlspec'
+local fx = require 'effects'
+
+-- mix paramset
+local mix = paramset.new("mix", "mix")
+local cs_MAIN_LEVEL = cs.new(-math.huge,0,'db',0,0,"dB")
+mix:add_control("output", "output", cs_MAIN_LEVEL)
```
pq: aside: it'd  be interesting to refactor these to use the table param API.  i'm hoping this'd help with general legibility...

tehn: agreed. tiny steps for now :)

PR #614, opened 2018-10-22
free engine on cleanup

tehn:
properly resets engine on cleanup

@markwheeler shouldn't need to dummy-load testsine now


markwheeler:
great, thank you!

Issue #618, opened 2018-10-24
TAPE playback broken


tehn:
regression

record works fine

tehn:
quite possibly an sc issue since we updated to 3.10

https://github.com/monome/norns/pull/533

tehn:
fixed by https://github.com/monome/norns/pull/620

PR #620, opened 2018-10-24
Crone tape

catfact:
tape synthdef looked funky, like it was multiplying the output by an envelope specifiction.

also doesn't really need varispeed


tehn:
fixed https://github.com/monome/norns/issues/618

PR #621, opened 2018-10-26
reverb light

tehn:
revert to lighter-cpu reverb. adds only 9% now.


tehn:
seeing some wild variation in cpu use, that 9% isn't accurate. will keep testing.

PR #632, opened 2018-11-02
make parent directories for sc install locations

catfact:
...  in case they got nuked somehow

(@okyeron is this what you meant?)

### DIFF
--- a/sc/install.sh
+++ b/sc/install.sh
@@ -3,7 +3,7 @@
 sc_ext_dir=~/.local/share/SuperCollider/Extensions
 norns_ext_dir=$sc_ext_dir/norns
 rm -rf $norns_ext_dir
-mkdir $norns_ext_dir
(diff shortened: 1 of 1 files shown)


okyeron:
I think it's the supercollider extension subdirectories for  `norns` and  `dust` that are the problem

my errors looked like this:
```
mkdir: cannot create directory ‘/root/.local/share/SuperCollider/Extensions/norns’: No such file or directory
ln: failed to create symbolic link '/root/.local/share/SuperCollider/Extensions/norns/TrigPhasor.sc': No such file or directory
ln: failed to create symbolic link '/root/.local/share/SuperCollider/Extensions/norns/FaustCompressor.sc': No such file or directory

```

catfact:
Right, this would fix that by adding -p like you said. Dunno why it would error in first place unless you wiped out whole SC extensions dir, but we should allow for that anyways

PR #638, opened 2018-11-08
Arc Support

okyeron:
Revisiting @scanner-darkly PR for arc support.

This adapts the PR scanner-darkly did back in June to the most recent norns/grid codebase.
https://github.com/monome/norns/pull/424/files

Some comments by @artfwo on the original PR still apply and need to be addressed.

### DIFF
--- a/lua/arc.lua
+++ b/lua/arc.lua
@@ -1,0 +1,236 @@
+--- Arc class
+-- @module arc
+-- @alias Arc
+require 'norns'
+
+
+---------------------------------
+-- Arc device class
+
+local Arc = {}
+Arc.devices = {}
+Arc.list = {}
+Arc.vport = {}
+for i=1,4 do
+  Arc.vport[i] = {
+    name = "none",
+    callbacks = {},
+    index = 0,
+    led = function() end,
+    all = function() end,
+    refresh = function() end,
+    attached = false
+  }
+end
+Arc.__index = Arc
+
+--- constructor
+-- @tparam integer id : arbitrary numeric identifier
+-- @tparam string serial : serial
+-- @tparam string name : name
+-- @tparam userdata dev : opaque pointer to device
+function Arc.new(id, serial, name, dev)
+  local a = setmetatable({}, Arc)
+  a.id = id
+  a.serial = serial
+  name = name .. " " .. serial
+  --while tab.contains(Arc.list,name) do
+  --  name = name .. "+"
+  --end
+  a.name = name
+  a.dev = dev -- opaque pointer
+  a.key = nil -- key event callback
+  a.remove = nil -- device unplug callback
+  a.encs = arc_encs(dev)
+  a.ports = {} -- list of virtual ports this device is attached to
+
+  -- autofill next postiion
+  local connected = {}
+  for i=1,4 do
+    table.insert(connected, Arc.vport[i].name)
+  end
+  if not tab.contains(connected, name) then
+    for i=1,4 do
+      if Arc.vport[i].name == "none" then
+        Arc.vport[i].name = name
+        break
+      end
+    end
+  end
+
+  return a
+end
+
+--- static callback when any arc device is added;
+-- user scripts can redefine
+-- @param dev : a Arc table
+function Arc.add(dev)
+  print("arc added:", dev.id, dev.name, dev.serial)
+end
+
+--- scan device list and grab one, redefined later
+function Arc.reconnect() end
+
+--- static callback when any arc device is removed;
+-- user scripts can redefine
+-- @param dev : a Arc table
+function Arc.remove(dev) end
+
+--- set state of single LED on this arc device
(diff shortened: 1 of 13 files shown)


tehn:
rad, thanks! i'll take a look.

okyeron:
With last commit, the function for setting all leds is now 2 parameters - `arc.all(1,0)` so all leds for individual encoders can be set. Will probably need another lua function to set *all* leds on *all* encoders.

Not sure how led/map commands are working yet. Seems like not treating each encoder as a `quad` (like on grid) would be more efficient when sending led/map to an individual encoder group of leds.

artfwo:
@okyeron is this code working for you? I'm getting a `PANIC: unprotected error in call to Lua API (attempt to index a nil value)` when trying to run this locally.

artfwo:
yeah, but then the process crashes. something wrong with my local build, doesn't seem to happen with a clean checkout from you branch. is that okay to commit changes to your branch directly?

okyeron:
Yes - i think so?

artfwo:
Okay, then go ahead with testing your changes, I can work with your currently pushed code in the meantime.

okyeron:
I created an arc-dev branch that has additional changes:
* added enc parameter to refresh so each ring can be updated individually instead of all four at once. This seems to fit better with the /ring/map command
* added support for enc buttons. These exist on only one generation of Arc, but why not support it?
* found a problem last night that device was not being "released" when changing scripts. Still investigating this.

Here are a couple lua files I started on for testing
https://gist.github.com/okyeron/29d0a8be36c15baaaec3990a00aed05e
https://gist.github.com/okyeron/406ad051b432234d97ded8fb9a1fbde9

artfwo:
good job! a couple of comments regarding the later branch:

push encoder support, while good for the owners of the pushable arcs, doesn't provide the api that will be compatible with the rest of devices. scripts for push encoders won't run with the rest of arcs.

`refresh` in grid api is supposed to refresh the entire device, so let's leave it the same way in arc for  consistency. we don't have a method similar to /grid/map or /ring/map, because you're supposed to use the api (set_led) to draw "offscreen", then call refresh() to update the leds on device.

okyeron:
FWIW - my want for push encoder support is because I'm making an encoder box which impersonates an Arc. 

I'm not sure I understand what you mean by "doesn't provide the api that will be compatible with the rest of devices". Which rest of devices? Norns would just end up sending a serial `/prefix/enc/key n d` command I assume this would be ignored by any device that doesn't support that command. So why is it a problem to include that?

Seems like this would be more of an easter egg for those button arc folks (and useful to me for future diy things). 

On `refresh`, the particular use case I'm thinking of  is refreshing different arc rings on various different timescales - which could get bogged down (or blinky?) by refreshing all 4 every time at high framerates.

tehn:
it's a good idea to support encoder key input. i like the idea of DIY emulated devices.

it shouldn't break the pattern, it just needs a separate callback. (usually this callback won't be used.) so there needs to be another arc event.

for refresh, we should have dirty flags for each ring the same way grids have a dirty flag per quadrant. see here: https://github.com/monome/norns/blob/master/matron/src/device/device_monome.c

the user should absolutely only have one refresh function without args, follow the grid pattern.

okyeron:
>it just needs a separate callback.

I *think* I did this (there's both  norns.arc.enc and norns.arc.key callbacks), but I don't think i did it correctly (and not sure what correct would be here). Initially I ended up with both enc and key events triggering at the same time, but I added a variable to tell them apart. Maybe that needs to be a table like the midi events use, but I'm not sure how to do that? 

FWIW - I'm kinda flying by the seat of my pants here - mostly copying what was there for grid and coding by trial and error. :)

tehn:
i haven't checked your code yet, so i was commenting a little blind based on the discussion.

should have time tomorrow to look though! thank you for diving in!

okyeron:
OK - new commit to master that 
- rolls back the led_all to have just one argument (brightness) and updates all leds on device
- adds button support

I can't seem to find one issue in my lua code where I have to set arc.all(1) instead of arc.all(0) to get the leds to clear entirely. I double checked this in weaver.c and i don't see anywhere else where this is handled.

test scripts:
https://gist.github.com/okyeron/3b2a3278a1b02b28f1de32d67f44174d
https://gist.github.com/okyeron/e7e2e94393f69937fa16a4f82dffc3f5

artfwo:
@okyeron i cannot push more code to this PR (have you enabled 'edits from maintainers'?), so I'm creating a new one with some cleanup and overhaul to your code, see #641

okyeron:
understood.

Not sure why you could not push to this PR, although 'edits from maintainers' is enabled. Sorry... I'm still a bit of a git newb.

artfwo:
that's fine, please send me your github email address (via lines), so I can include you in the later PR co-authors.

PR #643, opened 2018-11-19
grid rotation

okyeron:
Adds function to rotate grid.

```
local g = grid.connect()
g.rotation(1) -- 0,1,2,3 are options for 0,90,180,270
```

Needs menu.lua work to make it a configurable setting in menu.

### DIFF
--- a/lua/grid.lua
+++ b/lua/grid.lua
@@ -18,6 +18,7 @@ for i=1,4 do
     index = 0,
     led = function() end,
     all = function() end,
+    rotation = function() end,
     refresh = function() end,
     cols = 0,
     rows = 0,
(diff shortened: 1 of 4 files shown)


tehn:
nice!! have you tested this on hardware?

i can do the menu UI.

okyeron:
I've only got my teensy grid to test on - and it works at 90 degrees for me. 

The 180 and 270 rotations don't work right on my diy grid from the Monome Home Maxpat, so there's some other bug in my grid code so I cant confirm they are working properly.

tehn:
ok i'll test it out on hardware first

tehn:
works great! thank you!

Issue #646, opened 2018-11-20
midi note-on, velocity zero not handled as note-off


okyeron:
while checking out midi with molly_the_poly script tonight I found I'm not getting note off messages while using a QuNexus.

Looking at midi.lua, it appears it does not account for some instruments that do not send a note-off byte, but rather a note-on byte with velocity 0 instead.

For example if I do
```
m = midi.connect(2)
m.event = function(data)
  local d = midi.to_msg(data)
  
  local status = data[1]
  local data1 = data[2]
  local data2 = data[3]
  print(status .. " " .. data1 .. " ".. data2)
  tab.print(d)
end
```
returns
```
144 59 105
vel	105
type	note_on
note	59
ch	1
144 59 0
vel	0
type	note_on
note	59
ch	1
```

so perhaps this needs to be changed in midi.lua?

```
   -- note on
  if data[1] & 0xf0 == 0x90 then
    if data[3] > 0 then
    --print("note")
    msg = {
      type = "note_on",
      note = data[2],
      vel = data[3],
      ch = data[1] - 0x90 + 1
    }
    elseif data[3] == 0 then
    --print("noteoff")
    msg = {
      type = "note_off",
      note = data[2],
      vel = data[3],
      ch = data[1] - 0x80 + 1
    }
    end
```

Should I submit this as a pull request?

Or... are there cases where a note-on, velocity zero should not be considered a note-off?

tehn:
yes that sounds good. send a PR

Issue #610, opened 2018-10-21
MIDI event called with empty data table


markwheeler:
When sending clock to norns, the MIDI event is often called with an empty `data` table.
This sometimes happens every tick (seems related to stopping/starting the external clock source) and sometimes happens when stop messages are sent. I mainly tested sending clock from Ableton Live via iConnectivity mio.

It's easy to ignore the empty tables but seems odd.

okyeron:
just posted this PR which I think will help with your issue
https://github.com/monome/norns/pull/658

markwheeler:
great yes that fixes it!

PR #648, opened 2018-11-20
midi note-off variation

okyeron:
Fix to midi.lua for some midi instruments that send a note-on, velocity zero instead of note-off

per https://github.com/monome/norns/issues/646

### DIFF
--- a/lua/midi.lua
+++ b/lua/midi.lua
@@ -186,13 +186,24 @@ function Midi.to_msg(data)
   local msg = {}
   -- note on
   if data[1] & 0xf0 == 0x90 then
-    --print("note")
-    msg = {
-      type = "note_on",
-      note = data[2],
-      vel = data[3],
-      ch = data[1] - 0x90 + 1
-    }
+    if data[3] > 0 then
+      --print("note")
+      msg = {
+        type = "note_on",
+        note = data[2],
+        vel = data[3],
+        ch = data[1] - 0x90 + 1
+      }
+      -- if velocity is zero then send note off
+    elseif data[3] == 0 then 
+      --print("noteoff")
+      msg = {
+        type = "note_off",
+        note = data[2],
+        vel = data[3],
+        ch = data[1] - 0x80 + 1
+      }
+    end
   -- note off
   elseif data[1] & 0xf0 == 0x80 then
     msg = {
--- a/lua/midi.lua
+++ b/lua/midi.lua
@@ -186,23 +186,15 @@ function Midi.to_msg(data)
   local msg = {}
   -- note on
   if data[1] & 0xf0 == 0x90 then
-    if data[3] > 0 then
-      --print("note")
-      msg = {
-        type = "note_on",
-        note = data[2],
-        vel = data[3],
-        ch = data[1] - 0x90 + 1
-      }
-      -- if velocity is zero then send note off
-    elseif data[3] == 0 then 
-      --print("noteoff")
-      msg = {
-        type = "note_off",
-        note = data[2],
-        vel = data[3],
-        ch = data[1] - 0x80 + 1
-      }
+    msg = {
+      note = data[2],
+      vel = data[3],
+      ch = data[1] - 0x90 + 1
+    }
+    if data[3] > 0 then 
+      msg.type = "note_on"
+    elseif data[3] == 0 then -- if velocity is zero then send note off
+      msg.type = "note_off"
     end
   -- note off
   elseif data[1] & 0xf0 == 0x80 then


okyeron:
I’m not actually sure what that byte is doing in the channel message. But I think you are correct and just changing msg.type would be fine.

okyeron:
I'm sure there's probably some situation where note-on plus vel=0 should NOT be a note-off, but if half the MIDI instrument manufacturers are doing it that way, what can you do? 🤷‍♂️ 

A configuration option would be nice, but wouldn't know how to implement that.

tehn:
accidentally approved, disregard.

> I’m not actually sure what that byte is doing in the channel message. But I think you are correct and just changing msg.type would be fine.

just change the msg.type. the change to channel will result in the wrong channel being sent. did you test this with your hardware?

lua/midi.lua
```diff
@@ -186,13 +186,24 @@ function Midi.to_msg(data)
...
   if data[1] & 0xf0 == 0x90 then
-    --print("note")
-    msg = {
-      type = "note_on",
-      note = data[2],
-      vel = data[3],
-      ch = data[1] - 0x90 + 1
-    }
+    if data[3] > 0 then
+      --print("note")
```
artfwo: no need for these commented out lines, the comment about this velocity check could be moved before this block.

lua/midi.lua
```diff
@@ -186,13 +186,24 @@ function Midi.to_msg(data)
...
+        ch = data[1] - 0x90 + 1
+      }
+      -- if velocity is zero then send note off
+    elseif data[3] == 0 then 
+      --print("noteoff")
+      msg = {
+        type = "note_off",
+        note = data[2],
+        vel = data[3],
+        ch = data[1] - 0x80 + 1
```
artfwo: has to be `ch = data[1] - 0x90 + 1`, because data byte is still for `note_on` message

artfwo:
there indeed was a controversy about this logic, but i think it was related to the low-level midi events and for the high-level api it will not harm.

okyeron:
updated. tested with Qunexus and Livid Block. 
(Having an problem with MK-225C controller, but that seems to be issue https://github.com/monome/norns/issues/610 - will investigate later)

is the following syntax OK?

```
  -- note on
  if data[1] & 0xf0 == 0x90 then
    msg = {
      note = data[2],
      vel = data[3],
      ch = data[1] - 0x90 + 1
    }
    if data[3] > 0 then 
      msg.type = "note_on"
    elseif data[3] == 0 then -- if velocity is zero then send note off
      msg.type = "note_off"
    end
````

**Comment by okyeron** [comment]
FWIW - my problem with the MK-225C controller is that it constantly sends an Active Sensing byte which barfs inside `function Midi.to_msg(data)`

my quick hack around this was to wrap most of Midi.to_msg in a check to be sure there is a second data byte
```
if data[2] ~= nil then     -- Ignore Active Sensing and others
~
end
```

Are clock/start/stop/continue (which also have a nil second byte) ignored somewhere?

**Comment by artfwo** [comment]
clock events don't have a second byte, their length is determined in matron (device_midi.c). what exactly is an active sensing byte?

**Comment by okyeron** [comment]
>what exactly is an active sensing byte?

Not really sure myself, but I see that message when I plug that device into my mac with MIDI monitor running.

[The MIDI spec](https://www.midi.org/specifications-old/item/table-2-expanded-messages-list-status-bytes) shows this:

11111110= FE= 254 | Active Sensing | -- | --

a google search gives me this description:
>A MIDI message sent by some MIDI devices all the time to confirm that they are still there and hasn’t gone off line. It carries no other information and hasn’t been widely implemented in the industry.

**Comment by artfwo** [comment]
a message with a status byte of 0xfe will be handled as other 1-byte messages, Midi.to_msg should just return `{ type = "other" }` for these messages, what exactly is the problem with them?

**Comment by okyeron** [comment]
Apparently 0xfe is not being handled as a 1-byte message and it tries to evaluate it as note/cc and maiden spits out a stream of errors about attempting a bitwise operation on nil.

looking at device_midi.c 0xfe is not handled explicitly and ends up with the default msg_len = 2;

I can file a separate issue for this

**Comment by okyeron** [comment]
Ok so I have a very basic connection script
```
m = midi.connect(1)
m.event = function(data)
  local d = midi.to_msg(data)
  tab.print(d)
end
```

With the MK-225C   as soon as this runs i get a stream of the following errors:

```
/home/we/norns/lua/midi.lua:188: attempt to perform bitwise operation on a nil value (field '?')
stack traceback:
	/home/we/norns/lua/midi.lua:188: in function 'midi.to_msg'
	/home/we/dust/scripts/okyeron/midi-example.lua:3: in field 'event'
	/home/we/norns/lua/midi.lua:120: in local 'event'
	/home/we/norns/lua/midi.lua:304: in function </home/we/norns/lua/midi.lua:292>
```

So here it appears to be trying to evaluate the 0xfe message as a note-on. and the second byte of data  is nil and causes the error.

**Comment by tehn** [comment]
FYI doesn't seem like it's handled, and defaulting to length 2:

https://github.com/monome/norns/blob/master/matron/src/device/device_midi.c#L98

feel free to submit a PR that has a case for 0xFE and just throws it away

**Comment by tehn** [comment]
oh wait, it's handled here (missed the bitmask on line 66)

https://github.com/monome/norns/blob/master/matron/src/device/device_midi.c#L79

so there needs to be a nested case for 0x0E which just sets len to 0

@okyeron if you can test this with your device that'd be helpful (i don't have any sensing devices)

**Comment by tehn** [comment]
ie, just add

```
case 0x0E:
```

after line 89, recompile, and it should be good?

**Comment by okyeron** [comment]
debugging this active sensing thing further...

It seems norns.midi.event = function(id, data) is receiving `0xfe` byte and then a `nil` byte.

I think what's happening is that there's 2 events being sent from device_midi.c. It looks like there might be a bracket out of place at
https://github.com/monome/norns/blob/39afbf7bd446dcb4021aa6e8195f9ccc4d82f866/matron/src/device/device_midi.c#L105
and it needs to move down to line 119

**Comment by artfwo** [comment]
@tehn @okyeron there's no need to add additional cases for handling the 0xfe status byte, see https://github.com/monome/norns/blob/master/matron/src/device/device_midi.c#L55

the bracket at line 105 is in the right place, moving it to line 119 will cause nothing. adding `continue` operator to https://github.com/monome/norns/blob/master/matron/src/device/device_midi.c#L60 might actually help, but i also don't have any hardware to test that change. @okyeron can you try that instead?

**Comment by tehn** [comment]
@artfwo good find, i clearly didn't go back far enough!

@okyeron i'd suggest getting some debug prints in here to trace what's going on

**Comment by okyeron** [comment]
@artfwo the bracket change totally fixes the issue for me. 

I added some debug in testing and the `if (byte >= 0xf8) {` events were executing event_post(ev);  twice

**Comment by okyeron** [comment]
FWIW - this should be testable with any midi hardware sending clock or start/stop messages.

**Comment by okyeron** [comment]
Getting back to the original pull request... Are the changes above good?

And then... Would it be reasonable to add the following items to `Midi.to_msg()` instead of just being set to type "other" ?
(and then it could also ignore active sensing from "other" when printing out the whole event stream as well)

i think this might be useful for someone working outside of the beatclock library - or for making beakclock more readable.

```
  -- start
  elseif data[1] == 0xfa then
    msg.type = "start"
  -- stop
  elseif data[1] == 0xfc then
    msg.type = "stop"
  -- continue
  elseif data[1] == 0xfb then
    msg.type = "continue"
  -- clock
  elseif data[1] == 0xf8 then
    msg.type = "clock"
  -- ignore active sensing
  elseif data[1] == 0xfe then
    -- do nothing
```

tehn:
this is now probably fine as-is. i can't think of anything that would send a note-on with vel 0 expecting an actual note-on.

we definitely can't add more msg.type categories without much more modification to `midi.lua`, look there for further ramifications

okyeron:
>more modification to midi.lua, look there for further ramifications

(I hope I'm not being too thick here but...) I'm not sure I see those ramifications since it just sets the msg.type for those incoming bytes. Unless you mean that we would then need to support those same types being **sent** out? (which I could look at as well)

okyeron:
@tehn just in case you did not notice since it got buried by the other conversation here... The change for just updating msg.type has already been made: https://github.com/monome/norns/pull/648/commits/f9aa361c3d971b0bf74fc3773078102aa8f1c473

Issue #656, opened 2018-11-25
Accidentally setting engine method breaks engine until matron restart


rwhaling:
While working through the studies, I accidentally ran `engine.cutoff = 100` in a script, which I rapidly fixed.  

However, that created a numeric field named `cutoff` in the `engine` table, which means that when I correctly ran `engine.cutoff(100)`, I got a `attempt to call a number value (field 'cutoff')` error.

Surprisingly - the error persisted even after removing the bad assignment, since the `engine` table itself does *not* appear to be re-initialized upon script load - I had to restart .  Huge thanks to @catfact for figuring this out.

After a bit more testing, I've confirmed that this issue persists even after loading a different script, that uses a different engine - I believe due to the way the `engine` object itself persists across loading/unloading particular engines

I think this is a pretty big potential footgun for Lua novices (like me), so worth fixing.  I'm still wrapping my head around how script-loading works, but I think two ways to fix would be:

1. Disable metatable sets by raising an error from `engine`'s `__newindex` method
2. Re-initialize `engine` upon script load, which seems generally like a good idea, but more architecturally invasive.

Small script to repro:

```lua
engine.name = "TestSine"

function init()
  engine.hz(math.random(10)*50+100)
end

function key(n,z)
  -- engine.hz = n * 200 -- comment/uncomment this line to test
  engine.hz(n * 200)
end

function redraw() end
```

PR #657, opened 2018-11-25
Added a __newindex method to protect Engine

rwhaling:
Naive attempt to solve #656 - prevent accidentally messing up the Engine metatable with `engine.cutoff = 100`, e.g.

Still working on getting a local norns dev environment set up, not sure how to properly test this in the meantime.

### DIFF
--- a/lua/engine.lua
+++ b/lua/engine.lua
@@ -101,6 +101,10 @@ function Engine.__index(self, idx)
   end
 end
 
+function Engine.__newindex(self, idx, val)
+  error("attempted to set a new field of Engine; try engine.field(value) instead")
+end
+
 setmetatable(Engine, Engine)
 
 return Engine


tehn:
thanks for identifying this probably-common mistake, and i think the prevention method makes a lot of sense!

curious is @pq has any strong opinions on this approach?

pq:
oh!  i love the idea of guarding norns metatables.  i'm not totally sure this would quite work as-written but will try and carve out some time to play around.

fwiw @rwhaling, we explored something similar in #426; the conversation anyway may be interesting to you.

thanks a million for nudging this.  i totally agree that these accidental over-writes are a major gotcha.

catfact:
yea i did it originally. i had some notion of making things read-only and "private" by hiding in properties table, and it goes well with requiring weird setters for some things.

but i didn't follow through, partially due to having zero lua experience :0

so yea, i think this is an excellent direction, as long as it works and __newindex doesn't need to test for any exceptions - apologies haven't really been looking at the scripting side lately

artfwo:
`__newindex` will only be called for unset keys, e.g. if `Engine.name == nil`, so we need to make sure that we're still able to set the engine name in any edge cases (after script.clear, etc.).

catfact:
sorry, of course you're both right @artfwo and @pq , `__newindex` is only invoked for keys that are nil.

the "proxy" idea is to route get/set metamethods through an empty table. `__index` passes through to the original table, `__newindex` is disabled; since the user sees the proxy and the proxy is empty, `__newindex` is invoked for every set. 

the big caveat is that this breaks iteration methods like `ipairs`.

sketch based on example in lua manual:

read-only module `ro.lua`
```
function readonly (t)
    local proxy = {}
    local mt = {
      __index = t, 
      __newindex = function (t,k,v)
        error("attempt to update a read-only table", 2)
      end
    }
    setmetatable(proxy, mt)
    return proxy
end

Ro = readonly({ foo = function() print("the true foo") end })

return Ro;
```

in use:
```
> ro = require('ro')

> ro.foo()
the true foo

> ro.foo  = function() print('the false foo') end 
stdin:1: attempt to update a read-only table
stack traceback:
	[C]: in function 'error'
	/home/emb/code/scratch/lua/ro.lua:6: in function '__newindex'
	stdin:1: in main chunk
```

pq:
> the big caveat is that this breaks iteration methods like ipairs. which is fine.

actually, there's a trick for that in 5.3.  i *think* we can forward the `__pairs` and `__ipairs` metamethods to the proxied table and it should "just work".  iow:

```lua
function readonly (t)
    local proxy = {}
    local mt = {
      __index = t, 
      __newindex = function (t,k,v)
        error("attempt to update a read-only table", 2)
      end,
      __pairs = function (t) return pairs(proxy) end,
      __ipairs = function (t) return ipairs(proxy) end,
    }
    setmetatable(proxy, mt)
    return proxy
end
```

> the other caveat is that members of members of the table are still mutable unless they also use a proxy.

yeah.  this is a currently unaddressed short-coming of the way we're managing script-written global state too.  we could recurse i guess or protect select members of members but that may be a bridge too far.

tehn:
more complete solution in #663

PR #658, opened 2018-11-25
fix for nil data bytes (active sensing, etc)

okyeron:
bracket out of place -- status bytes 0xF8 and above were sending 2 chunks of event data, second one being nil.


okyeron:
updated

PR #662, opened 2018-12-02
hid update start

okyeron:
I wanted to dig into HID with a keyboard and mouse today and I found it was really confusing.

I thought It might be better to have it's connection method work like grid and midi so I set about rewriting hid.lua

Thus a HID script could be much simpler:
```
local keyb = hid.connect(1)
function keyb.event(typ, code, val)
    print("hid.event ", typ, code, val)
end
```

This is not yet plugged into menu.lua to have hid devices show up in the SYSTEM menus.

I am also getting an error on device removal that I don't understand
```error: double free in device_hid.c```

### DIFF
--- a/lua/hid.lua
+++ b/lua/hid.lua
@@ -61,10 +61,6 @@ function Hid.new(id, name, types, codes, dev)
   return d
 end
 
---- write data to a device
-function Hid:write(msg)
-  hid_send(self.dev, msg)
-end
 
 --- static callback when any hid device is added;
 -- user scripts can redefine
@@ -235,7 +231,6 @@ norns.hid.add = function(id, name, types, codes, dev)
   Hid.devices[id] = d
   Hid.update_devices()
   if Hid.add ~= nil then Hid.add(d) end
-  
 end
 
 --- remove a device
@@ -265,14 +260,12 @@ norns.hid.event = function(id, ev_type, ev_code, value)
   local ev_code_name = Hid.event_codes[ev_type][ev_code]
   --print("norns.hid.event ", id, ev_type_name, ev_code_name, value)
   local dev = Hid.devices[id]
-  if  dev then
-    local cb = dev.callbacks[ev_code_name]
-    if cb then
-      cb(value)
-    else
-      -- print("norns.hid.event unhandled ", id, ev_type, ev_type_name, ev_code, ev_code_name, value)
-    end
-  end
+  --if  dev then
+  --  local cb = dev.callbacks[ev_code_name]
+  --  if cb then
+  --   cb(value)
+  --  end
+  --end
 
     if dev ~= nil then
     if dev.event ~= nil then
@@ -572,14 +565,6 @@ Hid.event_codes[Hid.event_types_rev['EV_KEY']] = {
   [0x107] = 'BTN_7',
   [0x108] = 'BTN_8',
   [0x109] = 'BTN_9',
-  -- BTN_10 through BTN_15 are nonstandard, not in libevdev
-  [0x10A] = 'BTN_10',
-  [0x10B] = 'BTN_11',
-  [0x10C] = 'BTN_12',
-  [0x10D] = 'BTN_13',
-  [0x10E] = 'BTN_14',
-  [0x10F] = 'BTN_15',
-  -- end nonstandard codes
   [0x110] = 'BTN_MOUSE',
   [0x110] = 'BTN_LEFT',
   [0x111] = 'BTN_RIGHT',
--- a/lua/hid.lua
+++ b/lua/hid.lua
@@ -5,37 +5,40 @@ require 'norns'
 norns.version.hid = '0.0.2'
 
 local Hid = {}
+Hid.devices = {}
+Hid.list = {}
+Hid.vport = {}
+for i=1,4 do
+  Hid.vport[i] = {
+    name = "none",
+    callbacks = {},
+    index = 0,
+    attached = false
+  }
+end
 Hid.__index = Hid
 
-Hid.devices = {}
-
---- device-added callback;
--- script should redefine to handle device hotplug events
--- @param device - a Hid
-Hid.add = function(device)
-  print("device added: ", device.id, device.name)
-end
-
---- device-removed callbacks;
--- script should redefine to handle device hotplug events
--- @param device - a Hid
-Hid.remove = function(device)
-  print("device removed: ", device.id, device.name)
-end
 
 -- `codes` is indexed by event type, values are subtables
 -- @tparam integer id : arbitrary numberical index of the device
 -- @tparam string name: device name string from USB
 -- @param types: array of supported event types. keys are type codes, values are strings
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by okyeron** [comment]
Been poking at this more today and device removal seems broken or funky. My old keyboard* shows up as 2 devices (keyboard and mouse types) so I'm guessing part of the problem - and one of those 2 is not getting removed properly.

With a USB mouse, I get the error i mentioned above twice when I unplug. The error is coming from `TEST_NULL_AND_FREE(di->codes[i]);` in `dev_hid_deinit(void *self)` in [device_hid.c](https://github.com/monome/norns/blob/440ad13f44f898ddf957c8a70fa89450ddb6428d/matron/src/device/device_hid.c#L151) but I don't understand enough about what's happening there.

*Apple Extended II with ADB->USB adapter.

**Comment by catfact** [comment]
thanks for looking into this, the feature has been pretty neglected since early-days hacking.

lucky issue 666
https://github.com/monome/norns/issues/666

**Comment by okyeron** [comment]
Just doing some reading/skimming of libevdev docs (to see if there's something about multiple interface devices)...

Query: in `device_hid.c` - in `TEST_NULL_AND_FREE` why is `free()` being called instead of `[libevdev_free()](https://www.freedesktop.org/software/libevdev/doc/0.2/group__init.html#gacd9fe760d15be25fc99ce469034bd78c)`?

**Comment by okyeron** [comment]
Little help?

Should this line https://github.com/monome/norns/blob/1d1b4a669a04c3a79f0748034b47d8418b1e16e1/matron/src/device/device_hid.c#L56 be using `d->dev `and not `dev`?

(like this?)
`if( libevdev_has_event_code(d->dev, type, code) ) `

Also, shouldn't there be a call to `libevdev_free(di->dev);`  in the `dev_hid_deinit` function?
https://github.com/monome/norns/blob/1d1b4a669a04c3a79f0748034b47d8418b1e16e1/matron/src/device/device_hid.c#L151

**Comment by okyeron** [comment]
OK - I think I solved the double free issue - there were event types that got ignored for length, so I queried for the max value for all types and it fixed the null codes.

Still have an issue with my keyboard adding 2 devices when plugged in and then only releasing one or not releasing either (not exactly sure which), but the devices keep multiplying on unplug/plug.

**Comment by tehn** [comment]
apologies for the delay, i'll be able to look at this by the end of the week.

**Comment by okyeron** [comment]
Writing up notes here for future reference:

we have 2 input device nodes:
`/dev/input/event0`
`/dev/input/event1`

[device_monitor.c](https://github.com/monome/norns/blob/1d1b4a669a04c3a79f0748034b47d8418b1e16e1/matron/src/device/device_monitor.c#L236) `handle_device()` sends `dev_list_remove(t, node)` once for each node.

[device_list.c](https://github.com/monome/norns/blob/1d1b4a669a04c3a79f0748034b47d8418b1e16e1/matron/src/device/device_list.c#L95) `dev_list_remove()` gets the request and removes the first node (event0). The second time around (event1)  in `dev_list_remove()` `dev_lookup_path(node)` returns null and does not process any further. My theory is this is because the system has already removed the node `/dev/input/event1` (since it's paired with event0) and `dev_lookup_path` wont resolve. 

Thus `ev->hid_remove.id` does not get set for the second node and it's not removed.

At the moment I am not sure how to fix or work around that. (thinking some more...) perhaps if device_monitor.c was able to determine the device id first, it could send that to a separate dev_list_remove function specifically for HID/event types?

**Comment by okyeron** [comment]
Circling back on this tonight...

Can someone tell me if I'm on the right track here:
The first time device_list.c is called - [this line](https://github.com/monome/norns/blob/1d1b4a669a04c3a79f0748034b47d8418b1e16e1/matron/src/device/device_list.c#L118) sets `dq.head` to NULL. 

The next time this function gets called for the second device removal [line 97](https://github.com/monome/norns/blob/1d1b4a669a04c3a79f0748034b47d8418b1e16e1/matron/src/device/device_list.c#L97) returns NULL because `dq.head` is now NULL in `dev_lookup_path()`

It looks like if I change [line 118](https://github.com/monome/norns/blob/1d1b4a669a04c3a79f0748034b47d8418b1e16e1/matron/src/device/device_list.c#L118) to 
```
if(dq.head == dn) { dq.head = dn->next; }
```
The second device gets removed. (yay!)

My question - Is this OK as is?  Or... does `dq.size` needs to be checked before `dq.head` is set to `dn->next` or to `null` (if size is zero or 1) ?

tehn:
@okyeron has this been working for you?

okyeron:
Yes it has.

catfact:
it doesnt seem like checking size would be a bad idea at all

catfact:
@okyeron sorry this has been languishing for weeks and weeks. i for one have barely had time to use norns hardware and each session is kinda laser-focused on DSP issues. 

the code looks fine to me but i'd like to confirm on hardware. do you maybe have some kind of test script you've used, that we could take for a spin?

okyeron:
I'll see if there's a test script i can dig up in the morning. I've forgotten precisely where this left off, but it was working well for me with a variety of inputs (mouse, keyboard, wacom, etc)

artfwo:
@okyeron i'm currently revising the vport api, the WIP PR is here: https://github.com/monome/norns/pull/698 

would you like to update this PR accordingly? otherwise, i can work on that when done with grid and midi improvements.

okyeron:
FWIW- I'm really fuzzy on all this since I've not really touched norns in a month (waiting for 2.0 things to settle into a place where I could install and test).

@catfact Here's 2 gists of the (janky) sample hid code I was using:
https://gist.github.com/okyeron/c3a5af2c04d8daf32ef290fa9e7923f5
https://gist.github.com/okyeron/f959f8301329e4be8a18b477145c2f39

@artfwo I can take a look at the new vport api, but again - i'm really fuzzy on this right now. 

Is the arc code from that pull the best to use for reference?

tehn:
@okyeron thanks for the gists! i can help test.

artfwo:
@okyeron yes, i'm also working on grid and midi at the moment.

okyeron:
@artfwo I will try to refresh myself on this tonight. If you get done with grid/midi quickly - please go ahead and add the needed changes here

okyeron:
@artfwo - i made an attempt to update hid to the vport changes based on your arc changes and I can't get it working. hid.devices is empty. Device (mouse) connects and vports is populated but then I'm stuck.

artfwo:
@okyeron sorry, i missed the edit for your last comment and pushed my hid module in the vport cleanup branch already. right now i'm going to cherry-pick some of your fixes above into https://github.com/monome/norns/pull/698 and continue working from there.

okyeron:
apologies - I didn't realize those text edits would not send notifications. I'll try to make new comments in future.

the important bit is in the comment: https://github.com/monome/norns/pull/662#issuecomment-447229707 for fixing device_list.c - for devices getting properly removed.

artfwo:
@okyeron thanks, I've cherry picked the relevant commits and will give some testing to removing devices too.

artfwo:
as this has now been merged with https://github.com/monome/norns/pull/698, i'm closing the PR. thanks @okyeron for the matron fixes!

Issue #471, opened 2018-07-17
Cater for moved .sc files on update


antonhornquist:
It seems like moved sc files (such as me moving CroneGenEngine.sc from the sc engine folder to  ”~/dust/lib/sc/abstractions”) might mess up/cause sclang discrepancies in the update process. This might be due to users git cloning https://github.com/monome/dust in between updates, I dunno. Anyhow, is there any way to fix this during update? (update is a process I’m not familiar with)

catfact:
adding/removing/moving .sc files requires running the `norns/sc/install.sh` script (currently). this is actually not really necessary given that i've finally realized that you can add arbitrary search paths for SC class files (see Quark.sc for example)

i also have been wondering about the update process, since some upcoming changes require more than just copying files (e.g. internet connection)

@tehn is there a reason not to have `update.sh` in git?

i see `update.lua` -

tehn:
`update.sh` changes per update... though much is the same.

i'll add it

tehn:
can you clarify what you mean by adding arbitrary search paths?

ref: https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Common/Quarks/Quark.sc

trying to comprehend how we can have .sc files scattered around dust, ie, encapsulated within a "project" re #446 

would it be possible to just have the search path include a whole subtree ie `~/dust/scripts/`?

catfact:
> would it be possible to just have the search path include a whole subtree ie ~/dust/scripts/?

Yes I think so

antonhornquist:
> include a whole subtree ie `~/dust/scripts/` ?

yeah that's what i'm thinking

tehn:
ok, so my experiments using EXPORT to set paths failed. what's the magic word?

but i'm also confused about what the actual problem is. i had success changing `install.sh` to link `/dust` instead of `/dust/lib/sc` and i tried moving .sc files around the dust repo, without re-executing `install.sh` and all seems fine. was the earlier problem @antonhornquist  reported about duplicate files?

catfact:
sorry i linked the wrong file
https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Common/Quarks/Quarks.sc#L219

```
*link { |path|
		path = path.withoutTrailingSlash;
		if(LanguageConfig.includePaths.includesEqual(path).not, {
			path.debug("Adding path");
			LanguageConfig.addIncludePath(path);
			LanguageConfig.store(LanguageConfig.currentPath);
[...]

PR #664, opened 2018-12-03
folder restructure, explode dust

tehn:
this is the first step for a new approach to `dust`

- `dust` is no longer a git repo. it's just "the user folder" and still lives at `~/dust/`
- structure is "project-based" which means any subfolder of `dust` is a "project" with its own `data` subfolder (where psets/whatever are saved)
- top-level scripts are allowed ie `~/dust/earthsea.lua`
- menu SELECT is now flat, without hierarchy. entire folder is scanned, and all `lua` files are shown. `data` and `audio` folders are skipped and not traversed. files with the same name as their enclosing folder are represented by that single name (ie, `awake` instead of `awake/awake` for `~/dust/awake/awake.lua`)
- BREAKING: paramset read/write functions now take only a single filename, not a path to be appended to the `data_dir`. `norns.state.path` is established at script load, which is the path of the file loaded. the data dir is `/data/` appended to this path, and fully handled by the functions. (previously there was the requirement to specify (if desired) a subfolder for data files... this will be a big cleanup point for existing dust scripts)
- MAJOR and BREAKING and NEEDS-EVAL: all `lib` scripts from `dust` moved to `/norns/lua/user/lib` and the path has been modified so `require 'musicutil.lua'` will work. _engine_ libs moved to `/norns/lua/user/engines/` and `require 'ack'` should continue to work. **however** i flattened both folders and they are not organized any longer by contributor. basically my feeling is that anything contributed to the main `norns` repo should be essentially communal ie up for collaboration (within the norms of open source). we should notify any authors of this change and get confirmation for inclusion. any subfoldering should be functional identification. (and some is needed).
- i created a folder `norns/lua/core/` which includes the system files, so `require 'core/metro'` etc
- all engines moved to `norns/sc/engines/` so now they're in the `norns` repo
- `SYNC`, `UPDATE`, and `LOG` removed entirely from menu system
- some `matron` changes so be sure to re-run `./waf`

### forward

this build will be a little cranky if run with an existing `dust` repo. back it up, and make a new folder.

`~/dust` with subfolders
`data`
`audio`

and then throw some scripts into the root, or named subfolders if you want/need separate data/audio/lib folders. a subfolder also allows for multiple versions to be stored ie

```
/dust/awake/awake.lua
/dust/awake/more.lua
/dust/awake/not.lua
```

show up on the menu as `awake`, `awake/more`, `awake/not` and share the same data/audio/lib folders.

**!!!** you can do a git clone into to `dust` folder with your own repo with a script. future feature of matron to clone/pull repos via pasted url

so: until then, managing `dust` means a lot of SFTP, which is really ok. people don't have to wait for official updates to gather/update/share scripts.

all comments appreciated!

### todo

- [x] "local" sc file inclusion. fixes https://github.com/monome/norns/issues/471
- [x] `include(something)` function so scripts can `do_file(norns.state.path .. '/lib/something.lib')` in `init` (basically a private lib)
- [x] exclude `lib` folders from SELECT menu tree builder
- [x] move `fileselect` and `textentry` to "public" `norns/lua/user/lib`
- [x] collision, `formatters.lua` @antonhornquist  vs. `meformatters.lua` (renamed from `formatters.lua`) @markwheeler --- do we have a vision to consolidate any of this?
- [x] @catfact there are some emb engines copied over, do you want to delete them?
- [x] `require 'lib/tabutil'` should be `require 'tabutil'` for clarity, even though the former works
- [ ] bug with paramset saving for many-folder-depth scripts (test with scripts/tehn/awake)
- [ ] **DOCS** !!!

---

update: changed some folders re @pq suggestion. changes reflected above.

---

note: `~/.local/share/SuperCollider/Extensions` needs to be cleaned out for this to work. `rm -rf ~/.local/share/SuperCollider/Extensions/*` and then `install.sh` from `norns/sc`


markwheeler:
> collision, formatters.lua @antonhornquist vs. meformatters.lua (renamed from formatters.lua) @markwheeler --- do we have a vision to consolidate any of this?

I originally made my own formatters lib in case others weren't keen on my particular way of rounding etc which may not be the most universal. I also wanted to be able to use the functions for just formatting numbers as well as params. I'm open to combining if @antonhornquist thinks it makes sense? If we were to then I guess we'd want all the formatting functions to have both a param and number version for consistency.

tehn:
general suggestion for user scripts:

have the first comment line include the name + version.

```
-- awake 1.2.0
```

as for non-git scripts tracking versions will be essential for users who are collecting scripts.

tehn:
good ideas.

it's probably nonsense to separate out `norns` and `menu` etc. i can just push it all into `core` like before.

background on `config.lua`, it's designed so that paths could be set differently but you're right that we could just have this be the top of `norns.lua`... though a separate file makes it easy to find, and `norns.lua` is currently a mess and needs to be cleaned. here's where weaver launches the lua files: https://github.com/monome/norns/blob/master/matron/src/weaver.c#L325

i like:

```
lua/
  core/
  user/
    engines/
    lib/
```

will push an update to fit this

tehn:
currently pset/pmap save/load is disjointed due to integrating global state vars (current script folder)... the result being system.pset fail (and get written weird places).

given the system audio pset will is scheduled to get integrated into the main param menu/pset, this problem will disappear, so perhaps i should get going on that...

tehn:
vu's working. what's a sensible scaling? amp to db and then -64 up to 0?

https://github.com/monome/norns/pull/671/commits/2e641dcad1081b1b01b0e433677f9e9a0321bdea#diff-addc4bc2f21cbb4bdbc5d8ac3cfc3000R62

catfact:
that works. i usually like a linear region at the bottom. b/c with that you can set it up so that zero implies zero amplitude, and >0 pixels implies >0 amplitude.

tehn:
do you have example s of this is any other repo?

catfact:
eh, nothing very tidy, here's how i did the "reverse taper" in supercollider before:
https://github.com/monome/dust/blob/master/lib/sc/abstractions/AudioTaper.sc

if you just leave it with dB scaling and -64db floor i can go in and replace with a LUT

Issue #663, opened 2018-12-02
add a util function for making readonly table proxies


pq:
follow-up from the discussion in https://github.com/monome/norns/pull/657#issuecomment-443471472, a function that creates immutable tables would be really handy.

something like this, minimally, may work:

```lua
function readonly (t)
    local proxy = {}
    local mt = {
      __index = t, 
      __newindex = function (t,k,v)
        error("attempt to update a read-only table", 2)
      end,
      __pairs = function (t) return pairs(proxy) end,
      __ipairs = function (t) return ipairs(proxy) end,
    }
    setmetatable(proxy, mt)
    return proxy
end
```

PR #667, opened 2018-12-03
engine table protection

pq:
a crack at guarding tables generally and `engine` in particular.

* adds `readonly` API to `tabutil`
* wraps `engine` import in `startup` with a read-only proxy

image

closes: #663; fixes: #656 

/cc @tehn @catfact @artfwo

### DIFF
--- a/lua/startup.lua
+++ b/lua/startup.lua
@@ -1,5 +1,8 @@
 -- STARTUP
 
+tab = require 'tabutil'
+util = require 'util'
+
 require 'math'
 math.randomseed(os.time()) -- more random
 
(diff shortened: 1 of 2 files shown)


pq:
fwiw, even with `error(msg, 2)` i can't for the life of me get [error level setting](http://www.lua.org/manual/5.3/manual.html#pdf-error) right and so we've got less than awesome stack traces at the moment.

```
lua: stdin:1: 'hz', a read-only key, cannot be re-assigned!
stack traceback:
	[C]: in function 'error'
	/home/we/norns/lua/tabutil.lua:196: in metamethod '__newindex'
	stdin:1: in main chunk
```

ideally it'd point to where in source the set was attempted...

i wonder if this has to do with the fact that scripts are executed in a `dofile`?

PR #670, opened 2018-12-07
Feature/network manager mods

tehn:
added `listselect` and made sub-menu selections for each action, so ENC3 scrolling isn't so awkward.

considering a UI change for the wifi screen:

```
status:
network: (network name)
ip: 192.168.1.10
strength: -64dBm

OFF    HOTSPOT   CONNECT    ADD    DEL
```

where the selection is horizontal at the bottom?

### DIFF
--- a/lua/listselect.lua
+++ b/lua/listselect.lua
@@ -8,6 +8,7 @@ function ls.enter(list, callback)
   ls.pos = 0
   ls.callback = callback
   ls.done = false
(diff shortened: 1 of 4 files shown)


tehn:
ok i implemented change suggested above.

tehn:
FIX:

- [ ] bug with paramset saving for many-folder-depth scripts (test with scripts/tehn/awake)

PR #671, opened 2018-12-12
[wip] new oracle/weaver and lua lib mods

tehn:
matron glue for new crone

### DIFF
--- a/matron/src/oracle.c
+++ b/matron/src/oracle.c
@@ -1,3 +1,75 @@
+/* TODO
+ /poll/start/cut/phase []
+ /poll/start/vu []
+ /poll/stop/cut/phase []
+ /poll/stop/vu []
+ /set/enabled/compressor [f]
+ /set/enabled/cut [if]
+ /set/enabled/reverb [f]
+ /set/level/adc [f]
+ /set/level/adc_cut [f]
+ /set/level/aux_dac [f]
+ /set/level/cut [if]
+ /set/level/cut_aux [f]
+ /set/level/cut_cut [iif]
+ /set/level/dac [f]
+ /set/level/ext [f]
+ /set/level/ext_aux [f]
+ /set/level/ext_cut [f]
+ /set/level/ins_mix [f]
+ /set/level/in_cut [iif]
+ /set/level/monitor [f]
+ /set/level/monitor_aux [f]
+ /set/level/monitor_mix [if]
+ /set/level/tape [f]
+ /set/pan/cut [if]
+ /set/param/compressor/attack [f]
+ /set/param/compressor/gain_post [f]
+ /set/param/compressor/gain_pre [f]
+ /set/param/compressor/ratio [f]
+ /set/param/compressor/release [f]
+ /set/param/compressor/threshold [f]
+ /set/param/cut/fade_time [if]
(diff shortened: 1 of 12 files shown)


tehn:
- [ ] VUs (can't seem to get matron to receive OSC from new-crone)
- [ ] test cut-phase poll
- [ ] figure out `audio.lua`, if it should have `effects` and `tape` and `cut` combined
- [ ] hook up tape functions properly in `menu.lua`
- [ ] TAPE position counter (just use a metro. for playback, get file length first)

tehn:
@catfact i think this is ready to merge if you'd like to glance over it.

i didn't figure out what's going on with `crone`->`matron` poll reporting (OSC doesn't seem to hit `matron`), but i'll keep working on it.

catfact:
> i didn't figure out what's going on with crone->matron poll reporting (OSC doesn't seem to hit matron), but i'll keep working on it.

my bad, i committed with this flag switched off for debugging with sclang. 
https://github.com/monome/norns/blob/feature/new-crone/crone/src/OscInterface.cpp#L44

it should work now

port numbers should really be program arguments or something

catfact:
oh... nvmd? you were using the correct port? then i'm not sure whats up with that

matron/src/oracle.c
```diff
@@ -484,136 +496,141 @@ void o_restart_audio() {
...
-    lo_send(crone_addr, "/set/level/ins_mix", "f", value);
+void o_set_cut_param(const char* name, int voice, float value) {
+    static char buf[128];
+    sprintf(buf, "/set/param/cut/%s", name);    
+    lo_send(crone_addr, buf, "if", voice, value);
 }
 
+void o_set_level_input_cut(int src, int dst, float level) {
+    lo_send(crone_addr, "/set/level/in_cut", "iif", src, dst, level);
+}
```
catfact: i haven't really verified that it actually works to "overload" an OSC method like this; it should probably just be a separate command.

matron/src/oracle.c
```diff
@@ -484,136 +496,141 @@ void o_restart_audio() {
...
+} 
+
+
 
-/// FIXME: doesn't need channel count?
-// monitor mix -> aux level (stereo!)
-void o_set_aux_fx_input_level(int channel, float value) {
-    (void) channel;
-    lo_send(crone_addr, "/set/level/monitor/aux", "f", value);
+
```
catfact: NB: i made several arbitrary decisions about OSC paths, which you should feel free to change. (e.g. if it seems clearest if paths match lua / oracle method names.)

one decision was to refer to effects by function and not by their place in the chain. this is because the parameter names are baked into the paths, and particular param names only make sense for particular effect types.

tehn: renamed `aux` to `rev` and `ins` to `comp`

feels good to me?

https://github.com/monome/norns/pull/671/commits/4c2eab67025c7b04f00061db2b83d57dc7faae28

matron/src/oracle.c
tehn: this separates them:

https://github.com/monome/norns/pull/671/commits/80c48ee372bd0f0f3bd9740f1c2382a8349d9bfc

PR #674, opened 2018-12-22
Add start/stop/continue/clock helpers

okyeron:
Update to previous midi fixes.

Adds 
- midi send helper functions for start, stop, continue and clock
- `Midi.to_msg(data)` conversions for those types as well.

I did not add `to_data` conversions for these as they have no data so just the send helper function can be used.

### DIFF
--- a/lua/midi.lua
+++ b/lua/midi.lua
@@ -139,16 +139,23 @@ function Midi.connect(n)
       d.send{type="channel_pressure", val=val, ch=ch or 1}
     end
   d.start = function()
-      d.send{0xfa}
+      d.send{type="start"}
     end
   d.stop = function()
-      d.send{0xfc}
+      d.send{type="stop"}
     end
   d.continue = function()
-      d.send{0xfb}
+     d.send{type="continue"}
+      --d.send{0xfb}
     end
   d.clock = function()
-      d.send{0xf8}
+      d.send{type="clock"}
+    end
+  d.song_position = function(lsb, msb)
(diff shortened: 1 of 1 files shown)


lua/midi.lua
```diff
@@ -138,6 +138,19 @@ function Midi.connect(n)
   d.channel_pressure = function(val, ch)
       d.send{type="channel_pressure", val=val, ch=ch or 1}
     end
+  d.start = function()
+      d.send{0xfa}
```
tehn: readability (less magic numbers) would be

```
d.send{type="start"}
```

since you already created a `type` for `start`

etc for others...

okyeron: oh - nice. I'll update

okyeron: Hmm... in that case wouldn't the magic numbers need to be included in `Msg.to_data` instead?

okyeron:
updated for comments above - it did however need to have the magic numbers added to `Msg.to_data` instead - but now it matches the other code better.

Also included types for `song_position` and `song_select`. I'm not sure if song position is 100% but song select well works with my TR-09

tehn:
looks good.

are `song_position` and `song_select` in the midi standard? or are these commands specific to the TR-09?

okyeron:
Yes - those are in the [MIDI spec](https://www.midi.org/specifications-old/item/table-2-expanded-messages-list-status-bytes). The only thing really missing at this point is SysEx/End-of-SysEx (bytes 0xF0, 0xF7), but I wasn't sure how to handle the data from that.

PR #675, opened 2018-12-24
update filename of tape recordings to aiff extension

### LINKED ISSUE #673: tape .wavs are actually AIFF (some programs will fail to read)
Running latest master as of [this commit](https://github.com/monome/norns/commit/a0237c8d14c8103dbcb28da3646a3f07e3124f54)

When I use SFTP to get a wav out of tape recorded on norns and try to drag it over to ableton, I get the message "could not be read.  it may be corrupt or not licensed"

Here is a tape wav that exhibits the issue: [0005.wav.zip](https://github.com/monome/norns/files/2701494/0005.wav.zip)

Apple preview plays it fine

jlmitch5:
fixes #673

@catfact is this all that is needed?  I grepped .wav and that was the only use I could find

### DIFF
--- a/lua/menu.lua
+++ b/lua/menu.lua
@@ -1406,7 +1406,7 @@ m.key[pTAPE] = function(n,z)
     else -- REC CONTROLS
       if m.tape.rec.sel == TAPE_REC_ARM then
         tape_diskfree()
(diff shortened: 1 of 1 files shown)


catfact:
that will do if AIFF is ok

PR #683, opened 2018-12-31
Cut phase

tehn:
glue for cut phase

not sure i understand the args? no matter the second arg i get a firehose of data.

test using `tehn/halfsecond` then call

`audio.poll_start_cut_phase(0,1)` for example


catfact:
the poll seemed to send quantized phase as expected when i tried it last. can take a look later tonight

PR #685, opened 2018-12-31
add softcut master level to mixer client

catfact:


catfact:
address #685

catfact:
Note that I haven’t tested this at all yet!

PR #694, opened 2019-01-15
Lua intonation

catfact:
added a small lua lib with some just-intonation tables.

should also add:
- [ ] some glue methods to map midi notes given fundamental, facilitating integration with scripts that use `musicutil`
- [ ] would like to add a full-on Scala parser (the tuning format, not the language)... time permitting

the larger gamuts are useful for playback rate knobs w/ softcut


lua/user/lib/intonation.lua
```diff
@@ -0,0 +1,285 @@
...
+   }
+end
+
+-- finally, jeff snyder's full 168-tone gamut
+-- see his disseration for reference:
+-- http://scatter.server295.com/full-dissertation.pdf
+
+function JI.gamut()
+   return {
+      3  1/1 ,
```
tehn: is this `3` a typo?

catfact: yep

PR #695, opened 2019-01-15
Softcut lua

catfact:
kinda WIP interface for softcut in lua, moves things from audio.lua 

also more importantly adds Control factory, which gives script writers the correct parameter ranges without having the look them up.

the WIP part is to also have some sort of convenience methods to add params with appropriate setters.


tehn:
ah! i should've PR'd my work-in-progress from last week to save you some time

https://github.com/monome/norns/compare/dev...lua-softcut?expand=1

but i'd rather merge yours first as it has more functionality. thanks for getting this started.

tehn:
oh dang you started this two weeks ago :)

PR #698, opened 2019-01-20
vport api overhaul

artfwo:
input/output API changes:

- use colon syntax when sending data to peripheral devices (`g:led`, not `g.led`)
- use dot syntax for now when assigning callbacks (`g.key = function(x, y, s) ... end`)
- `grid.event` has been renamed to `grid.key`

hid usage example:
```
local point_x = 0
local point_y = 0

h = hid.connect(2)

hid.event = function(type, code, value)
  if type == hid.types.EV_REL then
    if code == hid.codes.REL_X then
      point_x = point_x + value
    elseif code == hid.codes.REL_Y then
      point_y = point_y + value
    end

    print(point_x, point_y)
  end
end
```

### DIFF
--- a/lua/core/arc.lua
+++ b/lua/core/arc.lua
@@ -15,8 +15,8 @@ for i=1,4 do
   Arc.vports[i] = {
     name = "none",
 
-    delta = function() end,
-    key = function() end,
+    delta = nil,
+    key = nil,
 
     led = function() end,
     all = function() end,
@@ -143,8 +143,8 @@ end
 --- clear handlers
 function Arc.cleanup()
   for i=1,4 do
-    Arc.vports[i].delta = function() end
-    Arc.vports[i].key = function() end
+    Arc.vports[i].delta = nil
+    Arc.vports[i].key = nil
   end
 end
 
--- a/lua/core/arc.lua
+++ b/lua/core/arc.lua
@@ -14,14 +14,15 @@ Arc.vports = {}
 for i=1,4 do
   Arc.vports[i] = {
     name = "none",
+    device = nil,
 
     delta = nil,
     key = nil,
 
-    led = function() end,
-    all = function() end,
-    refresh = function() end,
-    segment = function() end,
+    led = function(self, ...) if self.device then self.device:led(...) end end,
+    all = function(self, ...) if self.device then self.device:all(...) end end,
+    refresh = function(self, ...) if self.device then self.device:refresh(...) end end,
+    segment = function(self, ...) if self.device then self.device:segment(...) end end,
   }
 end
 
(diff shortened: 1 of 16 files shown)


tehn:
this looks great, just reading through. i'll test it out on hardware.

basic summary of changes:

- devices auto-assigned in ascending order, only one device in the vport list (no multiples)
- syntax for `led` etc methods use colon

did i miss anything?

artfwo:
right. additionally there's no menu logic yet to make sure you cannot choose same device for 2 different vports. i can work on that after updating other device modules.

tehn:
tested, working well! will make a note that the syntax needs to be updated in the docs.

okyeron:
Circling back to HID today. Questions about the changes to codes and the move codes/types to `hid_events.lua`

Previously you could reference the hid.event_codes by code number and now this is missing - so for example the event below returns the code for a key press like "30" for 'KEY_A'
```
local keyb = hid.connect(
function keyb.event(typ, code, val)
    print("hid.event ", typ, code, val)
end
```
`returns "hid.event 	1	30	1`

Now I need some kind of map table/array to know that code is KEY_A (or just "A" perhaps). [Previously](https://github.com/okyeron/norns/blob/9e40142cf7cc20669cb7df5684e8bf8827a3f329/lua/hid.lua#L307) this was in `Hid.event_codes[Hid.event_types_rev['EV_KEY']]`

It would be handy to me for keyboard parsing to have that previous array back or create a new one with standard keys mapped.

What's the best approach here? Am I missing something that would return what I need?

FWIW doing `if code == hid.codes.KEY_A` seems problematic for the entire keyboard.

artfwo:
Sorry, i do not understand the problem. If you need to map letters to keys, you can create a dictionary in the script, i.e. ```keycodes = { hid.codes.KEY_A = "A", ... }``` and retrieve the letters using `code` as the key, is that what you're trying to implement?

okyeron:
I was wanting to get back functionality that was previously there (which I was using in my previous dev/test scripts). Returning the text value "KEY_A " from the code "30". Which is the reversed version of what's in `hid_events.lua` now.

I started writing a possible solution (adding to `hid_events.lua`):
```
hid_events.keys = {}
hid_events.keys[hid_events.types.EV_KEY] = {
  [0] = 'KEY_RESERVED',
  [1] = 'KEY_ESC',
  [2] = 'KEY_1',
  [3] = 'KEY_2',
--etc
```
then getting a value with `hid.keys[typ][code]`

I suppose creating a dictionary as you've suggested gets the job done as well. Is that better than my "keys" solution? (attempting to provide flexibility to script writers here)

If so - should such a dictionary be a part of `hid_events.lua` or implemented as a user library?

artfwo:
No, that shouldn't be part of hid_events.lua, which is semi-autogenerated from kernel headers and is best left untouched. You can also omit types from your keys dictionary, unless you want to map other types of events for a specific use-case, i.e. it's enough to have `keys[code]` and then just check the type in the event callback.

Generally, in the current state, the API is flexible enough to allow mapping events to strings, as in your example above, but keys can also be mapped to different strings ('A', 'B', 'C'), midi notes, CC's or anything else, so it's best to avoid forcing any particular mapping here.

okyeron:
(using new HID) here's a first pass at a keyboard library and a test script - implemented upper and lowercase text printing on screen like a terminal

lib:
https://gist.github.com/okyeron/eaad77e4d484694526ac6ae40f7b6120
script:
https://gist.github.com/okyeron/1c77c4e9822903bac135a98156acd735

My thought is to eventually make it possible to do keyboard text entry for wifi passwords!

tehn:
looks good.

i suspect that plugging an keyboard in simply for wifi password entry isn't going to be the strongest use case-- if the knob interface is that difficult, i suspect more people will readily have a laptop at hand, and you can simply connect via serial and run `tui` which gives a full-fledged wifi-connect system familiar to normal desktop wifi ui's.

that said, there are a lot of creative uses for keyboards as script controllers... i've seen some fun musical uses in the past.

okyeron:
Heh... yeah, the wifi pw entry thing is an edge case, but I thought I might give it a try.

Of coures... the *real* goal would be to implement ZORK and have norns provide a spooky generative soundtrack :)

okyeron:
Working on my keyboard library (on RasPi ) and when plugging in my HDMI monitor I noticed the command prompt was still getting all these keystrokes and trying to login.

Does norns already disable the keyboard tty somewhere? And could I do that on RasPi via a lua script (without affecting its HID input)? Or should I just not worry about it. :)

artfwo:
Yes, norns has console enabled on ttyAMA0. I suppose for you case, it is best to do that outside of a script and even matron, i.e. tty configuration, but make sure you have some means of backup means of logging into the system.

Issue #697, opened 2019-01-18
tape play/rec CPU usage too high


tehn:
on new-crone. noticed the CPU spike as clicks get into the playback and recording.

i wasn't seeing this before and so far i'm not sure what changed.

with `htop` open (for easy sub-process visibility) the cpu spike can be seen right after executing:

```
audio.tape_play_open("/home/we/dust/audio/tape/0003.aiff")
audio.tape_play_start()
```

@catfact are you able to replicate this?

catfact:
> i wasn't seeing this before and so far i'm not sure what changed.

before what?

catfact:
i don't see a spike but i do see a lot of CPU usage for playback, more than there should be. enough to push one core near 90% 

if you are really seeing a "spike" at the beginning then maybe the actual file preparation is blocking a long time on your system. i haven't seen this yet but will try some more

first step for CPU usage might be to address these FIXMEs in Tape.h
https://github.com/monome/norns/blob/dev/crone/src/Tape.h#L104
https://github.com/monome/norns/blob/dev/crone/src/Tape.h#L312

re: click
unlike the SC implementation, this doesn't have an amplitude envelope. should add one for playback at least. 

are you hearing something more than the hard on/off switch? (i will try listening to file with soft onset)

PR #699, opened 2019-01-24
Crone tape optimize

catfact:
fix a pretty stupid bug where disk worker threads in crone-tape weren't updating their "needs data / data ready" flags. so they were basically busy-waiting on ringbuffer contents, instead of sleeping on their condition variables.

(this took an embarassing amount of work to find, but at least now i have a demangling `perf` on rasbpi!)

i also changed read/write logic to use the largest possible chunk size instead of single frames, which should also speed things up in theory. (the footprint of the actual disk work is vanishingly small though.) 

hopefully this addresses the usage reported in issue #697, though i have not yet addressed the hard click on stop/start. (next PR.)

### DIFF
--- a/crone/src/Tape.h
+++ b/crone/src/Tape.h
@@ -81,6 +81,7 @@ namespace crone {
                 }
 
                 /// libsndfile requires interleaved data. we do that here before pushing to ringbuf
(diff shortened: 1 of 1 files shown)


tehn:
fixes #697

PR #700, opened 2019-01-24
Crone tape envelope

catfact:
added small raised-cosine amp envelope on tape start and stop, for both read and write. 

could use a fixed precomputed LUT for more efficiency

depends on PR #699 

haven't tested on norns HW yet...

### DIFF
--- a/crone/src/Tape.h
+++ b/crone/src/Tape.h
@@ -31,7 +31,7 @@ namespace crone {
         //-----------------------------------------------------------------------------------------------
         //-- base class for sound file access
 
-        class SfAccess {
+        class TapeStreamer {
         protected:
             SNDFILE *file;
             std::unique_ptr<std::thread> th;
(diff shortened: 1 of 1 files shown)


catfact:
hm ok, sorry about that, will make a more proper attempt

catfact:
@tehn do-over

tehn:
dang you are fast! thank you.

tested on hardware. start/stop sound great, totally de-clicked.

i still get a click when a soundfile reaches the end (this is with a soundfile recorded from the same tape system). i figure this click is actually due to the abrupt end of recording, but not sure if it's better to window at recording or just playback?

catfact:
hrrm, yes you're right. but actually it is already being windowed at playback and again this is a logic issue from not correctly dealing with the edge case of half a buffers worth of content at the end 

currently not  automatically windowing the EOF on playback, if you think about it, this is a little more involved. but of course i can do it, we know the count of frames in the file.

catfact:
PS i'm not really fast i just stay up all night

catfact:
ok now EOF is handled more correctly. so if you are playing back something recorded with this system it should already have a fadeout and should sound smooth on playback.

(effectively this means that fadein is raised-cos-squared and fadeout is regular raised-cos, but it's a 50ms fade so who cares.)

haven't yet added auto-fadeout to reader. reader is already keeping track of frames processed and frames in file so this shouldn't actually be a big deal

catfact:
ehh, i guess it's a nice thing to have so i'll just go ahead and add it while i'm here

catfact:
added it. it is definitely happening. can't really hear the difference TBH but it can't hurt (except to add ~ 3 instructions per frame during playback.)

catfact:
fixed and refactored, now instruction hit is basically nil

also added the (very trivial) source for generating the window, in case we actually want to make it a bit shorter

catfact:
easiest to hear click / no click on fadeout with a lowish sinewave, like this
[sine220.aif.zip](https://github.com/monome/norns/files/2796614/sine220.aif.zip)

tehn:
oh i see some .sc files were added back in. was this intentional?

catfact:
woops. no

tehn:
i can remove them now

PR #701, opened 2019-01-24
add delta/key function nil check

tehn:
otherwise a bunch of nil errors are thrown (trying to execute a nil function) when the functions are initially undefined. of course the user script should redefine them, but this will help avoid confusion.

we could alternatively define these functions as `function() end` instead of `nil`. there's a helper function `norns.none` that is a nil function, but this should really be renamed and moved (so probably not used)

similar fixes possibly needed with grid/midi/etc?

### DIFF
--- a/lua/core/arc.lua
+++ b/lua/core/arc.lua
@@ -200,7 +200,9 @@ norns.arc.delta = function(id, n, delta)
     end
 
     if device.port then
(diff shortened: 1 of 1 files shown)


artfwo:
good catch, will update my wip changes accordingly!

artfwo:
good catch, will update my wip changes accordingly!

PR #702, opened 2019-01-24
dust data folders

tehn:
- auto-create `dust/data/(script)` when needed
- auto-copy "preset data" `dust/script/data` to "local data" `dust/data/`

basically the new system distinguishes "project" data from "local" data. think of it as default presets. so a project (git repo usually) can include a bunch of default pset data that then gets copied to the local data folder on first run.

future: restore default pset function? (or just copy stuff with sftp)


tehn:
i'm going to just merge this as i'm solo-deep into this part of the rework

PR #704, opened 2019-01-28
change softcut to 1-indexed

tehn:
also move `softcut` lua user folder/lib


tehn:
tested on hardware, all is well

PR #707, opened 2019-02-02
Crone osx

catfact:
small change to cmake file allowing `crone` to be built on macOS.

still requires a slightly painful amount of setup (compiling libsndfile without flac/vorbis, installing jack, &) - details in `crone/readme-macos.md`. but at least it's usable for development.


catfact:
woop sorry, this is based on `sofcut-improvements`, will make cleaner PR or just merge it into that branch. (changes nothing except `crone/CMakeLists.txt` which isn't even used for norns builds)

PR #713, opened 2019-02-18
2.0 Add display png function

okyeron:
Since export to png was just added I thought it would also be great to be able to load png files to the display.

lua syntax: `screen.display_png(filename, x, y)`


tehn:
@okyeron please PR this against `dev` branch, thanks!

PR #722, opened 2019-02-23
rename arc led function

antonhornquist:
i'm assuming this should be `arc:led`, not `arc:let`


artfwo:
Thanks @antonhornquist!

PR #725, opened 2019-02-28
grid and midi ldoc cleanup

okyeron:
fixes, additions and cleanup of midi.lua and grid.lua

### DIFF
--- a/lua/core/midi.lua
+++ b/lua/core/midi.lua
@@ -85,55 +85,87 @@ function Midi:send(data)
   end
 end
 
+--- send midi note on event
+-- @param integer note : note number
+-- @param integer vel : velocity 
+-- @param integer ch : midi channel
 function Midi:note_on(note, vel, ch)
   self:send{type="note_on", note=note, vel=vel, ch=ch or 1}
 end
 
+--- send midi note off event
+-- @param integer note : note number
+-- @param integer vel : velocity 
+-- @param integer ch : midi channel
 function Midi:note_off(note, vel, ch)
   self:send{type="note_off", note=note, vel=vel or 100, ch=ch or 1}
 end
 
+--- send midi continuous controller event
+-- @param integer note : cc number
+-- @param integer val : value 
+-- @param integer ch : midi channel
 function Midi:cc(cc, val, ch)
   self:send{type="cc", cc=cc, val=val, ch=ch or 1}
 end
 
+--- send midi pitchbend event
+-- @param integer val : value 
+-- @param integer ch : midi channel
 function Midi:pitchbend(val, ch)
   self:send{type="pitchbend", val=val, ch=ch or 1}
 end
 
+--- send midi key pressure event
+-- @param integer note : note number
+-- @param integer val : value 
+-- @param integer ch : midi channel
 function Midi:key_pressure(note, val, ch)
   self:send{type="key_pressure", note=note, val=val, ch=ch or 1}
 end
 
+--- send midi channel pressure event
+-- @param integer val : value 
+-- @param integer ch : midi channel
 function Midi:channel_pressure(val, ch)
   self:send{type="channel_pressure", val=val, ch=ch or 1}
 end
 
+--- send midi start event
 function Midi:start()
   self:send{type="start"}
 end
 
+--- send midi stop event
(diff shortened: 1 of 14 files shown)


lua/core/midi.lua
```diff
@@ -85,55 +85,87 @@ function Midi:send(data)
...
 end
 
+--- send midi song select event
+-- @param integer val : value
 function Midi:song_select(val)
   self:send{type="song_select", val=val}
 end
 
---- create device, returns object with handler and send
+--- create device, returns object with handler and send.
```
artfwo: why some function descriptions have periods and some don't?

pq: right.  i think they should all have a trailing period.

lua/core/midi.lua
```diff
@@ -194,6 +226,7 @@ local to_data = {
 }
 
 --- convert msg to data (midi bytes)
+-- @tparam msg
```
pq: maybe type info too?

lua/core/midi.lua
```diff
@@ -203,6 +236,7 @@ function Midi.to_data(msg)
 end
 
 --- convert data (midi bytes) to msg
+-- @tparam data
```
pq: and type here too?

lua/core/midi.lua
okyeron: "type info"?  I'm not sure whats needed. Do you mean the data type of `msg`?

pq: sorry, yeah.  that's exactly what i meant.

lua/core/midi.lua
okyeron: so something like this:

```
--- convert msg to data (midi bytes).
-- @tparam table msg : 
-- @return hex : midi status bytes - command and value


--- convert data (midi bytes) to msg.
-- @tparam table data :
-- @return table msg : midi message table, contents vary depending on message
```

pq: right.  looking at https://github.com/luarocks/lua-style-guide#documentation, it seems the convention might be `<var> <type>: description` though we should confirm w/ ldoc proper.

pq: ok, hmmm. ldoc has examples that are a little different.  for example:

https://stevedonovan.github.io/ldoc/examples/four.lua.html

i don't feel super strongly to be honest but, like you, think we should just settle on something and then try and be consistent.

i'd be curious what @artfwo, @tehn, @catfact  and others think...

simonvanderveldt: @pq seems like ldoc uses `@ <name> <description>` by default, see chapter "Tags" [here](https://stevedonovan.github.io/ldoc/manual/doc.md.html).
Don't really see if that can include type info.

Apparently they also have an "alternative" `-- type: <name> <description>` way, see "Alternative way of specifying tags".

P.S. Sorry, but I can't link to specific headings/chapters because no anchors are included in ldoc's own docs :man_facepalming: 
Hope that's not something that carries over to the docs generated with ldoc.

pq: aha.  sorry yeah.  it looks like i was confusing tools (and advice) and maybe @okyeron was spot on to begin with.  sorry for the noise...

@simonvanderveldt: i'm curious if otherwise you have any preferences on standards?

simonvanderveldt: No strong opinion, I've done very little in Lua so don't really have a clue :P Mainly work in Python where I'm mostly used to either Google's style or reStructuredText (some examples here https://stackoverflow.com/a/24385103).

Do you/we want to include type information in the docstrings? It could be I've missed it but it seems like the `@ <name> <description>` variant doesn't support type info?

lua/core/midi.lua
okyeron: Fixed periods and added data type in last commit

okyeron:
Pushed an update with improvements to arc,metro,midi,norns,osc files

These changes clear errors `ldoc` throws when run on device.

okyeron:
Also - Can someone clarify - what type are `osc` "args"

tehn:
this looks great. i'd like to merge so others (myself included) can continue from `dev`. feel free to open a new PR with further changes.

PR #728, opened 2019-03-02
remove obsolete supernova refs, use scsynth

antonhornquist:
shell commands threw a jack could not connect error
kill scsynth instead of supernova in `stop.sh`


tehn:
thanks!

i think it may be better to comment these lines out rather then delete them, as i suspect some people may want to attempt using supernova. this was it's just a matter of swapping some line comments.

Issue #739, opened 2019-03-08
TAPE noise


catfact:
> i heard the noise explosion after recording 4 minutes of audio to TAPE.
> successive TAPE recordings are all noise, even if short.

simonvanderveldt:
Can confirm, don't have full steps to reproduce but I made a 10 minute recording of Awake with all "out of the box" settings (using the image provided by @tehn) then played it, put tape's volume up and after +-2 to 3 minutes I got static/noise as well.
It's purely coming from tape, turning tape's volume down turns down the noise.

catfact:
ok, that msg means that for whatever reason, the disk writer thread can't keep up with incoming audio. will 
try more buffering and bigger write chunks or something.

simonvanderveldt:
So that message wasn't there for fun :P The audio written to the file is broken

![image](https://user-images.githubusercontent.com/204286/54058148-575cc000-41f5-11e9-817c-d22078656ed7.png)

okyeron:
I think we also have the reverse of this previous problem: https://github.com/monome/norns/issues/673

The file format appears to be WAVE, but the extension is written as AIFF (causing the file not to open in QuickTime in one case)

tehn:
file extension is a simple fix (sorry!) will PR

catfact:
can't reproduce on x86 with SSD (many attempts of sequential recordings of 10-40+ minutes), using OSC interface directly with crone.

will try on hardware / with menu...

the fact that recordings after the first are borked from beginning, strongly suggests the application state is being left in a bad way. 

hard to imagine a platform discrepancy doing this, easier to imagine some incorrect sequence of commands doing it. should add better error handling. would appreciate help double-checking the glue from menu -> crone.

tehn:
i'll double-check the OSC event order in crone to see what gets printed in the OSC debug

the process is:

- open file (arm)
- start recording
- stop recording

https://github.com/monome/norns/blob/dev/lua/core/menu.lua#L1433

catfact:
i dunno, menu -> module -> weaver -> oracle looks normal.

catfact:
also, user said they got a float32 file. i don't know if this is true; if it is, something is very wrong. i'm requesting 24b fixed PCM from libsndfile, and this is what i'm getting on x86 debian:
```
Input #0, wav, from 'tape-test-7.wav':
  Duration: 00:15:54.47, bitrate: 2304 kb/s
[...]

PR #751, opened 2019-03-12
fix tape (2nd try)

catfact:
issue #739 

re-wrote, doublechecked and cleaned up multithreading bits.

in particular, more correct signaling. tried lockless variant but simply didn't work. lock doesn't block audio thread.  

ringbuffer correctly reset on open

i can no longer reproduce the garbage fire meltdown on norns HW. there are still occasional overruns while writing - these will result in dropouts. no longer printing anything on overruns because we don't actually want to exacerbate glitches with I/O on audio thread.

time permitting, will come back to this with a clean test case and maybe ask for help from LAD community.

### DIFF
--- a/crone/src/Tape.h
+++ b/crone/src/Tape.h
@@ -56,10 +56,10 @@ namespace crone {
 
         public:
             SfStream():
-            file(nullptr),
-            status(0),
-            isRunning(false),
-            shouldStop(false)
+		file(nullptr),
+		status(0),
+		isRunning(false),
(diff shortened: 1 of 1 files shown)


tehn:
apologies, i've been away from hardware the last two days. will test this tonight!

phonk:
I have recorded more than 20 minutes without hiccups, if this helps. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #753, opened 2019-03-12
Remove ugens that have been replaced by native code

simonvanderveldt:
As discussed these can be removed since we're not using them anymore.
**Please review thoroughly** since it could very well be I removed something that's still used.

I did search this repo for usages of the removed ugens which is why I deleted `c/tests/softcuthead-test.scd`.

I also noticed `SoftCutHead` being used here https://github.com/monome/norns/blob/3ca97e31481493afed26e21af61f34d4db8926e9/sc/engines/abstractions/SoftCutVoice.sc#L54
But wasn't sure if the whole `SoftCutVoice.sc` file was still being used or not.


tehn:
this seems good, but i think it'd be really smart (since the softcut supercollider ugen is cool) to preserve these files somewhere? a different repo? @catfact what do you think? or does the new softcut supersede everything?

simonvanderveldt:
@tehn In general I'd say that's what you use a version control system for ;) You can always pull it out of git.
But up to you/ @catfact

tehn:
@simonvanderveldt oh indeed but when good work gets lost in git history i've had a hard time locating it. this happened recently with @catfact 's command-line version of maiden, which i revived after quite a crawl of the history.

catfact:
firstly, i agree with simon, and secondly yeah, the old softcut ugen is too broken; would rather make SC/max/pd wrappers for new-softcut if that is interesting

catfact:
erm, i would say that `BufWrPre` is really useful in general, but maybe that's not relevant.

simonvanderveldt:
We can keep `BufWrPre` if we feel it's useful still?

Normally I wouldn't want to keep unused code around but since parts of this repo are also sort of a library it might make sense.
I'd be more comfortable with it if we'd have a consumer of it in the repo as well so we have a higher chance of catching issues when updating SC for example.

catfact:
i'm totally OK with deleting it. maybe i should look into packaging the more "utility" ugens as quarks.

simonvanderveldt:
> maybe i should look into packaging the more "utility" ugens as quarks.

That sounds like a good idea! We could potentially also contribute them to sc3-plugins?

PR #755, opened 2019-03-14
move engines to dust, folder rework

tehn:
- engines (and engine libs) moved out of main norns repo. they now live in the `we` collab dust repo, and authors are welcome to move them into personal repos. see https://github.com/monome/we/pull/4
- `dust` is now a flat folder. scripts go directly in `dust/` (not in `dust/scripts/`)
- now use `~/audio` instead of `~/dust/audio` for consistency, so that all folders inside `dust` are actually scripts/projects. `system.state` (norns state ie last script loaded) and `system.pset` (norns audio settings) are still stored in the root of `dust/`
- `data` (ie, most often psets) are now stored in local data folders ie `dust/awake/data/` instead of a centralized user data folder (previously was `dust/data/awake/`)

- bonus: cleaned up device menu aesthetics (@simonvanderveldt !!!)


tehn:
just added some `param:write()` and `param:read()` helpers.

`param:read(2)` will read preset 2 of the currently running script.

`param:write(path)` writes to absolute path `path`

`param:read()` is same as `param:read(0)` which reads the default pset (ie `/dust/awake/data/awake.pset`)

PR #757, opened 2019-03-15
lua/core/config.lua: Fix dust scripts package path

simonvanderveldt:
I think this is what it actually should be :roll_eyes: 
This way `require 'we/lib/<somelib>'` works :)

### DIFF
--- a/lua/core/config.lua
+++ b/lua/core/config.lua
@@ -10,7 +10,7 @@ local core = norns..'/core/?.lua;'
 local params = norns..'/core/params/?.lua;'
 local lib = norns..'/lib/?.lua;'
 local softcut = norns..'/softcut/?.lua;'
-local dust = home..'/dust/?.lua;'
+local dust = home..'/dust/scripts/?.lua;'
 
 package.path = sys..core..params..lib..softcut..dust..package.path
 -- print('package.path: ' .. package.path)


catfact:
i haven't been following the convo too well, but i thought the idea was to get rid of `/scripts` and make it flatter

e.g. i saw this example from brian on slack:
```
dust/
  mark_eats/
    molly_demo.lua
    lib/
      molly.sc
      molly.lua
```

simonvanderveldt:
But then `mark_eats` it would we at the same level as `audio` and `data`, which doesn't really make sense, right? For example:
```
dust
├── audio
├── data
├── mark_eats
└── we
```

Tbh I might've missed a change as well, since `dust` is a not a repo anymore it could be I've missed a restructure there? I've still got `~/dust/scripts/we` on my device.

Still the issue remains the same, `we` and `mark_eats` from your example would otherwise be at the same level as `data` and `audio` whereas otherwise they would all live in `scripts` with `scripts` being on the same level as `data` and `audio`. So it becomes/stays:
```
dust
├── audio
├── data
└── scripts
    ├── mark_eats
    └── we
```

catfact:
no, you're right. i don't understand the plan at all.

tehn:
i apologize for the confusion over the last couple days.

@catfact is correct. i've eliminated the `scripts` folder. `dust` is now flat, and only contains script folders. they can include `lib` folders which are hidden from the SELECT menu, which can contain lua libs and sc engines etc.

`data` is internal per-project. there is no "global" data folder.

`audio` was moved to the home folder.

attempted to outline this here: https://github.com/monome/norns/wiki/dust-2.0

tehn:
also with newest pull of `we` and `norns` everything should work as you describe already. let me know if you see breakage somehow.

PR #758, opened 2019-03-16
revert data folder

tehn:
this PR reverts the `dust` structure to 190303

```
dust/
  audio/    -- all user audio
  data/      -- user pset/pmap data, gets tree'd same as script folder
  code/    -- user scripts/projects/etc. (includes 'we'/etc)
```

example script folder

```
mlr/
  mlr.lua    -- main version, shows up as MLR
  mod.lua   -- alt version, shows up as MLR/MOD
  data/
    mlr.pset    -- default pset. script is responsible for copying it/etc
    other.data   -- whatever other data
  lib/
    sub-mlr-lib.lua  -- arbitrary lib, imported via require 'mlr/lib/sub-mlr-lib'
    some-engine.sc   --- engine file
    some-engine.lua   --- engine lib, require 'mlr/lib/some-engine'
```

i'd like to bundle this in to a new beta installer soon so people don't sit with the botched 190314 for long. but please let me know if there are outstanding concerns that need immediate attention. i understand that there are some long-term v3 proposals on the table re non-sc-engines, but right now let's focus on v2.

### DIFF
--- a/lua/core/state.lua
+++ b/lua/core/state.lua
@@ -4,8 +4,8 @@
 state = {}
 state.tape = 0
 state.script = ''
-state.path = _path["dust"]
-state.data = _path["dust"]
+state.path = _path.scripts
+state.data = _path.data
 state.name = ''
 state.shortname = ''
 state.clean_shutdown = false
@@ -13,14 +13,14 @@ state.clean_shutdown = false
 --- read state.lua and set parameters back to stored vals.
 state.resume = function()
   -- restore mix state
-  mix:read(_path["dust"].."system.pset")
+  mix:read(_path.data.."system.pset")
   mix:bang()
 
   -- restore state object
-  local f = io.open(_path["dust"]..'system.state')
+  local f = io.open(_path.data..'system.state')
   if f ~= nil then
     io.close(f)
-    dofile(_path["dust"]..'system.state')
+    dofile(_path.data..'system.state')
   end
 
   -- update vports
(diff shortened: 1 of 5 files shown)


tehn:
removed any data auto-copy as extensively discussed :)

i believe this is now ready to go.

lua/core/state.lua
```diff
@@ -4,23 +4,23 @@
...
+state.data = _path.data
 state.name = ''
 state.shortname = ''
 state.clean_shutdown = false
 
 --- read state.lua and set parameters back to stored vals.
 state.resume = function()
   -- restore mix state
-  mix:read(_path["dust"].."system.pset")
+  mix:read(_path.data.."system.pset")
```
simonvanderveldt: Could very well be I'm misunderstanding something, but wouldn't it make more sense to use `state.data` here instead of `_path.data`?

antonhornquist: both would work - but i agree that minimizing the use of globals established elsewhere is a good thing

lua/core/state.lua
```diff
@@ -4,23 +4,23 @@
 state = {}
```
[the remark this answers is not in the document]
tehn: thanks for finding this! fixed.

lua/core/state.lua
tehn: they are not identical.

`state.data` is dynamic, it's the user data folder for the script ie `/dust/data/awake/`

whereas `_path.data` is always `/dust/data/`

PR #759, opened 2019-03-16
softcut buffer fixes

catfact:
fixed a couple lua/OSC commands:

`/softcut/buffer/clear` : this was broken in `SoftCutClient`
`/softcut/buffer/clear_region_channel` : this was broken in `matron/weaver`

added:
-  `/softcut/buffer/write_mono [sffi]`; args: path, start (seconds), dur (seconds), channel (zero-based)

-  `/softcut/buffer/write_stereo [sff]`; args: path, start (seconds), dur (seconds)

for all the buffer functions, `dur=-1` means to read/write/clear the entire buffer. (or, read the entire file if that is shorter.) (there is an attempt in `OscInterface.cpp` to provide defaults for missing arguments, but it turns out `liblo` doesn't work this way and the handlers aren't called at all. so this should be cleaned up too.)

have **not** yet added lua/weaver glue for write commands, nor tested on norns HW (only on desktop `crone` with raw OSC.) 

(happy to add glue and test sometime next week, unless someone else can get to it first.)

also: 
- both read and write are unnecessarily slow (1 frame at a time.)
- it might be good to have `crone` send some OSC back to `matron` when these operations finish.

### DIFF
--- a/crone/src/OscInterface.cpp
+++ b/crone/src/OscInterface.cpp
@@ -555,6 +555,47 @@ void OscInterface::addServerMethods() {
     });
 
 
+    // FIXME: hrm, our system doesn't allow variable argument count. maybe need to make multiple methods
+    addServerMethod("/softcut/buffer/write_mono", "sffi", [](lo_arg **argv, int argc) {
+        float start = 0.f;
+        float dur = -1.f;
+        int chan=0;
+        if (argc < 1) {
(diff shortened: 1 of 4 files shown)


catfact:
oh, NB: as with `Tape`, output format is 24-bit fixed, 48k, wav. format is arbitrary; chose 24b instead of float as a compromise for the limited disk space on norns.

tehn:
aha `dur=-1` awesome, thank you! and file writing, wow. i can add the glue.

Issue #529, opened 2018-08-31
ControlSpec params don't allow selection of every integer value


markwheeler:
For example, if I'd like to have a param for setting a MIDI value from 0-127 I might try:
`params:add_control("Test", ControlSpec.new(0, 127, "lin", 1, 64, ""))`

However, despite specifying `1` for the `step` argument, turning the encoder skips some values. For example I cannot dial in 66 - however slowly I turn it jumps from 65 to 67.

I would expect setting the `step` argument to anything other than 0 would make that the exact amount the encoder will increment in.

antonhornquist:
this issue is related to `Control` (which handles delta movements, they're too coarse right now - see #431) rather than `ControlSpec` (which just specifies the mapping of a control).

Apart from adding a way to make more fine grained deltas (#431) I agree we should consider stepping.

In the meantime I suggest to use `params:add_number()` which is designed to work with integers.

tehn:
feel free to reopen if there is a more general issue that fine delta control doesn't provide (or the `number` param type)

Issue #431, opened 2018-06-18
fine grained Control


antonhornquist:
Control delta adjustment is too coarse right now, improve this part

tehn:
fixed by #782

PR #782, opened 2019-03-28
WIP: fine parameter adjustment

ngwese:
when experimenting with softcut (playback rate) and at times with other scripts using the generic params menu i've found myself wanting a fine adjustment mode.

this change is a rather basic solve; if key 3 is held down, the delta value from enc 3 is somewhat arbitrarily divided by 20.

some downsides of this approach that i see:
- the fine divisor is arbitrary and results in floating point values being sent to `param:delta(...)`
- parameters may benefit from being able to define course vs fine steps (like octaves vs semitones)
- the fine adjust behavior is only exposed when interacting in the menu but a script would have to roll its own.

if there is general interest in exposing fine parameter control as part of the param abstraction then i'm inclined to close this and look at a general abstraction.

### DIFF
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -469,6 +469,7 @@ m.params.map = {}
 m.params.init_map = function()
   for i = 1,params.count do m.params.map[i] = -1 end
 end
+m.params.fine = false
 
 m.key[pPARAMS] = function(n,z)
   if menu.alt then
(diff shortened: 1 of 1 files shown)


ngwese:
sounds good. as i continue to experiment i'll keep the idea of doing something at the parameter level in  the back of my mind.

tehn:
curious what you mean by "parameter-level" given `param:delta(id, d)` can accept whatever number for `d`?

ngwese:
by parameter level i meant that one extension i think would be useful to have is to support essentially `param:delta(id, d, fine)`. one would send in the raw encoder delta as done previously but specify fine = true when fine adjustment was wanted. the parameter; based on its type/definition, would choose the appropriate fine increment to use.

tehn:
i see now. that's a very smart idea!

antonhornquist:
This probably fixes #529 and #431 ?

Issue #749, opened 2019-03-11
RasPi Framebuffer assignment


okyeron:
In building norns on Raspberry Pi I've run into this issue:

With a Pi 3b+ the onboard HDMI seems to want framebuffer `fb0`. So with my DAC board/hat which uses the same OLED display -  I've had to recompile norns `screen.c` to use `fb1` for the OLED.

I recently found this reference:

>Application software that uses the frame buffer device (e.g. the X server) will use /dev/fb0 by default (older software uses /dev/fb0current). You can specify an alternative frame buffer device by setting the environment variable $FRAMEBUFFER to the path name of a frame buffer device, e.g. (for sh/bash users):

`export FRAMEBUFFER=/dev/fb1`

So perhaps it would be possible to have `screen.c` reference $FRAMEBUFFER - and then set an environment variable in the norns startup scripts - instead of hard coding `fb0`?

tehn:
we could also just have the fb be a command line arg passed to matron, which defaults to fb0.

i'd prefer to avoid more environment setup if possible. but i suppose if $FRAMEBUFFER isn't available we also just use fb0?

ngwese:
i agree with the notion that if matron and co are going to become more configurable that it is done via explicit command line arguments (with the current values as default) as opposed to environment variables.

okyeron:
Addressed with #783

PR #783, opened 2019-03-28
framebuffer select argument

catfact:
added `matron` argument for specifying path to framebuffer used by cairo

defaults to `/dev/fb0`

change it like so:

`matron -f/dev/fb1`

have only tested on norns to make sure nothing is broken. @okyeron (or anyone else requiring this feature) - would be great to get feedback before we merge

addresses issue #749

### DIFF
--- a/matron/src/args.c
+++ b/matron/src/args.c
@@ -9,24 +9,32 @@ struct args {
     char loc_port[ARG_BUF_SIZE];
     char ext_port[ARG_BUF_SIZE];
     char crone_port[ARG_BUF_SIZE];
+    char framebuffer[ARG_BUF_SIZE];
 };
 
 static struct args a = {
     .loc_port = "8888",
-    .crone_port = "9999",
     .ext_port = "57120",
+    .crone_port = "9999",
+    .framebuffer = "/dev/fb0",
 };
 
 int args_parse(int argc, char **argv)
 {
     int opt;
-    while( ( opt = getopt(argc, argv, "e:l:c:") ) != -1 ) {
+    while( ( opt = getopt(argc, argv, "e:l:c:f:") ) != -1 ) {
         switch(opt) {
         case 'l':
             strncpy(a.loc_port, optarg, ARG_BUF_SIZE-1);
             break;
         case 'e':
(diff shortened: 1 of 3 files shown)


okyeron:
Clarification - where would this argument be used?

`matron.sh` ?  Attempting this:
```
#!/bin/bash
./build/ws-wrapper/ws-wrapper ws://*:5555 ./build/matron/matron -f/dev/fb1
```
lead to errors:
```
10.0.1.18 ~/norns $ ./matron.sh 
attempting to bind socket at url ws://*:5555
./build/matron/matron: invalid option -- 'f'
./build/matron/matron: invalid option -- '/'
./build/matron/matron: invalid option -- 'd'
```

catfact:
ah right, hm - i'm not sure if ws-wrapper currently passes arguments correctly since we've never used them before. that might require a fix.

can you try running `matron` with the argument, without ws-wrapper?

okyeron:
Update - I screwed up and didn't have the updated args.c compiled. So I fixed that and now I can run matron.sh without errors... but no display activity (with either `matron.sh` or `~/norns/build/matron $ ./matron -f/dev/fb1`).

catfact:
and you usually get display to work by hardcoding `/dev/fb1`?

okyeron:
Gah - apologies. I'm doing it wrong. I didn't get all the changes. One sec and I'll report back

catfact:
things i have verified:
- with no argument, the default `/dev/fb0` is used and norns continues to work.
- with `-f/dev/fb1`, `screen.c` uses the requested path. norns doesn't work cause this FB doesn't exist.

that's about all i can do from here :man_shrugging:

okyeron:
Sorry - i did something dumb and only grabbed the args.c changes, not the whole branch. 🤦‍♂️ 

Works like a champ!

okyeron:
Last question then - does this need to update anywhere aside from `matron.sh`?

`norns-image/config/norns-matron.service` perhaps?

catfact:
yeah, i guess that's where you would make the change persistent on your system. we don't need to change anything upstream to keep it working here.

PR #786, opened 2019-03-30
Chords in musicutil

markwheeler:
- Added a simple function to generate chords with inversions.
- Added a function to return chords that fit a given root note and key. Took kind of a brute force approach to this - generating a table offline by testing all the chords against all the scales. Would love to hear if there's a smarter way or if it's even that useful?!
- `tparams` for ldocs


tehn:
this looks great!

PR #791, opened 2019-04-01
crone routing docs

catfact:
so far, added tex sources for block diagram of the mixer client.

working on softcut client (less complicated, routing-wise)

rendered PDF looks like this:
[crone-process-routing.pdf](https://github.com/monome/norns/files/3029640/crone-process-routing.pdf)

will add some kind of SVG render as well, for web 

please make suggestions / corrections. labels are easy to change; node layout is less easy but also doable...


tehn:
holy smokes that is crispy!

PR #793, opened 2019-04-02
include() path change

tehn:
many apologies for the breakage!

`include(file)` now includes from `dust/code` (and appends `.lua`)

previously was including from current folder (ie `dust/code/awake`) which made cross-folder sharing difficult (ie for engine libs)


tehn:
> we should get rid of the static HOTSPOT connection definition in norns-image as well

fixed https://github.com/monome/norns-image/pull/73

PR #801, opened 2019-04-05
ext->eng

tehn:
fixed up the routing diagram.

@catfact what do you use for pdf export?


catfact:
`pdflatex`

i dunno if it really makes sense to keep pdf in github. up to you.

here is render with changed label. 
[crone-process-routing.pdf](https://github.com/monome/norns/files/3049340/crone-process-routing.pdf)

honestly i thought the label change would just be a one-liner cosmetic thing on the menu display, but w/e.

PR #802, opened 2019-04-05
Cleanup docs

simonvanderveldt:
Remove redundant interface readme, all of it is part of the main docs now


readme.md
```diff
@@ -1,13 +1,11 @@
 # norns
 
-general [documentation](https://github.com/monome/docs/tree/master/docs/norns) is in [github.com/monome/docs](https://github.com/monome/docs).
+general documentation is in [github.com/monome/docs](https://github.com/monome/docs).
 
-api docs are hosted on [monome.github.io/norns/doc](https://monome.github.io/norns/doc/).
+Lua api docs are hosted on [monome.github.io/norns/doc](https://monome.github.io/norns/doc/).
```
pq: i'd be inclined to lowercase this (for consistency here and elsewhere)... so:

> lua api docs are hosted on ...

pq: but don't feel super strongly

simonvanderveldt: I used the official name 
> Like most names, it should be written in lower case with an initial capital, that is, "Lua".

From https://www.lua.org/about.html

Issue #805, opened 2019-04-07
softcut: buffer-clear commands are slow / stall OSC  thread


catfact:
from lines (https://llllllll.co/t/cranes/21207/8?u=zebra)

 [quote="Dan_Derks, post:8, topic:21207"]
> occasionally when recording (even when on a brand new instance) i get phantom little echos of whatever im recording that appear WHILE i’m recording

> ah. while updating this script, i found that the order of softcut’s buffer-clearing and reset commands seem to matter, but i’m still not entirely sure what order is best. i landed on this version after extensive command juggling and while it has not exhibited this behavior on my machine since then, i think i know what you’re describing. it’s like the playhead somehow gets engaged just a few ms behind the write head, yeah?
[/quote]

i suspect this might be due to the fact that buffer clear commands are executed directly from OSC handler thread, and they are relatively heavy. so could be delaying the execution of subsequent commands or something.

one simple thing to try is just slapping some worker threads in there.

catfact:
pushed branch with worker threads, still basically untested
https://github.com/monome/norns/tree/softcut-buffer-workers

also: we are seeking each frame on buffer read! too slow
https://github.com/monome/norns/blob/master/crone/src/SoftCutClient.cpp#L219

itsyourbedtime:
Not sure if this needs dedicated issue, but i encountered strange behaviour with buffer_clear* -it works, but also breaking recording, so you have to restart the script to get it work again. Or am i missing smth?

catfact:
closing, replaced by #934

PR #810, opened 2019-04-09
[WIP] softcut buffer worker threads

catfact:
looking at issue #805

added worker threads from OSC handler functions for buffer read/write/clear.

this is untested. opening PR in case anyone wants to leave feedback before i can get around to it.


catfact:
woops, this is backwards!

PR #813, opened 2019-04-12
softcut defaults

### LINKED ISSUE #812: softcut reset
https://github.com/monome/norns/blob/master/lua/core/softcut.lua#L153

this needs _way_ more default settings (ie filter)

this function gets called when scripts are loaded.

tehn:
fixed #812 

@catfact i need guidance on `rec_offset` --- not sure what the proper default is. -8 (which i saw later in the softcut.lua) locks `halfsecond.lua` in a small loop. other values manually set seem to break things.


catfact:
Sorry I forgot I changed it to take seconds instead of samples (b/c all other time params take seconds)

-0.0003 or something

tehn:
thanks. all good now. do these defaults look sane?

catfact:
i pushed some small changes. 
- figure you want at least a tiny amount of slew on level changes, and a tiny amount of xfade
- assign voices to alternate buffers
- assign voices to non-overlapping loop regions

catfact:
oh also i bumped up the offset. but it should probably be even more.

the offset's most important job is to keep read and write heads outside of each other's resampling windows. the window is +/- 2 samples for each head at rate=1. so it should be a minimum of 4 samples. but if the rate is higher, the window will be larger. 

you had it at -0.00005, which is ~2.5 samp.
i just bumped it up to >7 samp, which is super arbitrary
for safety, i would perhaps make it ~17 samples for dealing with rate=8.
but i would test this on HW. 

i don't see a real downside to having a 17-sample delay between write and read. 

(i debated whether to expose this parameter at all instead of just hardcode it, but i thought it might be useful as a simple delay? maybe too confusing.)

tehn:
thanks for the clarification!

will test on hardware tonight. I think the offset is useful (theoretically, haven't explored)

catfact:
i have an idea: i'll just have the offset clamped to some small number of negative samples (in crone) if you give it ~0.f seconds.

PR #818, opened 2019-04-17
flip pan

### LINKED ISSUE #817: softcut.pan() is reversed
Working with softcut.pan() has the locations reversed from expected values. I'm guessing 0.0 is left, 1.0 is right - but reality is opposite. As an example, change the pan line in ash/orbit to:

    softcut.pan(i, i - 1)

You will see that the channels are reversed from expected. Probably missed because most examples center the softcut voices.

catfact:
fixes #817 

flipped the pan. so now 1=R, 0=L.

could also change this to bipolar if that's desired.

### DIFF
--- a/crone/src/Bus.h
+++ b/crone/src/Bus.h
@@ -196,8 +196,8 @@ namespace  crone {
                 x = a.buf[0][fr];
                 l = level.update();
                 c = pan.update();
-                buf[0][fr] += x*l*c;
-                buf[1][fr] += x*l*(1.f-c);
+                buf[1][fr] += x*l*c;
+                buf[0][fr] += x*l*(1.f-c);
             }
         }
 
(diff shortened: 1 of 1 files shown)


tehn:
ooh bipolar is attractive to me. 

anyone have preferences?

simonvanderveldt:
`-1 to +1` seems to make sense?
A quick search for SC pan stuff also shows that being used for at least this example http://doc.sccode.org/Classes/Pan2.html

tehn:
learning harder towards bipolar. 0 as center feels elegant.

i know there will be some minor script breakage but it's the easiest fix ever, and still i think only a handful of softcut scripts out there

catfact:
ok will do that. so -1 == hard left, +1 == hard right

while we're at at: there is a linear pan and an equal-power pan. (`Bus.panMixFrom(...)`, `Bus.panMixEpFrom(...)`)

currently, softcut is the only place pan functions are used (iirc). it uses the linear pan law. happy to change this. it makes for a "smoother" experience when panning / mixing, but also means `(50% + 50%).summed` will clip.

Issue #816, opened 2019-04-15
softcut phase offset


tehn:
one feature from the old softcut didn't make it forward: `phase_offset`

the `phase_quant` is always relative to 0 (this makes sense), but i'm looking to do weird things (in mlr specifically) where i want to check phase quantum with an offset.

ie, if phase quantum is 0.4s but the loop i'm interested in looking at starts at 1.0s, things don't line up as the start point is not a multiple of the quantum.

my hack for the old softcut was to add the offset command. will re-attempt to dip into crone a bit later.

catfact:
so, IIUC, you would add an offset before quantizing, like here?
https://github.com/monome/norns/blob/master/crone/src/softcut/SoftCutVoice.cpp#L187

tehn:
yep that'd do it i think

thanks for the line #

tehn:
fixed by #819

PR #819, opened 2019-04-19
softcut phase offset

tehn:
function to add offset to the phase quantization calc.

helpful for loops that don't start at position 0 (or a nice multiple thereof), needed for mlr

(also commented out osc packet prints)

### DIFF
--- a/crone/src/softcut/SoftCutVoice.cpp
+++ b/crone/src/softcut/SoftCutVoice.cpp
@@ -188,7 +188,7 @@ void SoftCutVoice::updateQuantPhase() {
     if (phaseQuant == 0) {
         quantPhase = sch.getActivePhase() / sampleRate;
     } else {
-        quantPhase = std::floor( (sch.getActivePhase() + phaseOffset) /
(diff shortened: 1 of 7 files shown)


tehn:
fixes https://github.com/monome/norns/issues/816

crone/src/softcut/SoftCutVoice.cpp
```diff
@@ -184,7 +188,8 @@ void SoftCutVoice::updateQuantPhase() {
     if (phaseQuant == 0) {
         quantPhase = sch.getActivePhase() / sampleRate;
     } else {
-        quantPhase = std::floor(sch.getActivePhase() / (sampleRate *phaseQuant)) * phaseQuant;
+        quantPhase = std::floor( (sch.getActivePhase() + (phaseOffset * sampleRate)) /
```
catfact: ... here.

tehn: great suggestion, thanks!

PR #827, opened 2019-05-26
tape->cut and tape->reverb routing options

Dewb:
So far this PR
* adds tape->cut and tape->aux level params to the crone mixer, initialized to 0 / -inf dB
* exposes tape->cut and tape->aux as OSC endpoints
* adds tape->cut and tape->aux methods to the Audio module in matron
* adds `rev tape input` menu param to `system > audio`

Tape to reverb works great. There's no direct UI exposure of the tape->cut parameter; I was able to test it by running `local audio = require('core\audio') audio.level_tape_cut(18.0)` in the maiden console. This is pretty fun, it ends up being a back door way of getting recorded material into mlr.

Making the tape->cut default to unity gain would make this step unnecessary, but that seems like a breaking change for people who want to play a backing track unaltered while sampling live input. Not sure if this should be a system level control or up to scripts to manage.

(as discussed here: https://llllllll.co/t/norns-help/14016/1406)

### DIFF
--- a/crone/src/MixerClient.cpp
+++ b/crone/src/MixerClient.cpp
@@ -44,8 +44,6 @@ void MixerClient::process(jack_nframes_t numFrames) {
         float *dst[2] = {static_cast<float*>(bus.tape.buf[0]), static_cast<float*>(bus.tape.buf[1])};
         tape.reader.process(dst, numFrames);
         bus.tape.applyGain(numFrames, smoothLevels.tape);
-        // FIXME: probably want other options for tape playback routing.
-        /// for now, just sum tape to insert bus
         bus.ins_in.addFrom(bus.tape, numFrames);
     }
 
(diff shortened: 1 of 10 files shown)


Dewb:
@catfact thanks for the pointers on this! thoughts on the default tape->cut setting?

crone/src/MixerClient.cpp
```diff
@@ -217,6 +227,7 @@ void MixerClient::SmoothLevelList::setSampleRate(float sr) {
     monitor.setSampleRate(sr);
     cut_aux.setSampleRate(sr);
     ext_aux.setSampleRate(sr);
+    tape_aux.setSampleRate(sr);
```
Dewb: I noticed that none of the ->cut levels have their sample rates set here, is that intentional?

catfact: Nope! It’s an oversight

tehn:
super rad!

my feeling is that the user script should set up the softcut inputs, which includes tape (and engines). so default should be 0?

will be able to look over the code later today

Issue #525, opened 2018-08-29
device monitor arbitrary paths


okyeron:
I'm looking for some sort of enhancement to allow for 3d party tty devices* to be recognized by norns alongside regular monome devices.

`device_monitor.c` has the following
```
// watchers
// FIXME: these names / paths are really arbitrary.
static struct watch w[DEV_TYPE_COUNT] = {
    {
        .sub_name = "tty",
        .node_pattern = "/dev/ttyUSB*"
    },
//..etc
```
* I'm working on a couple of Teensy based devices right now. One is a DIY grid, another an encoder knob box. Both can speak serialosc/libmonome so for testing I've had to comment out the current "tty" node_pattern and replace it with `.node_pattern = "/dev/ttyACM*"`

Thus it would be great if this watcher section would allow for multiple types of tty devices (for when I get around to releasing these diy projects for others to use).

catfact:
at the moment, i guess it would work with just `tty*`. (since we are already checking for usb subsystem as noted in issue #527.) 

otherwise, could replace `fwatch` with regex and use more sophisticated pattern matching.

if you want to use something thats not libmonome-compatible, it gets more involved of crouse.

okyeron:
Did a quick test and changing to  `.node_pattern = "/dev/tty*"` seems to work ok for my DIY grid

PR #839, opened 2019-06-11
crow integration

tehn:
initial crow device integration. (crow is a usb device, eurorack CV interface with onboard lua interpreter). see https://github.com/monome/crow

### status

- crow add/remove works, auto-attached. only one unit is supported at once, though the device lib could do more, it'd require some SYSTEM MENU integration similar to grids/midi/etc with vports.
- command set:

```
-- send arbitrary lua line to be executed by crow
crow.send("print('hello')")
-- set output 1 to 3.0V
crow.output[1].volts = 3.0
-- set input mode to 'none'
crow.input[1].mode("none")
-- query input 1
crow.input[1].query()
-- redefine input event function
crow.input[1].receive = function(n,v) ... end
```

### todo

- [x] `^^` commands are not working ie `^^version` suggesting some some of lower-level comm problem
- [x] crow is having some buffer trouble, ie adjacent characters are sometimes copied over one another
- [x] device detection is _very_ dumb, just claiming `ttyACM?` devices. needs to perform an id handshake before adding them as a device.
- [x] ~~`dev_crow_send` should maybe be threaded as to not block~~ (doesn't seem like a problem)
- [x] incoming data is NOT buffered. if an incoming event is not read before the next comes in, the previous just gets overwritten... see `dev_crow` struct. 
- [x] input/output syntax is different than onboard crow, which uses indexed arrays ie input[1].mode
- [x] output asl not implemented
- [x] sloppy globals. `ret_cv` and `change` should get moved to a table (use a substitution instead of substring when parsing incoming crow messages for ^^ check)
- [x] ansible i2c commands


trentgill:
Regarding the device detection being dumb, i'm doing the same thing in the Max object for crow support, so if it's a lua solution would be great to get that into the crowmax.lua object in crow too.

okyeron:
Using `.node_pattern = "/dev/ttyACM*"` will cause conflicts with teensy based devices connecting to norns. 

see #525

tehn:
crow will have a serial handshake before using the device, but it will attempt this handshake on any path that matches. failure is expected for non-crow devices, and they will then be released.

okyeron:
OK - but as the code reads to me right now in `device_monitor.c` a teensy device like my grid or arc clone would be identified as `.sub_name = "crow"` instead of `.sub_name = "tty"`

Is there a better way to filter those devices at that stage?

tehn:
yes we'll need to find a solution for that. honestly it's difficult considering you're talking about code that isn't in master right now. can you make a more complete proposal for device interaction in https://github.com/monome/norns/issues/525 ?

simonvanderveldt:
Can't we set `.sub_name = tty` when we reach this point?
> failure is expected for non-crow devices, and they will then be released.

okyeron:
Understood. I was just hoping to get this "on the radar" since I saw the note about device detection being "dumb" and grabbing all ttyACM* devices.

I'll try to look at this further and see if I can propose something, but I don't really know C enough to code a solution. 

Seems like checking node_pattern first and then getting more information to determine if it should be tty, crow, etc. would be good.

PR #842, opened 2019-06-15
connect softcut to supercollider

lazzarello:
Needs review. Is this a good idea? Check the diff, it connects the output of softcut to the input of supercollider.

### DIFF
--- a/sc/core/Crone.sc
+++ b/sc/core/Crone.sc
@@ -81,6 +81,8 @@ Crone {
 		//Crone.runShellCommand("jack_connect \"crone:output_6\" \"supernova:input_2\"");
 		Crone.runShellCommand("jack_connect \"crone:output_5\" \"SuperCollider:in_1\"");
 		Crone.runShellCommand("jack_connect \"crone:output_6\" \"SuperCollider:in_2\"");
+		Crone.runShellCommand("jack_connect softcut:output_1 SuperCollider:in_1");
+		Crone.runShellCommand("jack_connect softcut:output_2 SuperCollider:in_2");
 
 		//Crone.runShellCommand("jack_connect \"supernova:output_1\" \"crone:input_5\"");
 		//Crone.runShellCommand("jack_connect \"supernova:output_2\" \"crone:input_6\"");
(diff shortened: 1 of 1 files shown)


catfact:
if you want to create a feedback loop like this, i would at least do it on separate inputs to `scsynth`.

haven't tried it, but i think you can set `server.options.numInputBusChannels` to add an additional input pair, before booting the server, here-ish:

https://github.com/monome/norns/blob/master/sc/core/Crone.sc#L70

sc/core/Crone.sc
```diff
@@ -81,6 +81,8 @@ Crone {
 		//Crone.runShellCommand("jack_connect \"crone:output_6\" \"supernova:input_2\"");
 		Crone.runShellCommand("jack_connect \"crone:output_5\" \"SuperCollider:in_1\"");
 		Crone.runShellCommand("jack_connect \"crone:output_6\" \"SuperCollider:in_2\"");
+		Crone.runShellCommand("jack_connect softcut:output_1 SuperCollider:in_1");
```
catfact: this will create an always-on feedback path from softcut to supercollider, which is undesirable in many situations.

i would at least create an additional, separate input bus pair when starting the scsynth process. but i'm not sure i see this as a big win for functionality.

tehn:
i'm wondering if a better solution might be to have softcut be able to mix into the send-to-engine bus? (by mix i mean toggle what gets sent to the engine... input/softcut/tape)

still feels entirely feedback prone.

catfact:
Yeah that would make sense. It would be easy to add that route with 1 buffer of extra latency. (I think)

And in fact today I came across a forum post from a few months ago asking for this

sc/core/Crone.sc
lazzarello: sorry about that! fixed.

catfact:
Still not a good idea IMO. This needs an actual feedback route in the mixer client that can be turned in and off.

catfact:
sorry @lazzarello but i'm gonna close this. i think it's clear what we need to do to make this feature properly (viz, in a way that doesn't break existing usage of softcut with SC engines that process input.) 

see this earlier PR for a good example of how to add mix routes:
https://github.com/monome/norns/pull/827

it's unfortunately quite a lot more than a 2-line change, mostly because of the many layers of setter indirection.

PR #846, opened 2019-06-19
add engine loading status flag

catfact:
this makes it so that `engine.load()` will skip issuing a load command if there's already one pending. (the function will also return `false` in this case, and `true` otherwise.)

also cleaned up some bad-looking logic in engine-loaded callback, which was conceivably responsible this apparent bug: incomplete command table at script init time, when engine takes long time to load, and no script-level callback is defined.
(https://llllllll.co/t/timber/21407/70)

### DIFF
--- a/lua/core/engine.lua
+++ b/lua/core/engine.lua
@@ -13,6 +13,8 @@ Engine.names = {}
 Engine.name = nil
 -- current command table
 Engine.commands = {}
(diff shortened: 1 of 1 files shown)


neauoire:
Would you prefer `.is_ready`, nstead of `.is_loading`?

PR #852, opened 2019-07-01
first pass at clock.lua ldocs

okyeron:
started on ldoc for clocks based on info at #712

might need additional clarification on data types for each function.

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -29,7 +29,7 @@ clock.sleep = function(...)
 end
 
  
---- schedule a coroutine to sync up with the next fraction of a beat (beats);
+--- yield and schedule waking up the coroutine at beats beat;
 -- the coroutine will pause for the time required to reach a given fraction of a beat;
 -- must be called from within a coroutine started with clock.run.
 -- @tparam float beats : Next fraction of a beat at which the coroutine will be resumed. May be a value larger than 1.
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -27,9 +27,9 @@ clock.sleep = function(...)
   return coroutine.yield(SLEEP, ...)
 end
 
---- suspend execution until the given fraction of a beat is reached in time;
+--- suspend execution until the next fraction of a beat is reached in time;
 -- must be called from within a coroutine started with clock.run.
--- @tparam float beats : 
+-- @tparam float beats : Next fraction of a beat at which the coroutine will be resumed. May be a value larger than 1.
 clock.sync = function(...)
   return coroutine.yield(SYNC, ...)
 end
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -21,7 +21,7 @@ local SLEEP = 0
 local SYNC = 1
 
 
---- schedule a coroutine to sleep and then wake up in s seconds;
+--- yield and schedule waking up the coroutine in s seconds;
 -- must be called from within a coroutine started with clock.run.
 -- @tparam float s : seconds
 clock.sleep = function(...)
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -5,7 +5,8 @@ local clock = {}
 
 clock.threads = {}
 
---- create a coroutine from the given function and immediately run it. 
+--- create a coroutine from the given function and immediately run it;
+-- the function parameter is a task that will suspend when clock.sleep and clock.sync are called inside it and will wake up again after specified time.
 -- @tparam function f
 -- @treturn integer : coroutine ID that can be used to stop it later
 clock.run = function(f)
@@ -19,15 +20,16 @@ end
 local SLEEP = 0
 local SYNC = 1
 
---- pause coroutine execution for s seconds.
--- @tparam integer s : seconds
+--- suspend execution of a coroutine for s seconds;
+-- requires being called from within a coroutine launched with clock.run.
+-- @tparam float s : seconds
 clock.sleep = function(...)
   return coroutine.yield(SLEEP, ...)
 end
 
---- pause execution until the given fraction of a beat is reached in time;
--- must be called from within a coroutine launched with clock.run.
--- @tparam integer beats
+--- suspend execution until the given fraction of a beat is reached in time;
+-- requires being called from within a coroutine launched with clock.run.
+-- @tparam float beats
 clock.sync = function(...)
   return coroutine.yield(SYNC, ...)
 end
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -20,14 +20,17 @@ end
 local SLEEP = 0
 local SYNC = 1
 
---- suspend execution of a coroutine for s seconds;
+
+--- schedule a coroutine to sleep and then wake up in s seconds;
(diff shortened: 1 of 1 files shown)


lua/core/clock.lua
```diff
@@ -1,7 +1,13 @@
+--- clock coroutines
+-- @module clock
+
 local clock = {}
 
 clock.threads = {}
 
+--- create a coroutine from the given function and immediately run it. 
+-- @tparam function f
+-- @treturn integer : coroutine ID that can be used to stop it later
```
artfwo: the function parameter could be described better, we expect a certain kind of function here (a task within a script) that will suspend when `clock.sleep` and `clock.sync` are called inside it and wake up again after specified time.

lua/core/clock.lua
```diff
@@ -13,10 +19,15 @@ end
 local SLEEP = 0
 local SYNC = 1
 
+--- pause coroutine execution for s seconds.
+-- @tparam integer s : seconds
```
artfwo: missing requirement that it should be called from a function passed to clock.run

artfwo: seconds parameter isn't integer, is float

lua/core/clock.lua
```diff
@@ -13,10 +19,15 @@ end
...
 clock.sleep = function(...)
   return coroutine.yield(SLEEP, ...)
 end
 
+--- pause execution until the given fraction of a beat is reached in time;
+-- must be called from within a coroutine launched with clock.run.
+-- @tparam integer beats
```
artfwo: beats parameter isn't integer, is float

lua/core/clock.lua
artfwo: i suggest using word 'suspend' instead of 'pause', that matches lua terminology better, see https://www.lua.org/pil/9.1.html

okyeron:
Added @artfwo 's suggested changes

lua/core/clock.lua
```diff
@@ -13,10 +20,16 @@ end
 local SLEEP = 0
 local SYNC = 1
 
+--- suspend execution of a coroutine for s seconds;
+-- requires being called from within a coroutine launched with clock.run.
+-- @tparam float s : seconds
```
artfwo: is the colon part of ldoc syntax? also why there are semicolons in some function descriptions?

okyeron: Not sure (was copying from other norns libraries). I will do some checking/testing and adjust as needed.

okyeron: A period effective ends the description section. A semicolon let's you continue the description with a new comment line without it looking like a run-on sentence.

artfwo: cool, then can you also make sure that the style is consistent with other lua code, i.e. clock.set_source doesn't have period in this change.

okyeron: added periods

artfwo: what about colons?

okyeron: As I understand, the `param` or `tparam` tags can have a colon to separate the param description 
Screen Shot 2019-07-03 at 12 55 25 AM

okyeron: Looks like a colon could be also used instead of a semicolon to separate the main description text.

okyeron: added additional descriptive text for params. Not sure how best to describe "beats"

artfwo: it's a bit awkward name, it's kinda the next fraction of a beat at which we'd like to resume the coroutine (need better wording for the sentence). documentation should mention that it can be more than 1.

lua/core/clock.lua
```diff
@@ -13,10 +20,16 @@ end
 local SLEEP = 0
 local SYNC = 1
 
+--- suspend execution of a coroutine for s seconds;
+-- requires being called from within a coroutine launched with clock.run.
```
artfwo:
```suggestion
-- must be called from within a coroutine started with clock.run.
```

lua/core/clock.lua
```diff
@@ -13,10 +20,16 @@ end
...
+-- @tparam float s : seconds
 clock.sleep = function(...)
   return coroutine.yield(SLEEP, ...)
 end
 
+--- suspend execution until the given fraction of a beat is reached in time;
+-- requires being called from within a coroutine launched with clock.run.
```
artfwo:
```suggestion
-- must be called from within a coroutine started with clock.run.
```

lua/core/clock.lua
okyeron: is this any better?
```
--- suspend execution until the next fraction of a beat is reached in time;
-- must be called from within a coroutine started with clock.run.
-- @tparam float beats : Next fraction of a beat at which the coroutine will be resumed. May be a value larger than 1.
```

okyeron:
I reworked the text on sleep and sync once more

lua/core/clock.lua
```diff
@@ -1,7 +1,14 @@
...
+-- @module clock
+
 local clock = {}
 
 clock.threads = {}
 
+--- create a coroutine from the given function and immediately run it;
+-- the function parameter is a task that will suspend when clock.sleep and clock.sync are called inside it and will wake up again after specified time.
+-- @tparam function f
+-- @treturn integer : coroutine ID that can be used to stop or sync it later
```
artfwo:
```suggestion
-- @treturn integer : coroutine ID that can be used to stop it later
```

lua/core/clock.lua
```diff
@@ -13,10 +20,19 @@ end
 local SLEEP = 0
 local SYNC = 1
 
+
+--- schedule a coroutine to sleep and then wake up in s seconds;
```
artfwo:
```suggestion
--- yield and schedule waking up the coroutine in s seconds;
```

lua/core/clock.lua
```diff
@@ -13,10 +20,19 @@ end
...
+-- must be called from within a coroutine started with clock.run.
+-- @tparam float s : seconds
 clock.sleep = function(...)
   return coroutine.yield(SLEEP, ...)
 end
 
+ 
+--- schedule a coroutine to sync up with the next fraction of a beat (beats);
```
artfwo:
```suggestion
--- yield and schedule waking up the coroutine at beats beat;
```

lua/core/clock.lua
```diff
@@ -13,10 +20,19 @@ end
...
+-- the coroutine will pause for the time required to reach a given fraction of a beat;
```
artfwo:
```suggestion
-- the coroutine will suspend for the time required to reach the given fraction of a beat;
```

lua/core/clock.lua
```diff
@@ -13,10 +20,19 @@ end
...
+-- must be called from within a coroutine started with clock.run.
+-- @tparam float beats : Next fraction of a beat at which the coroutine will be resumed. May be a value larger than 1.
```
artfwo:
```suggestion
-- @tparam float beats : next fraction of a beat at which the coroutine will be resumed. may be larger than 1.
```

artfwo:
Thanks, I added a few suggestions. Please feel free to merge them as you see fit and merge the PR when you consider it ready.

PR #854, opened 2019-07-08
[WIP] Softcut output filters

catfact:
added output filter to each softcut voice. the DSP is basically a one-line change, adding commands, OSC and lua is a lot more. 

note that this is a breaking change to OSC command list and lua API! the input and output filters need to be differentiated; i replaced `filter_` with `pre_filter` and `post_filter`

this is completely untested! have not even cleaned up formatting! putting it up just in case anyone has time to get to it before i do (which will be soon but probably not today)


tehn:
i can help on the c/lua glue!

catfact:
the new parameters use `_norns.cut_param()`, so i don't think new c/lua glue is required, and i updated `softcut.lua` 

the main other thing that needs updating is scripts / libraries that call `softcut.filter_xyz`

tehn:
aha, i see what you mean. this seems done then? just needs doc updates?

breaking change is the addition of `pre_` to the existing filters, correct? pretty sure hardly any scripts use them currently, but good to know.

will test. excited to hear this.

side note: i should implement a `softcut_defaults` OSC message that resets crone instead of sending 400 individual OSC messages...

catfact:
>  this seems done then? just needs doc updates?

it needs _testing_

PR #855, opened 2019-07-08
[WIP] Softcut reset

catfact:
another totally untested PR for issue #812 

depends on PR #854, sorry

PR #856, opened 2019-07-09
[WIP] Softcut postfilter reset

catfact:
this is a combination of last two PRs (#854, #855) b/c i mixed them together and it didn't seem worth it to untangle them. 

tested on desktop with OSC only. fixed problem with post-filter initialization. sounds nice.

reset seems fine after some more small fixes


catfact:
still haven't tested any of the lua/c glue, there could be more typos... but pretty striaghtforward

tehn:
awesome. thursday is test day for me, will go through it on hardware.

lua/core/softcut.lua
```diff
@@ -153,39 +172,51 @@ SC.event_phase = function(f) norns.softcut_phase = f end
...
+     state[i].post_filter_hp = 0
+     state[i].post_filter_bp = 0
+     state[i].post_filter_br = 0
+     state[i].level_slew_time = 0.001
+     state[i].rate_slew_time = 0.001
+     state[i].phase_quant = 1
+     state[i].phase_offset = 0
   end
-  SC.buffer_clear()
+  return voice_state
```
tehn: typo?

`voice_state` should be `state`?

catfact: yaa

catfact:
i would like to make a couple small tweaks/additions that depend on this change but are basically separate. so i'm gonna go ahead and merge it. plz open issues if you find any additional breakage...

PR #858, opened 2019-07-11
fix softcut pre-filter mod

catfact:
un-bork pre-filter frequency modulation.


catfact:
also: filter FC wasn't clamped to nyquist and could blow up the filter with high values; fixed

Issue #829, opened 2019-05-30
battery low alert


tehn:
presently there is no alert if battery is very low.

perhaps we add a screen overlay when battery is below a threshold to alert the user.

neauoire:
I'd love to try this one, is there already a class that is allowed to draw over the currently running library?

tehn:
the screen library is lua-managed here: https://github.com/monome/norns/blob/master/lua/core/screen.lua

the c bindings are the `s_etc()` functions.

scripts call `screen.etc()` which forwards the calls.

menu.lua manages PLAY vs. MENU mode. the script's `redraw()` is saved and redirected when toggling into MENU. https://github.com/monome/norns/blob/master/lua/core/menu.lua#L170

basically for a battery alert it seems reasonable to redefine `screen.update()` when the battery level drops below a certain level... with an overlay alert. it should be restored to normal when battery level is fine.

battery level callback is here: https://github.com/monome/norns/blob/master/lua/core/norns.lua#L39

welcome to the tangled mess which is the menu system (would love to rewrite it from scratch)

neauoire:
I can't manage to preserve what is underneath, what I am trying to do is create a `:modal()` method that will allow the lua core to write messages over the user's running script without completely overriding it. As soon as I redefine update(), it clears it. Any idea?

tehn:
that's peculiar. i'll have to investigate the deeper workings.

PR #859, opened 2019-07-11
low battery alert

tehn:
fixes https://github.com/monome/norns/issues/829


tehn:
@neauoire done

neauoire:
Ah! Oh wow that looks really neat, that's exactly what I was trying to do. 
What's the `s_` prefix?

PR #860, opened 2019-07-12
make backwards-compatibility wrappers around pre_filter

catfact:
i don't necessarily know if this is a great idea, but leaving it here.


atomboyd:
Thanks again for this.

PR #866, opened 2019-08-01
report duplicates, improve missing include warning

### LINKED ISSUE #864: common problem checking and warnings
- if AUDIO ENGINE, check for duplicate class files and alert user (on-screen and in maiden) of duplicate engine
- if include not found, alert user of which one loudly. (right now it's buried in a stream of errors)

further suggestions welcomed

tehn:
instead of AUDIO ENGINE error, now we see DUPLICATE ENGINES and then the duplicates are printed out in maiden.

fixes #864

### DIFF
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -281,9 +281,8 @@ m.redraw[pHOME] = function()
     screen.move(0,15)
     screen.level(15)
     local line = string.upper(norns.state.name)
-    --if(menu.scripterror and state.script ~= '') then
     if(menu.scripterror and menu.errormsg ~= 'NO SCRIPT') then
-      line = line .. " (error: " .. menu.errormsg .. ")"
+      line = "error: " .. menu.errormsg
     end
     screen.text(line)
   else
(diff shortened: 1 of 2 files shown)


tehn:
improved warning for missing include

simonvanderveldt:
@tehn this fails on the buildroot image/when using busybox with
```
Dec 21 18:53:52 norns ws-wrapper[128]: norns.startup_status.timeout
Dec 21 18:53:52 norns ws-wrapper[128]: find: unrecognized: -printf
```

Can't we do this natively in Lua instead of shelling out to `find`?
Or alternatively only use `find` to return the list of scripts and then move the logic of determining duplicates to Lua?

tehn:
you're right--- part of the logic could certainly be moved.

`find` seems pretty efficient, but you're also right we could do some stuff with `io.open()` but i'm not sure how fast it's going to be comparatively (ie, parsing `ls`?)

simonvanderveldt:
@tehn I could have a look at using find to return all the `.sc` files and then determining if there are duplicates in Lua as a first start?
Shall I create a new issue for this?

tehn:
worth making an issue so we can keep track, yes

PR #867, opened 2019-08-13
reset quant phase stuff in softcutvoice:reset

catfact:
should probably reset phase quantization and offset settings along with other softcut voice params.


catfact:
ech, nevermind, is already done here
https://github.com/monome/norns/blob/master/crone/src/SoftCutClient.cpp#L219-L220

PR #870, opened 2019-08-21
update start/stop to use systemctl

okyeron:
Could we update the start/stop scripts to use `systemctl` like this instead of using the manual `matron.sh/crone.sh` and `kill`?

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,3 +1,4 @@
-cd "$(dirname "$0")"
-./crone.sh > /dev/null &
-./matron.sh > /dev/null &
+sudo systemctl start norns-jack.service
+sudo systemctl start norns-matron.service
+sudo systemctl start norns-crone.service
+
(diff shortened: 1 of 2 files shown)


simonvanderveldt:
AFAIK these scripts are for use during development, when running the binaries from the build dir directly. So killing/restarting the systemd services won't really work for those.
I could be wrong though.

tehn:
sorry for the delay. these will sortof work but i do not think it's totally relevant.

during dev if restarting the applications a lot it's probably easier/common not to use systemd but just run them directly so the start/stop should work fine. typically during debug i'll actually run matron or crone on an open terminal to see print feedback.

that said, i manually shut down the services with systemctl prior to debugging usually.

perhaps better than modifying the scripts is a tiny doc addition somewhere about process control/

okyeron:
Ah... OK. 

I have a couple scripts like these that I'd been using so I thought they might be useful.

Good to get some additional background on them and hear how you all use them.

Issue #874, opened 2019-09-07
softcut: set_level_cut_cut src and dst arguments reversed


ryanlaws:
The `src` and `dst` arguments appear to be reversed in softcut's `set_level_cut_cut` API method.

Docs:
https://monome.github.io/norns/doc/modules/softcut.html#SC.level_cut_cut

Based on my limited understanding, it appears that the issue is on this line:
https://github.com/monome/norns/blob/master/crone/src/SoftCutClient.cpp#L92

PR #875, opened 2019-09-07
Match SET_LEVEL_CUT_CUT arg order to docs/API (#874)

ryanlaws:
It looks like the order of the `src` and `dst` args was reversed. Hopefully this fixes it.

Docs:
https://monome.github.io/norns/doc/modules/softcut.html#SC.level_cut_cut

### DIFF
--- a/crone/src/SoftCutClient.cpp
+++ b/crone/src/SoftCutClient.cpp
@@ -89,7 +89,7 @@ void crone::SoftCutClient::handleCommand(Commands::CommandPacket *p) {
             inLevel[p->idx_0][p->idx_1].setTarget(p->value);
             break;
         case Commands::Id::SET_LEVEL_CUT_CUT:
-            fbLevel[p->idx_0][p->idx_1].setTarget(p->value);
+            fbLevel[p->idx_1][p->idx_0].setTarget(p->value);
             break;
             //-- softcut commands
         case Commands::Id::SET_CUT_RATE:
--- a/crone/src/SoftCutClient.cpp
+++ b/crone/src/SoftCutClient.cpp
@@ -45,14 +45,14 @@ void crone::SoftCutClient::clearBusses(size_t numFrames) {
 }
 
 void crone::SoftCutClient::mixInput(size_t numFrames) {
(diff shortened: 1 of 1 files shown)


catfact:
thanks for catching that, jeez 😓 

i think i would prefer to fix this by changing the application of the structure from `src, dst` to `dst, src`
here:
https://github.com/monome/norns/blob/master/crone/src/SoftCutClient.cpp#L54

but of course, either fix is OK so i approve this  PR

ryanlaws:
> thanks for catching that, jeez 

I think this just hasn't been used a whole lot so far - I don't see very many examples with it.
 
> i think i would prefer to fix this by changing the application of the structure from `src, dst` to `dst, src`
> here:
> https://github.com/monome/norns/blob/master/crone/src/SoftCutClient.cpp#L54

Hmm, in this `mixInput` method? I don't see where that gets called other than `process` which looks pretty general to me (as in, maybe it could be used for mixing engines and ADC). I'm pretty new to this codebase, though, so I'm sure I'm missing something. 

I did poke around to see if `dst, src` could be done at the lua layer, but that argument order ended up going several levels deep, not just in lua but in C++ as well, and I didn't want them to be mislabeled anywhere.

> but of course, either fix is OK so i approve this PR

Thanks!

catfact:
ds `SoftcutClient::mixInput` is where the feedback levels are applied...
yeah, IMO that's the place to make the change so that all supporting methods are in agreement

ryanlaws:
Ah, OK... I think I'm starting to get it. 

Also, if I'm reading this right, it also looks like the inter-voice feedback is getting doubled by the outer `ch` loop? Because the `input[v].mixFrom(output[w], numFrames, fbLevel[v][w])` line will get called twice for each channel despite not using or being dependent on `ch`. In that case, the `v` loop should be the outer and the `ch` and `w` loops should be siblings, right?

I apologize if I'm way off here, again I'm just getting into this codebase and I'm very much a C++ noob!

ryanlaws:
@catfact is this what you had in mind?

catfact:
yes, thanks, that naming is much clearer

PR #878, opened 2019-09-09
audio.file_info() now uses absolute path

tehn:
previously had `_path.dust` built-in which basically makes this difficult to use.


tehn:
this will break scripts (ie mlr) so i might hold off merging this until an eminent update release

Issue #779, opened 2019-03-25
implement async os commands


tehn:
currently there are quite a few features that block. (wifi, file listings, etc).

UPDATE:

something like:

```
function callback_func(returned_txt)
  print(returned_txt)
end

util.cmd("ls ~", callback_func)
```

i can try this in c with a thread though i'm wondering how to parse the resulting txt as it may be long, ie not fit in an event data type. perhaps the txt gets sent line by line?

tehn:
context: https://github.com/monome/norns/issues/779

catfact:
i'm not sure what this issue is really about.

a coro is just a thing that can yield execution to it's calling context, then be resumed an arbitrary later time with it's old contents (stack contents.) by itself, i don't think they can actually perform concurrent execution.

for example, the clocks lib uses coroutine to create a seamless asynchronous syntax in lua , but requires the backing of system threads to actually resume the clock coros that have yielded.

tehn:
you're right, this is poorly worded.

one thing that would make sense is to create an os call command which can be asynchronous... provide a command and a callback upon completion. this would pretty much solve the use cases where blocking is a real problem.

tehn:
ahead of an attempt to solve this, considering some options...

1. should the os calls just be a queue? so that there's really only one user OS thread? otherwise it seems trickier to spawn threads for each call from lua, and track separate callback functions?

this however presents the issue of a time-consuming os process stalling out later calls, but i guess that's exactly what happens now with lua... but it could be more controlled in that it doesn't stall lua itself, so the behavior could be more predictable (and not stop metros (event processing) etc).

2. for the callback function, should the data sent back be line-by-line? along with perhaps with a command-finished flag? or should it wait til finished and throw a (potentially huge) chunk? lines would accommodate the event system better.

catfact:
imho:

> should the os calls just be a queue? so that there's really only one user OS thread? 
yes. need OS worker thread with its own request queue. i'd suggest something like (pseudocode):

```
//---- [from LVM main thread]

// called from lua
function make_os_call_request: 
   allocate os_call_start_data:
       - string containing OS command, copied from lua stack
       - lua coroutine ID to resume when finished (cleaner than using callbacks directly) 
   push os_call_start_data pointer to work queue

// called from main event loop
function handle_os_call_response_event:
    copy os_call_response_data results string to lua stack
    free os_call_response_data (pointer in main-loop event struct)
    call low-level lua handler with coroutine ID

//----- [from worker thread]
[...]

PR #879, opened 2019-09-14
async queued system cmd

tehn:
tested on hardware, works

system commands now don't block lua.

try:
```
function other_function(txt)
  print("results: "..txt)
end

norns.system_cmd("sleep 5; echo '$$$$$'")
print("lua is not blocked!")
norns.system_cmd("ls -l", other_function)
```

system commands get queued and run in the order they were issued.

each command can arg a callback function. if nil it will print.

fixes https://github.com/monome/norns/issues/779 but needs to be implemented properly in the menu logic.


tehn:
certainly, if you think this one function merits its own file. i will get on it.

tehn:
also i realize `system_cmd` is possibly not the most elegant function name, if there is a better suggestion. lua has `os.execute` etc

PR #880, opened 2019-09-14
flip and  rescale pan

tehn:
fixes https://github.com/monome/norns/issues/817

tested on hardware


tehn:
planning to merge all the open PR's soon, ahead of an update. looking for feedback and complaints about anything here.

catfact:
thats fine. sorry i haven't been more help , very busy. i would move the calculation out of the sample loop and into the setter to save a couple ops per sample, but this is not a world-shaking problem.

PR #881, opened 2019-09-14
weaver cleanup

tehn:
made a macro to clean it up

### DIFF
--- a/crone/src/Bus.h
+++ b/crone/src/Bus.h
@@ -195,9 +195,9 @@ namespace  crone {
             for(size_t fr=0; fr<numFrames; ++fr) {
                 x = a.buf[0][fr];
                 l = level.update();
-                c = pan.update()/2+0.5f; // scale -1 (L) to 1 (R)
-                buf[0][fr] += x*l*(1.f-c);
-                buf[1][fr] += x*l*c;
+                c = pan.update();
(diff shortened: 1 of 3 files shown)


catfact:
arg, i have several issues with this merge.

catfact:
starting with the fact that it pulls in several unrelated changes from other open PRs that are still in review!

tehn:
i rolled back the open PR changes (accidentally did this work on top of the other branch)

this merge only added the macro cleanup. (i think, unless you see something?)

were there other issues?

catfact:
oh, you're right, sorry.

my only specific issue with the `weaver.c` changes is the macro. there is no reason this shouldn't be a static function that takes the LVM state. the macro assumes the existence of a specific named variable at local scope, which is very brittle.

catfact:
there are a number of other ways we can add helpers to make weaver.c more concise if that is the goal. 

i'll just do another round of changes...

tehn:
ok i see what you're saying. sorry for the trouble.

indeed, i was about to get around to cleaning up globals/etc and scrolling through boilerplate code sidetracked me into this thing

tehn:
untested fix: https://github.com/monome/norns/compare/weaver-macro-fix?expand=1

will be near hardware tomorrow

PR #885, opened 2019-10-01
Crone meter scaling

catfact:
added "proper" dbFS meter scaling according to IEC standards:
iec-db-meter

applied this to input/output amplitude polls in crone.

(numerical source is in `crone/dsp/workbench`, which sets up a little scientific python environment for such things.)

i have **not tested this at all.** i'm pushing / opening this before i fall asleep. but i can get to it with hardware tomorrow evening.

(note: this pulls in PR #884 as well, because i couldn't compile from `norns/` without it.)


tehn:
thank you thank you. i will test this later this afternoon.

PR #886, opened 2019-10-03
Support multiple ports per MIDI device

### LINKED ISSUE #536: unable to use all usb midi ports
under usb a particular midi device may have a number of ports.
this is used extensively for things like midi hubs (as opposed to usb hubs), but also for things like simple USB to midi din converts

however, it appears that alsa is only creating a rawmidi file for ports 0
(and norns uses rawmidi) 

example:
we@norns:~ $ amidi -l
Dir Device    Name
IO  hw:0,0    pisound MIDI PS-3TG77SW
IO  hw:2,0,0  Ableton Push 2 MIDI 1
IO  hw:2,0,1  Ableton Push 2 MIDI 2
IO  hw:3,0,0  Virus TI MIDI
IO  hw:3,0,1  Virus TI Synth
IO  hw:4,0,0  MIDIMATE II MIDI 1
IO  hw:4,0,1  MIDIMATE II MIDI 2
IO  hw:5,0,0  Pyramid MIDI USB MIDI 1

we can see here 3 devices, Push2, Virus TI and the midi mate (usb to din converter) , all have multiple midi ports.

now if we look at /dev/snd/midi*

we@norns:~ $ ls -al /dev/snd/midiC*
crw-rw---- 1 root audio 116,   8 Aug 31 19:00 /dev/snd/midiC0D0
crw-rw-rw- 1 root audio 116,  72 Aug 31 19:00 /dev/snd/midiC2D0
crw-rw---- 1 root audio 116, 104 Aug 31 19:00 /dev/snd/midiC3D0
crw-rw---- 1 root audio 116, 136 Sep  1 12:51 /dev/snd/midiC4D0
crw-rw---- 1 root audio 116, 168 Sep  1 12:51 /dev/snd/midiC5D0

we can see we only end up with one rawmidi file for each device.

an in practice what I found is that this midi can only be ever sent to the first port

so in the above this means...
- virus cannot be used as a midi controller/synth - since you need to be connected to second port
- MIDIMATE din converter - i can either send midi or receive it but not both ;) 

----------

Im not sure if alsa can be configured to create a rawmidi file for each subdevice, this would be the no code change solution. 
but I couldn't figure it out from the alsa documentation, I think its configured in /usr/share/alsa/alsa.conf, but that a pretty cryptic file... 
anyone with more experience in alsa, no what magic runes may make this work? 

on other projects Ive avoided rawmidi and instead used the alsa devices were you can specify the full device (e.g. hw:3,0,1) so never had an issue with this.

ryanlaws:
Trying to fix #536 but this has issues:
- ~~MIDI device selection list not growing with ports~~
- ~~Unplugging devices doesn't remove them correctly~~
- Includes are dirtier
- Device list code is generally dirtier
- Probably needs more error logging
- I'm a C noob so I may have added memory leaks
- I'm a C noob so the code style is probably off

### DIFF
--- a/matron/src/device/device_list.c
+++ b/matron/src/device/device_list.c
@@ -92,7 +92,7 @@ void dev_list_add(device_t type, const char *path, const char *name) {
     case DEV_TYPE_MIDI:
         midi_port_count = dev_port_count(path);
         for (unsigned int pidx = 0; pidx < midi_port_count; pidx++) {
-            d = dev_new(type, path, name, midi_port_count == 1, pidx);
+            d = dev_new(type, path, name, midi_port_count > 1, pidx);
             ev = post_add_event(d, EVENT_MIDI_ADD);
             if (ev != NULL) {
                 ev->midi_add.dev = d;
--- a/matron/src/device/device_list.c
+++ b/matron/src/device/device_list.c
@@ -24,15 +24,21 @@ struct dev_q {
 
 struct dev_q dq;
 
-static struct dev_node *dev_lookup_path(const char *path) {
-    struct dev_node *n = dq.head;
+static struct dev_node *dev_lookup_path(
+    const char *path,
+    struct dev_node *node_head
+) {
     const char *npath;
-    while (n != NULL) {
-        npath = n->d->base.path;
+    if (node_head == NULL) {
+        node_head = dq.head;
+    }
+
+    while (node_head != NULL) {
+        npath = node_head->d->base.path;
         if (strcmp(path, npath) == 0) {
-            return n;
+            return node_head;
         }
-        n = n->next;
+        node_head = node_head->next;
     }
     return NULL;
 }
@@ -123,32 +129,8 @@ void dev_list_add(device_t type, const char *path, const char *name) {
     }
 }
 
-void dev_list_remove(device_t type, const char *node) {
-    struct dev_node *dn = dev_lookup_path(node);
-    if(dn == NULL) { return; }
-    union event_data *ev;
-    switch(type) {
-    case DEV_TYPE_MONOME:
-        ev = event_data_new(EVENT_MONOME_REMOVE);
-        ev->monome_remove.id = dn->d->base.id;
-        break;
-    case DEV_TYPE_HID:
-        ev = event_data_new(EVENT_HID_REMOVE);
-        ev->hid_remove.id = dn->d->base.id;
-        break;
-    case DEV_TYPE_MIDI:
-        ev = event_data_new(EVENT_MIDI_REMOVE);
-        ev->midi_remove.id = dn->d->base.id;
-        break;
-    case DEV_TYPE_CROW:
-        ev = event_data_new(EVENT_CROW_REMOVE);
-        ev->crow_remove.id = dn->d->base.id;
-        break;
-    default:
-        fprintf(stderr, "dev_list_remove(): error posting event (unknown type)\n");
-        return;
-    }
-    event_post(ev);
+void dev_remove_node(struct dev_node *dn, union event_data *event_remove) {
+    event_post(event_remove);
 
     if(dq.head == dn) { dq.head = dn->next; }
     if(dq.tail == dn) { dq.tail = dn->prev; }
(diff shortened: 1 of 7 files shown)


catfact:
thanks @ryanlaws! super helpful.

maybe this is a good candidate for an upstream feature branch. if we can get data to/from multiple ports then that's a great start, and the other issues can be taken on separately.

i won't have time to really look at this for a few days i'm afraid, but nothing strikes me as outrageous on the surface.

ryanlaws:
Thanks @catfact ! The first and second items I think hurt the UX a bit, so I'm going to dig in a bit more. Hopefully they aren't over my head. Otherwise, which upstream branch should I target instead?

ryanlaws:
Alrighty! I think I've taken care of the 2 issues I cared about. This could use some cleanup - specifically, if code standards allow I think the MIDI `case`s could be moved to the top of either `switch` and then prior assumptions and terseness could remain intact - but your time is limited and it's probably not worth holding up review.

ryanlaws:
OK, I'm happy now.

tehn:
super excited to check this out. i'm expecting i'll have time to dive in sunday. thank you so much!

ryanlaws:
You're so welcome! Thank you and I appreciate any feedback you offer!

tehn:
do we have a list of known multi-midi devices? i may be at a loss to test if i don't have one of them, but can make sure the changes don't break normal behavior

tehn:
(i think) the name always gets a port index at the end?

it'd be nice that if there is only one port, that we don't always add "1" to the end of the name.

ryanlaws:
> do we have a list of known multi-midi devices? i may be at a loss to test if i don't have one of them, but can make sure the changes don't break normal behavior

For what it's worth, I've tested with a Korg Monologue and an Edirol UM-2, both of which have 2 ports. I think @catfact said he had something on loan that should be able to repro. The referenced issue mentioned another piece of equipment too.

> (i think) the name always gets a port index at the end?
> it'd be nice that if there is only one port, that we don't always add "1" to the end of the name.
Great idea - I was rolling this one around myself, but I couldn't think of an elegant way to do it. I think this would definitely be an improvement.

ryanlaws:
OK, let me know if that works for you! It's a little ugly but it worked OK in my testing.

catfact:
@tehn or anyone get a change to check this out?  
(i didn't. on sudden business travelling. have beatstep pro at home that should be an appropriate thing to test with.)

tehn:
works with all my single point midi things!

no multi point here but if said to work, we could merge

tehn:
@artfwo perhaps you want to take a look before i merge this? (since i recall you worked on the original midi code?)

artfwo:
LGTM. i also don't have any multi-port devices at hand right now to test this. i would perhaps change `dev_new` signature to something like:

```
dev_new(
    device_t type,
    const char *path,
    const char *name,
    bool multi_port_device,
    unsigned int port_index,
)```

i.e. swap the multi_port_device flag and index and change the former to be true, if there are multiple ports, but it's a purely cosmetic suggestion and certainly isn't a blocker for merging.

**Comment by ryanlaws** [comment]
Good idea @artfwo . Updated.

**Comment by ryanlaws** [comment]
Thank you!

**Comment by ngwese** [comment]
I have several multi-port midi devices. I'll give it a test right now.

**Comment by ngwese** [comment]
I'm happy to report this is working _great_ with an iConnectivity mio2 - I can now send MIDI back between two norns, have two norns share access to MIDI din based devices, and send MIDI between norns and ansible (in MIDI mode). ...all selectable by picking different ports on the mio2 from the norns system menus.

Super awesome.

**Comment by ryanlaws** [comment]
Thank you for testing!

File: matron/src/device/device_list.c [hunk 1]
```diff
@@ -72,35 +75,86 @@ void dev_list_add(device_t type, const char *path, const char *name) {
...
+        d = dev_new(type, path, name, true, 0);
+        ev = post_add_event(d, EVENT_CROW_ADD);
         break;
     default:
         fprintf(stderr, "dev_list_add(): error posting event (unknown type)\n");
         return;
     }
-    event_post(ev);
+    if (ev != NULL) {
+        ev->monome_add.dev = d;
```
tehn: seems this is using the `monome_add` structure for HID and CROW

matron/src/device/device_list.c
```diff
@@ -72,35 +75,86 @@ void dev_list_add(device_t type, const char *path, const char *name) {
...
+    const char *node
+) {
+    struct dev_node *next_node; 
+    event_post(event_remove);
+
+    if(dq.head == dn) { dq.head = dn->next; }
+    if(dq.tail == dn) { dq.tail = dn->prev; }
+    remque(dn);
+    dq.size--;
+
```
tehn: fairly sure the problem is something with these three lines 138-140...

matron/src/device/device_list.c
okyeron: not sure it's relevant, but want to mention this comment from when HID was getting fixed up:
https://github.com/monome/norns/pull/662#issuecomment-447229707
Although... I can't remember specifics of the debug I did at that point.

PR #887, opened 2019-10-04
param remove

pq:
this *almost* works but inexplicably falls over for some but not all param ids...

```
params:print()
paramset [arcadia]
1 enc3 = timbre
2 shape = 0.0 
3 timbre = 0.5 
4 sub = 0.0 
5 noise = 0.0 
6 stereo width = 0.5 
7 pitch lag = 0.0 
8 filter gain = 0.0 
9 cut = 8.0 
10 cut env amt = 0.0 
11 cut attack = 0.05 
12 cut sustain = 0.9 
13 cut release = 1.0 
14 level = 0.15 
15 amp attack = 0.05 
16 amp decay = 0.1 
17 amp sustain = 0.9 
18 amp release = 1.0 
19 unnamed = ---
20 output = crow ii JF
<ok>
params:remove('enc3')
lua: /home/we/norns/lua/core/paramset.lua:219: invalid paramset index: enc3
stack traceback:
	[C]: in function 'error'
	/home/we/norns/lua/core/paramset.lua:219: in function 'core/paramset.lookup_param'
	/home/we/norns/lua/core/paramset.lua:142: in function 'core/paramset.remove'
	(...tail calls...)
```

😬 

but then:

```
params:remove('level')
<ok>
```

or maybe i'm going about this in entirely the wrong way?

/cc @tehn

### DIFF
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -121,7 +121,28 @@ end
 function ParamSet:print()
   print("paramset ["..self.name.."]")
   for k,v in pairs(self.params) do
-    print(k.." "..v.name.." = "..v:string())
+    local name = v.name or 'unnamed' -- e.g., separators
+    print(k.." "..name.." = "..v:string())
+  end
+end
+
+local function get_index(t, val)
+  for index, v in ipairs (t) do
+      if (v == val) then
+        return index
+      end
+  end
+  return nil
+end
+
+-- remove.
+function ParamSet:remove(id)
+  local param = self:lookup_param(id)
+  local index = get_index(self.params, param)
+  if (index) then
+    self.count = self.count - 1
+    self.lookup[id] = nil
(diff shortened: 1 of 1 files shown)


tehn:
this _seems_ reasonable, but i have to stare at the existing params code a bit to remember how it all works ;)

tehn:
the issue is that `table.remove` decrements all the indexes above the removal, so the reverse lookup table `params.lookup` basically gets trashed.

so if you remove something, then remove something later in the list it will hit the wrong param.

so the removal needs to do a sort of rebuild of the `lookup` table

pq:
ah; that makes sense. thanks for taking a look!

pq:
after thinking about this some more, i'm not convinced `remove(id)` is really the desired API and really it would be better to just `clear` and re-add.

Issue #888, opened 2019-10-05
wrapping artifacts when clipping in TAPE (?)


catfact:
see:
https://llllllll.co/t/norns-help/14016/1939

suspect 24b encoding not handling out-of-range signals well.

PR #889, opened 2019-10-05
Fix clip before tape write

catfact:
added hardclipper in tape writer class. 

out-of-range values were wrapping when encoded to fixed point. (seems like pretty bad behavior from libsndfile, but w/e.)

could probably implement the hardclipper more efficiently, without conditionals, but this will get it done.

### DIFF
--- a/crone/src/SoftCutClient.cpp
+++ b/crone/src/SoftCutClient.cpp
@@ -71,21 +71,15 @@ void crone::SoftCutClient::handleCommand(Commands::CommandPacket *p) {
     switch (p->id) {
         //-- softcut routing
         case Commands::Id::SET_ENABLED_CUT:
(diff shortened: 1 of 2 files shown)


simonvanderveldt:
@catfact https://github.com/erikd/libsndfile/issues/294#issuecomment-308981161

catfact:
@simonvanderveldt thanks! cutting through the snark i see the useful hint to try `SFC_SET_CLIPPING`. so, will do that and report back.

catfact:
thanks using `SF_SET_CLIPPING` instead, which is surely more efficient and definitely happens on the right thread. 

NB: have only tested on x64 so far

tehn:
fixes #888 

(will test on hardware today)

PR #897, opened 2019-10-08
Softcut: use a fixed record offset of 8 samples

synthetiv:
@catfact as [threatened](https://llllllll.co/t/norns-2-0-softcut/20550/133?u=synthetivv) earlier on lines. Is this useful? Let me know!

### DIFF
--- a/crone/osc-methods.txt
+++ b/crone/osc-methods.txt
@@ -45,7 +45,6 @@ osc methods:
  /set/param/cut/pre_level [if]
  /set/param/cut/rec_flag [if]
  /set/param/cut/play_flag [if]
- /set/param/cut/rec_offset [if]
  /set/param/cut/position [if]
  /set/param/cut/pre_filter_fc [if]
  /set/param/cut/pre_filter_fc_mod [if]
--- a/crone/osc-tests/softcut-engine.scd
+++ b/crone/osc-tests/softcut-engine.scd
@@ -18,7 +18,6 @@ n.sendMsg("/command/fade_time", 0, 0.1 );
 n.sendMsg("/command/rec_level", 0, 1.0 );
 n.sendMsg("/command/pre_level", 0, 0.6 );
 n.sendMsg("/command/rec_flag", 0, 1);
-// n.sendMsg("/command/rec_offset", 0, -20);
 n.sendMsg("/command/position", 0, 1 );
 n.sendMsg("/command/filter_fc", 0, 10000 );
 n.sendMsg("/command/filter_fc_mod", 0, 1.0 );
--- a/crone/src/Commands.h
+++ b/crone/src/Commands.h
@@ -67,7 +67,6 @@ namespace crone {
             SET_CUT_FADE_TIME,
             SET_CUT_REC_LEVEL,
             SET_CUT_PRE_LEVEL,
-            SET_CUT_REC_OFFSET,
 
             SET_CUT_PRE_FILTER_FC,
             SET_CUT_PRE_FILTER_FC_MOD,
--- a/crone/src/OscInterface.cpp
+++ b/crone/src/OscInterface.cpp
@@ -385,11 +385,6 @@ void OscInterface::addServerMethods() {
         Commands::softcutCommands.post(Commands::Id::SET_CUT_PLAY_FLAG, argv[0]->i, argv[1]->f);
     });
 
-    addServerMethod("/set/param/cut/rec_offset", "if", [](lo_arg **argv, int argc) {
-        if(argc<2) { return; }
-        Commands::softcutCommands.post(Commands::Id::SET_CUT_REC_OFFSET, argv[0]->i, argv[1]->f);
-    });
-
     addServerMethod("/set/param/cut/position", "if", [](lo_arg **argv, int argc) {
         if(argc<2) { return; }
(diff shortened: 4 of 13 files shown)


catfact:
thanks! i'm honestly undecided as to whether this is necessary. 
- having an arbitrary offset does allow some different/weird applications (like a multitap delay with a single voice)
-on the other hand, it's confusing, and messing with offset time is a good way to introduce surprising behaviors (long offset, short loop) and/or glitches (offset places wr/rd heads inside interp windows.)
- back on the first hand, it's entirely ignorable and `softcut.reset()` restores defaults.

so... i dunno? open to opinions.

tehn:
can you imagine any interesting abuse strategies?

otherwise yeah it is a tripping point. 

but we could also just delist the function in the docs... sinister

catfact:
> can you imagine any interesting abuse strategies?

well, sure. in a nutshell, having arbitrary offset and using the voice mix matrix lets you make pretty complex delay / filter structures with a single voice.

here's a supercollider script sending params to local `crone` process to demonstrate...

catfact:
```
( // execute this block first
~setup = {
	Routine {
		n = NetAddr("localhost", 9999);
		n.sendMsg("/softcut/reset");

		0.1.wait;

		n.sendMsg("/set/level/cut", 0, 1.0);
		n.sendMsg("/set/level/adc_cut", 1.0);

		n.sendMsg("/set/level/in_cut", 0, 0, 1.0);
		n.sendMsg("/set/level/in_cut", 1, 0, 1.0);

		n.sendMsg("/set/pan/cut", 0, 0);
		n.sendMsg("/set/enabled/cut", 0, 1.0);
	}.play;
};


~single_4s_echo = {
	// basic 1-second loop on voice one.
	// acts as a delay with no regeneration
	n.sendMsg("/set/param/cut/rate", 0, 1.0);
	n.sendMsg("/set/param/cut/loop_start", 0, 1.0);
	n.sendMsg("/set/param/cut/loop_end", 0, 5.0);
	n.sendMsg("/set/param/cut/loop_flag", 0, 1.0);
	n.sendMsg("/set/param/cut/fade_time", 0, 0.01);
	n.sendMsg("/set/param/cut/rec_level", 0, 1.0);
	n.sendMsg("/set/param/cut/pre_level", 0, 0.0);
	n.sendMsg("/set/param/cut/rec_flag", 0, 1);
	n.sendMsg("/set/param/cut/play_flag", 0, 1);
	n.sendMsg("/set/param/cut/position", 0, 1);
};


~move_filters = {
	n.sendMsg("/set/param/cut/pre_filter_dry", 0, 0.0);
	n.sendMsg("/set/param/cut/pre_filter_fc", 0, 6000);
	n.sendMsg("/set/param/cut/pre_filter_lp", 0, 1.0);
	n.sendMsg("/set/param/cut/pre_filter_rq", 0, 1.0);

	n.sendMsg("/set/param/cut/post_filter_dry", 0, 0.0);
	n.sendMsg("/set/param/cut/post_filter_hp", 0, 0.8);
	n.sendMsg("/set/param/cut/post_filter_fc", 0, 600);
	n.sendMsg("/set/param/cut/post_filter_bp", 0, 0.2);
	n.sendMsg("/set/param/cut/post_filter_rq", 0, 1.0);
}
)

// 0.
// this includes a call to `reset`, setting offset = -8.
// this means the record head is just behind the write head.
~setup.value;


// 1.
// a "normal" echo configuration.
// perceived echo time is controlled by loop length:
// newly written material isn't head until the read head loops around.
~single_4s_echo.value;

// 2.
// make a relatively large positive offset.
// now the write head is in front of the read head.
// perceived delay time is controlled by offset amount.
n.sendMsg("/set/param/cut/rec_offset", 0, 0.5);

// 3.
// we can make this weird by using both kinds of feedback at our disposal.
// feedback at the write head will make material repeat at the loop length,
// while the delay on new material is still equal to the offset!
n.sendMsg("/set/param/cut/pre_level", 0, 0.25);

// 4.
// we can also loop the read head back to the write head.
// this causes regeneration based on the offset amount;
// the two feedback paths are effectively crosspatched
// with some interesting(?) "cut-out" artifacts at fade points.
n.sendMsg("/set/level/cut_cut", 0, 0, 0.25);


// 5.
// one thing to note is that feedback through the read head goes through the post-filter!
// so we can make complex filtered-delay structures, with a single voice
// (lowpass on pre, highpass on post)
~move_filters.value;

// 6.
// you may notice that some material that was in the buffer between steps (1) and (2) is capture forever!
// this is because the offset record head position is _not_ explicitly constrained/wrappe dto the loop range;
// it is only wrapped to the buffer range.
// this is basically a mistake and i will PR a change for it...
```

catfact:
regarding point (6) in the above...

ok, actually, it's a little more complex. write index should be wrapped to [start, end+fade] when rate > 0, or [start-fade, end] when rate < 0. since rate can change per-sample this is complex.

so, i'd say leaving the present behavior is better. though i will be taking a close look to be sure that the minimal (default) offset isn't leaving a weird tiny gap somewhere when the fade time is zero.

synthetiv:
Yeah, I take it all back and look forward to trying out weird self-routings!

catfact:
huh actually yeah definitely need to make that change (https://github.com/monome/norns/issues/898) because otherwise there is always a gap in the loop when using a big offset.

PR #899, opened 2019-10-10
move api doc location

pq:
/cc @tehn


pq:
won't know if this is right until we land, but the new target `doc/docs/norns/api` should map to `docs/norns/api` on the deployed pages and so end up on `https://monome.org/docs/norns/api/`.

🤞

Issue #559, opened 2018-09-21
clean up system globals


pq:
follow-up from meta issue #557, tracking for clean up of leaked norns system globals.

---

<details><summary>initial run (out of date)</summary>
<p>

```lua
globals = { }
for k,v in pairs(_G) do table.insert(globals, k..' : '..tostring(v)) end
for _,v in ipairs(globals) do print(v) end

_G : table: 0x538b8
_VERSION : Lua 5.3
_reset_lvm : function: 0x1be28
assert : function: 0x76c2e7bc
audio_dir : /home/we/dust/audio/
audio_input_level : function: 0x1b1b4
audio_monitor_level : function: 0x1b114
audio_monitor_mono : function: 0x1a5b4
audio_monitor_off : function: 0x1a584
audio_monitor_on : function: 0x1a594
audio_monitor_stereo : function: 0x1a5a4
audio_output_level : function: 0x1b164
audio_pitch_off : function: 0x1a564
audio_pitch_on : function: 0x1a574
bit32 : table: 0x54898
clamped_value : 0.052140721099196
cleanup : function: 0x759a8
collectgarbage : function: 0x76c2e250
controlspec : table: 0x63a80
coroutine : table: 0x54418
data_dir : /home/we/dust/data/
debug : table: 0x55b98
dofile : function: 0x76c2e734
enc : function: 0x759a8
engine : table: 0x5fdc0
error : function: 0x76c2e1d0
fileselect : table: 0xef690
gain_hp : function: 0x1b5f8
gain_in : function: 0x1b590
get_time : function: 0x1a5c4
getmetatable : function: 0x76c2e6e4
grid : table: 0xaf500
grid_all_led : function: 0x1bf94
grid_cols : function: 0x1be6c
grid_refresh : function: 0x1bf34
grid_rows : function: 0x1bed0
grid_set_led : function: 0x1c00c
hid : table: 0x664b0
home_dir : /home/we
init : function: 0x759a8
io : table: 0x55398
ipairs : function: 0x76c2e4b0
key : function: 0x759a8
line : network >
load : function: 0x76c2e53c
load_engine : function: 0x1b438
loadfile : function: 0x76c2e674
math : table: 0x56830
metro : table: 0xa3330
metro_set_time : function: 0x1b380
metro_start : function: 0x1a608
metro_stop : function: 0x1b3e8
midi : table: 0xb6188
midi_send : function: 0x1b488
mix : table: 0x6a020
next : function: 0x76c2e060
[...]

PR #903, opened 2019-10-12
cleanup leaked globals

pq:
a few locals, leaked.

💧💧💧 

see: #559 

/cc @tehn @catfact @markwheeler


catfact:
nobody has ever used voice sync :man_shrugging:

PR #906, opened 2019-10-13
fix arg mismatch

pq:
not sure if this was tripping anyone up but it sure looks wrong 🤔 

/cc @catfact @tehn


catfact:
nobody has ever used voice sync from lua :man_shrugging:

Issue #910, opened 2019-10-13
crow input modes not changing?


rwhaling:
I found that crow's input[1] would not respond to triggers in `change` mode, even though I could validate with input[1].query() that the input signal was rising above the threshold (and back below the hysteresis level); meanwhile, input[2] responds just fine to the same signals.

More details in this thread: https://llllllll.co/t/crow-help-norns/25863/23, including a few other folks with similar issues.  

I *think* what is going on is - crow is disregarding the `.mode()` message that norns sends, because crow's method is called set_mode(): 
https://github.com/monome/crow/blob/master/lua/input.lua#L39
whereas the norns crow.input.mode() method definitely sends `.mode()` in the command text:
https://github.com/monome/norns/blob/master/lua/core/crow.lua#L56

So I think, as it is now, norns cannot set input modes at all for crow, but people have been masking it by relying on input 2 being set to change mode via First.

A fix might be as easy as adjust the command to run set_mode()?  I don't know either codebase well but I can make an initial PR if that helps.

tehn:
i am not able to replicate this bug. can you post a snippet?

also crow will report lua errors back to norns in the maiden window. `mode` is definitely correct--- see here: https://github.com/monome/crow/blob/master/lua/input.lua#L75

rwhaling:
thanks for taking a look @tehn - sorry, I should have added a quick script to repro instead of a PR.  try this:

```
-- counts gates at input 1 or 2
-- press key 2 to print volts at input 1
-- press key 3 to print volts at input 2

local count_1 = 0
local count_2 = 0

function init()
  crow.input[1].mode("change",1.0,0.1,"rising")
  crow.input[1].change = change_1
  crow.input[2].mode("change",1.0,0.1,"rising")
  crow.input[2].change = change_2
  screen.level(15)
  screen.aa(0)
  screen.line_width(1)
end

function change_1(s)
  count_1 = count_1 + 1
  redraw()
end

function change_2(s)
  count_2 = count_2 + 1
  redraw()
end

function redraw()
  screen.clear()
  screen.move(10,60)
  screen.text("count_1: "..count_1.." count:" .. count_2)
  screen.stroke()
  screen.update()
end

function key(n,z)
  if n==2 and z==1 then
    print(crow.input[1].query())
  elseif n==3 and z==1 then
    print(crow.input[2].query())
  end
end 
```
On my machine, only gates at input 2 get counted.
[...]

PR #911, opened 2019-10-13
fix issue #910: set crow input modes 

rwhaling:
haven't tested this yet, but wanted to put up a branch with a candidate fix, if it helps; I don't 100% grok the command serialization though, so apologies if I'm way off base.


tehn:
this is incorrect--- norns should work as is.

the crow syntax is here: https://monome.org/docs/crow/reference/

`input[1].mode()` should work!

tehn:
just confirmed working as-is with crow studies #2

PR #916, opened 2019-10-15
crow asl execute syntax

tehn:
match syntax with native crow

`output[1].execute` is now `output[1]()`

### DIFF
--- a/lua/core/crow.lua
+++ b/lua/core/crow.lua
@@ -75,7 +75,6 @@ function output.new(x)
   o._slew = 0
   o.query = function() crow.send("get_out("..o.n..")") end
   o.receive = function(v) print("crow output receive: "..o.n.." "..v) end
-  o.execute = function() crow.send("output["..o.n.."]()") end
   setmetatable(o,output)
   return o
 end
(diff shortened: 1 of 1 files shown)


tehn:
note: norns update needs to be synchronized with crow studies update and doc update (and fix to awake)

lua/core/crow.lua
```diff
@@ -75,7 +75,6 @@ function output.new(x)
   o._slew = 0
   o.query = function() crow.send("get_out("..o.n..")") end
   o.receive = function(v) print("crow output receive: "..o.n.." "..v) end
-  o.execute = function() crow.send("output["..o.n.."]()") end
   setmetatable(o,output)
```
pq: maybe leave the `o.execute` alias in for now w/ a todo to remove it in, say `2.3.0`?  you might also consider printing a "deprecated" warning to the console if it gets used...

this might make the transition a little less painful for folks.

tehn: good idea. not sure how many people have made anything yet... but definitely makes it easier

Issue #915, opened 2019-10-15
crow reconnect fail


tehn:
seeing this sometimes:
```
>> ttyACM found, but not a crow
dev_list_add: error allocating device data
```

it then connected if i turned off the input streams via `druid`.

something must have broken with v1.0.0 crow firmware re: matron detection.

tehn:
fixed by #917

PR #917, opened 2019-10-15
fix crow detection

tehn:
crow detection previously failed if input streams were on (flooding serial input so the `^^id` is lost)

### DIFF
--- a/matron/src/device/device_crow.c
+++ b/matron/src/device/device_crow.c
@@ -18,6 +18,8 @@
 #define TEST_NULL_AND_FREE(p) if( (p) != NULL ) { free(p); } \
     else { fprintf(stderr, "error: double free in device_crow.c\n"); }
 
+#define CROW_RETRIES 10
+
 int dev_crow_init(void *self) {
     struct dev_crow *d = (struct dev_crow *)self;
     struct dev_common *base = (struct dev_common *)self;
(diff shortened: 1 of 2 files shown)


tehn:
fixes https://github.com/monome/norns/issues/915

matron/src/device/device_crow.c
```diff
@@ -47,10 +49,15 @@ int dev_crow_init(void *self) {
     read(d->fd, s, 255); // clear buffer
     write(d->fd,"^^i\n\0",5);    
     usleep(1000);
-    read(d->fd, s, 255);
-    //fprintf(stderr,"crow init> %i %s",len,s);
-
-    if(strstr(s,"^^identity")==NULL) {
+    int retry = 0;
+    while(retry < CROW_RETRIES) {
```
pq: do you want a short wait (`usleep`) in the loop?  or is the `read` blocking enough?

matron/src/device/device_crow.c
```diff
@@ -47,10 +49,15 @@ int dev_crow_init(void *self) {
     read(d->fd, s, 255); // clear buffer
     write(d->fd,"^^i\n\0",5);    
     usleep(1000);
```
pq: maybe the `usleep` could move into the retry loop?

tehn: so the usleep is to give crow time to respond.

the loop is to simply clear the entire serial buffer, so no usleep is needed.

Issue #890, opened 2019-10-05
crow output action assignments use quotes


tehn:
looking for a way to make the action syntax the same for native crow (druid) and norns crow

natively on crow (via druid/etc) you can set action commands like this: `output[1].action = lfo(1,4)`

right now the norns script
```
output[1].action = "lfo(1,4)"
```
sends `output[1].action = lfo(1,4)` via serial to crow. AFAIK there is no way to just _capture_ the right side of the equals sign. here's where it happens: https://github.com/monome/norns/blob/master/lua/core/crow.lua#L90

i do not think this is not worth solving if it requires some sort of C level hack

catfact:
yea, i don't think there's a clean way to do this exactly. (maybe with the `debug` library but not sure.)

following up from slack, at present the main idea is to define `lfo()` and the other "action functions" globally so that they return strings. to _also_ keep these out of the global environment would require executing scripts in a sandboxed environment

rough mockup:

`crow.lua`
```
crow = {}

crow.send = function(what) print(what) end

crow.add_sandbox_globals = function(env)
   env.lfo = function(i,v) return 'lfo('..i..','..v..')' end
   env.volts = function(i,v) return 'volts('..i..','..v..')' end
   return env
end

crow.output = {}

crow.output.__newindex = function(self, key, value)
   print('newindex('..key..', '..value..')')
   
   if key == 'action' then
      -- "value" should be a string returned from `crow.env` function
      -- but good idea to check type here..
      if type(value) == 'string' then
	 crow.send("output = "..value)
      else
	 print('error...')
      end
   end
end

setmetatable(crow.output, crow.output)

crow.run_script = function(path)
   print('crow.run_script('..path..')')
   -- make a new environment that inherits from the global env
   local env = {}
   setmetatable(env, {__index = _G})
   
   -- add our "fake globals"
   env = crow.add_sandbox_globals(env)

   -- attempt to compile with new environment
   local fn = loadfile(path, 'bt', env)
   
   if fn == nil then print('error...')
   else fn() end
end

return crow
```

`crow_script_launcher.lua`
```
local crow = dofile('crow.lua')
print('script launcher')
crow.run_script('crow_script.lua')

-- ASL functions don't appear in global env:
print("global 'lfo:'", lfo) -- result: nil
```

`crow_script.lua`
```
local crow = dofile('crow.lua')
print('running script')
crow.output.action = lfo(1, 12)
```

catfact:
[...]

Issue #557, opened 2018-09-21
globals


tehn:
see #425 

- [x] reset globals created by scripts between loads - fixed by https://github.com/monome/norns/pull/436
- [x] clean up system globals #559
- [x] consider adding underscores to system globals
- [x] consider putting c functions into a table - see https://github.com/monome/norns/issues/425#issuecomment-397030353
- [ ] disable overwriting system globals - see https://github.com/monome/norns/pull/436#issuecomment-403100830 @pq 
- [ ] protect some specific tables (ie `engine`) #912 @pq

pq:
> @tehn: do we have a sense for what globals we want to protect and what we want to preserve across scripts?

this is related to the general cleanup task i guess.  there may not actually be much left to protect / restore when norns is less leaky...

PR #918, opened 2019-10-15
asl builder proof-of-concept

pq:
**food for thought.**

_caveat: i'm flummoxed as to why explicit unloading of `asl` is needed (where it isn't for other libraries brought in by a `require`, like, for example `beat_clock`)._

anyway, the idea is a norns script with a line like this:

`require 'asl'`

will then be able to do things like:

```lua
crow.output[1].action = loop{ to(0,0), to(5,0.1), to(1,2) }
crow.output[1].action = lfo(1, 5) 
```

and have it behave like in `druid`.
...

the dynamic case is tricky of course and requires a low-level `send`.  so, something like:

```lua
    crow.send[[
      output[1].action = ar(1, math.random(5))
    ]]
```

the bet is that a string-builder like this will be useful in the common-case and at least not confusing in the complex one...

it has the benefits of feedback from linters and IDEs (since it's valid lua) and scores bonus points for highlighting nicely in maiden

image

and generally in IDES (like vs code):

image

as for the global pollution issue (see #557), asl functions only get added to `_G` if `asl` is included and only hang around until the next script is loaded.

i think that's probably the right thing.

i should emphasize that i think of this purely as food for thought and i'm not hugely invested in it landing as is (or at all).  i just wanted to see the idea through...

fixes: #890

/cc @tehn @catfact @trentgill


tehn:
this is great! and perplexing about the `asl` require bug. i'll take a minute to look at it.

tehn:
i'm thinking i'll merge this in parallel with a doc update. i like the approach, thanks for taking the time

tehn:
ack, sorry for lagging on this. need to find to test. will get it in the next-next update (should be forthcoming)

pq:
no worries.  more food for thought than anything!

PR #919, opened 2019-10-16
Add some missing & new glyphs to the default font

Dewb:
This PR adds eight useful missing glyphs to the extended ASCII / Unicode Latin Extended range in the default norns font.

image

| char | lua code | description |
|-----|----------|-------------|
| « | `"\u{00AB}"` | LEFT-POINTING DOUBLE ANGLE QUOTATION MARK
| ° | `"\u{00B0}"` | DEGREE SIGN
| ± | `"\u{00B1}"` | PLUS-MINUS SIGN
| µ | `"\u{00B5}"` | MICRO SIGN
| º | `"\u{00BA}"` | MASCULINE ORDINAL INDICATOR
| » | `"\u{00BB}"` | RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK
| × | `"\u{00D7}"` | MULTIPLICATION SIGN
| ÷ | `"\u{00F7}"` | DIVISION SIGN

It also adds several potentially useful glyphs from the Unicode Arrows, Mathematical Operators, Miscellaneous Technical, Geometric Shapes, and Miscellaneous Symbols charts.
 
image

| char | lua code | description |
|-----|----------|-------------|
| ← | `"\u{2190}"` | LEFTWARDS ARROW
| ↑ | `"\u{2191}"` | UPWARDS ARROW
| → | `"\u{2192}"` | RIGHTWARDS ARROW
| ↓ | `"\u{2193}"` | DOWNWARDS ARROW
| ∓ | `"\u{2213}"` | MINUS-OR-PLUS SIGN
| ∕ | `"\u{2215}"` | DIVISION SLASH
| ∞ | `"\u{221E}"` | INFINITY
| ▲ | `"\u{25B2}"` | BLACK UP-POINTING TRIANGLE
| ▶ | `"\u{25B6}"` | BLACK RIGHT-POINTING TRIANGLE
| ► | `"\u{25BA}"` | BLACK RIGHT-POINTING POINTER
| ▼ | `"\u{25BC}"` | BLACK DOWN-POINTING TRIANGLE
| ◀ | `"\u{25C0}"` | BLACK LEFT-POINTING TRIANGLE
| ☐ | `"\u{2610}"` | BALLOT BOX
| ☑ | `"\u{2611}"` | BALLOT BOX WITH CHECK
| ☒ | `"\u{2612}"` | BALLOT BOX WITH X
| ☰ | `"\u{2630}"` | TRIGRAM FOR HEAVEN
| ☱ | `"\u{2631}"` | TRIGRAM FOR LAKE
| ☲ | `"\u{2632}"` | TRIGRAM FOR FIRE
| ☳ | `"\u{2633}"` | TRIGRAM FOR THUNDER
| ☴ | `"\u{2634}"` | TRIGRAM FOR WIND
| ☵ | `"\u{2635}"` | TRIGRAM FOR WATER
| ☶ | `"\u{2637}"` | TRIGRAM FOR MOUNTAIN
| ☷ | `"\u{2638}"` | TRIGRAM FOR EARTH
| ♩ | `"\u{2669}"` | QUARTER NOTE
| ♪ | `"\u{266A}"` | EIGHTH NOTE
| ♫ | `"\u{266B}"` | BEAMED EIGHTH NOTES
| ♬ | `"\u{266C}"` | BEAMED SIXTEENTH NOTES
| ♭ | `"\u{266D}"` | MUSIC FLAT SIGN
| ♮ | `"\u{266E}"` | MUSIC NATURAL SIGN
| ♯ | `"\u{266F}"` | MUSIC SHARP SIGN

I was compelled to add these because I was particularly hurting for the degree symbol, infinity, and a forward slash/division sign that was narrower than the five-pixel glyph for standard `/`. The rest, well, once I set myself up with the font editor, I guess I got a little carried away.

Reasons this might not be a good idea:
- Unsure of the risks/dynamics surrounding modifying cairo font resources.
- I did these really quickly and this might not be the best way to thoughtfully expand the symbol options. Having said that, I do have some experience with pixel fonts, and I did my best to keep to the implied rules in the existing character set. All are 3/4/5 pixels wide with appropriate glyph widths.
- Lua 5.3 has full unicode support, and these glyphs work great on the hardware, but there might be issues lurking in the ecosystem that encouraging higher code point usage might uncover. I believe using `print()` to send multibyte characters across the websocket can currently crash the maiden console.


catfact:
thanks @Dewb that is awesome.

- seems like the .ttf files should be self-contained and ok to edit.
- i've no doubt that there will be problems here and there supporting full utf8 in the lua eval pipeline.
- but i would really like to fix those problems if at all possible.

so, maybe would be reasonable to merge this in an upstream feature branch and start identifying those bad spots?

tehn:
@Dewb rad PR! (what font editor do you use?)

i see no problem merging this as any encoding issues will be present w/ every other font included--- not a specific issue with this updated font.

Dewb:
I used FontForge on OSX. It was the first thing I tried; there may be a better option, it’s a little crashy and doesn’t have many affordances for pixel fonts.

markwheeler:
Great to have these added!

I'm seeing that a few of the characters seem to sit off the pixel grid, creating some blurry edges. The 'music flat sign' is probably the most noticeable but it also applies to the arrows and a few others. Maybe they just need a little nudging onto grid in the font?

Dewb:
Whoa, interesting; I didn't observe that in testing, but maybe I messed something up in the final version of the font, I'll take a closer look.

Dewb:
I wrote a better test program and I'm not seeing any anti-aliasing on any of the characters using `screen.text()`. How are you drawing them?

Dewb:
The new characters do look like garbage at any size that isn't a multiple of 8, but that's true of most of the characters in the default font. They could probably distort more consistently, though. And it looks like I got the margin wrong on the double-note glyphs, I'll have to fix that.

markwheeler:
Here's some sample code:

```
function redraw()
  screen.clear()
  screen.aa(1)
  
  screen.level(15)
  screen.move(10, 15)
  screen.text("ABCDEFGHI abcdefghi !?@#")
  screen.move(10, 26)
  screen.text("♭ ♮ ← ↑ → ↓")
  
  screen.update()
end
```

For me the first string looks sharp but all the symbols in the second string have some 'blurring' on them, most noticeable on the first character. Perhaps you didn't have anti-aliasing enabled in your test? Thanks for looking into this!

Dewb:
Aha! My test script had both an `screen.aa(1)` *and* a `screen.aa(0)` in it, so antialiasing wasn't actually enabled. I see the issue now, will update the characters.

PR #921, opened 2019-10-17
Fix system cmd memleak

catfact:
due to a subtlety of the pthreads API, the worker threads created in `system_cmd.c` were leaking _all the memory_ of the child processes they spawned. 

(the child process in latest version being a big ol' `find` that ate up a rather shocking 10MB a pop with a goodly number of scripts installed. [um, actually that's not quite right. it's mapped but unwired, or something.])

detaching the thread fixes this, it seems.

also uncrustified that file, because _someone_ has their vim set on 2space again. :)

### DIFF
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -362,7 +362,8 @@ end
 m.init[pSELECT] = function()
   m.sel.len = "scan" 
   m.sel.list = {}
-  norns.system_cmd('find ~/dust/code/ -name "*.lua" | sort', sort_select_tree)
+  -- weird command, but it is fast, recursive, skips hidden dirs, and sorts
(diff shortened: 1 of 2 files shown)


catfact:
last commit is a bit silly, but it reduces mem usage of the subcommand by about 25% compared to `find`, and is probably faster

tehn:
@catfact i suspect this is the problem:

https://github.com/monome/norns/blob/71f2c1e620550628fc26f8c22deec1f5d1174cc6/matron/src/system_cmd.c#L17

for super super long folder listings perhaps it's just running out of memory and plowing over

catfact:
oh, ha. indeed. 

i'll do another round of cleanup on this. `strcat` is basically pure evil...

tehn:
oy oy oy i apologize

PR #922, opened 2019-10-17
musicutil uses proper sharp symbol

markwheeler:
Tiny change that I think makes sense but also realized doesn't look any different on device 😁

### DIFF
--- a/lua/lib/musicutil.lua
+++ b/lua/lib/musicutil.lua
@@ -2,12 +2,12 @@
 -- Utility methods for working with notes, scales and chords.
 --
 -- @module MusicUtil
--- @release v1.1.0
+-- @release v1.1.1
 -- @author Mark Eats
 
 local MusicUtil = {}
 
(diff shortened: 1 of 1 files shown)


Dewb:
Ha, yeah, I originally drew a fancier sharp symbol but it didn't look that great. I think the # is the best that can be done in 5 by whatever pixels.

I'm not sure using the unicode char in the tables is a good idea because it breaks anything that uses string comparison. I'm using the degree symbol in the stuff I'm writing for Roman numeral notation, but everything that acts on those strings also accepts `*` as equivalent to `º`.

markwheeler:
Totally makes sense. Right now there's nothing in musicutil that looks up via note name (it's always done using MIDI note number) so I think it should be OK. Your 'either or' approach would be good if we do end up adding something that lets you specify by note name string in future though for sure.

tehn:
seems nice and will watch out for breakage!

PR #923, opened 2019-10-18
System cmd more cleanup

catfact:
2nd look at `system_cmd.c`

i'm now reasonably sure this will not bork even when command output massively exceeds buffer capacity. (it will truncate to the last complete line that fits in the buffer.)

### DIFF
--- a/matron/src/system_cmd.c
+++ b/matron/src/system_cmd.c
@@ -14,8 +14,15 @@
 
 #include "events.h"
 
+
+#if 1 // nice big buffers (allocated in worker thread)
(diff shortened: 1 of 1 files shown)


catfact:
(uhh.. sorry not sure what happened with that history... 1 moment will force push)

catfact:
ok that's better

tehn:
indeed, that feels a million times better.

thanks for the malloc lesson, and i'm sorry to take your time cleaning up my mess

tehn:
bonus: this code is _way_ faster. not even a blink for me of the "scanning..." wait screen on SELECT

!!!

PR #924, opened 2019-10-18
crow connect status function

### LINKED ISSUE #892: add support for querying if crow is connected
useful, for example, on script-start if you want auto-connect behavior.

for now, i've got a helper:

```lua
function helper.is_crow_connected()
  return norns.crow.dev ~= nil
end
```

but it's a hack and something on `crow` proper would be better...

tehn:
`crow.connected()` returns true/false

fixes #892


pq:
thanks!

PR #926, opened 2019-10-19
copy 1 less byte with strncpy and let it null-terminate. fixes #925,

### LINKED ISSUE #925: system_cmd.c waf compile error on pi4
https://github.com/monome/norns/blob/e22003b050d27149e4150bc4c53d7b2a0afb7b47/matron/src/system_cmd.c#L91

```
../matron/src/system_cmd.c: In function ‘run_cmd’:
../matron/src/system_cmd.c:91:5: error: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
     strncpy(cap, capture, len);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~
../matron/src/system_cmd.c:89:18: note: length computed here
     size_t len = strlen(capture);
                  ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

Waf: Leaving directory `/home/we/norns/build'
Build failed
 -> task in 'matron' failed with exit status 1 (run with -v to display more information)
```

catfact:
okey dokey,

`system_cmd.c` continues to expand its collection of extravagant comments.

probably fixes #925 which i can't reproduce; someone with rpi4 needs to test. OG norns seems fine.

### DIFF
--- a/matron/src/system_cmd.c
+++ b/matron/src/system_cmd.c
@@ -85,11 +85,13 @@ void *run_cmd(void *cmd) {
     fprintf(stderr, "finished line loop; full buffer: \n");    
     fprintf(stderr, "%s\n", capture);
 #endif
-    
-    size_t len = strlen(capture);
+
+    // strncpy adds null terminator, *unless* src length >= nbytes.
+    // this is atually fine, but new gcc throws warning
+    // so, copy 1 less byte and let strncpy add the null instead of doing it ourselves/
+    size_t len = strlen(capture) - 1;
     char *cap = malloc((len + 1) * sizeof(char));
     strncpy(cap, capture, len);
-    cap[len] = '\0';
     union event_data *ev = event_data_new(EVENT_SYSTEM_CMD);
 
     // this will get freed when the event is handled
--- a/matron/src/system_cmd.c
+++ b/matron/src/system_cmd.c
@@ -86,12 +86,10 @@ void *run_cmd(void *cmd) {
     fprintf(stderr, "%s\n", capture);
 #endif
 
-    // strncpy adds null terminator, *unless* src length >= nbytes.
-    // this is atually fine, but new gcc throws warning
-    // so, copy 1 less byte and let strncpy add the null instead of doing it ourselves/
-    size_t len = strlen(capture) - 1;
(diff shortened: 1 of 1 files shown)


okyeron:
Just tried on the pi4
```
../matron/src/system_cmd.c: In function ‘run_cmd’:
../matron/src/system_cmd.c:91:5: error: ‘strncpy’ specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
     strncpy(cap, capture, len);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~
../matron/src/system_cmd.c:89:18: note: length computed here
     size_t len = strlen(capture) - 1;
                  ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

Waf: Leaving directory `/home/we/norns/build'
Build failed
 -> task in 'matron' failed with exit status 1 (run with -v to display more information)

```

catfact:
pff, okay, have a `memcpy`

okyeron:
FWIW - I have no idea what I'm doing here and only suggested memcpy because it was mentioned in 5-6 different github/stackexchange threads.

But with that change, no error on building matron.

But now I'm getting an error on building crone - let me know if I should post a separate issue for this
```
[47/47] Linking build/crone/crone
/usr/bin/ld: crone/src/OscInterface.cpp.1.o: in function `std::_Function_handler<void (char const*), crone::OscInterface::init(crone::MixerClient*, crone::SoftCutClient*)::{lambda(char const*)#1}>::_M_invoke(std::_Any_data const&, char const*&&)':
OscInterface.cpp:(.text+0x18b4): undefined reference to `crone::AudioMeter::ampPosTableSize'
/usr/bin/ld: OscInterface.cpp:(.text+0x18b8): undefined reference to `crone::AudioMeter::ampPosTable'
/usr/bin/ld: OscInterface.cpp:(.text+0x18cc): undefined reference to `crone::AudioMeter::ampPosTable'
```

catfact:
memcpy is fine and makes sense. of course the other 2 versions were also fine and made sense.

(big caveat that there are probably lots of things that will break with multibyte characters, including maybe this. but strlen is, i think, explicitly length in bytes so maybe it's ok)

anyway cool!

PR #943, opened 2019-11-15
move c extern functions into _norns table

tehn:
these should only be called by lua libraries, so they are in the obfuscated `_norns` table now.

all lua updated though some thorough testing prior to release is needed.


tehn:
towards fixing https://github.com/monome/norns/issues/559

tehn:
current strategy (up for discussion):

- all c externs are in `_norns`
- extern callbacks are in `norns` which get redefined by modules.
- `norns` also contains script/state management tables and generally gets touched by modules that need shared memory.

user scripts should never need to touch `_norns` or `norns`. i did a pass where i renamed `norns` to `system` but it felt weird. not sure if the extern callbacks should just get their own table, and the management stuff should go elsewhere (ie `script`, `state`)

will keep thinking on it.

tehn:
- moved matron callbacks into `_norns`
- `_menu` is global.
- `norns` is still a general catch-place for shared stuff:

```lua
menu	table: 0x34d398
powerpresent	1
cpu	1
battery_current	152
system_cmd	function: 0x34a028
version	table: 0x349000
script	table: 0x342550
blank	function: 0x3494e0
try	function: 0x343668
scripterror	function: 0x373790
temp	47
disk	533
battery_percent	81
encoders	table: 0x347c00
none	function: 0x343620
state	table: 0x33bca0
crow	table: 0x351b90
init_done	function: 0x39abe8
```

i'm trying to address system design stuff at the same time but i'm getting a bit buried in this and need to move on. i think this gets our globals situation in a much better place, and this seems like a point to move forward re: params menu and other features i'd like to move onto.

will do some further testing with devices and real scripts, then i'll merge this. i do not believe there are breaking changes.

PR #950, opened 2019-11-25
error msg for missing engine

### LINKED ISSUE #949: better missing engine error message
presently if an engine is missing we just get a 0 params found, which is not actually an error-- so scripts continue to  load.

tehn:
fixes #949 

tested, works.


tehn:
that's what i had originally, but it cut off the name so i thought the name was a better clue as to how to fix.

Issue #853, opened 2019-07-07
Expand MIDI-learn to all MIDI channels


okyeron:
Looking at the `menu.lua` [code for midi-learn](https://github.com/monome/norns/blob/9f9aff28d047b8754ba47cdaeb19a938ff6e5e7c/lua/core/menu.lua#L679) it looks like it only works with CC messages on channel 1.

We could expand this to look at all the CC channels by
```
if data[1] > 176 and data[1] < 192 then -- all cc channels
```
but then we would need to store the channel number in `m.params.map` somewhere. 

How best to do that?

tehn:
midi learn overhaul is forthcoming in the new parameter system/menu/interface.

i'll be sure that channel is mappable.

okyeron:
Let me know if I can help :)

okyeron:
Additional feature request:

Make Levels controls mappable to midi CCs (system volume, input, mon, engine, tape, cut)

rbxbx:
as @okyeron suggests, I basically always want a (tactile) gain control for the norns. Relying on being attached to a mixer (that you have control over) is kind of a bummer, especially for lightweight sessions.

okyeron:
Closing as this appears to be done with #960

PR #960, opened 2019-12-02
new param menu

### LINKED ISSUE #491: parameter sets: GROUPS/HEADING
currently we can make SEPARATOR fields for grouping parameters within sets.

it'd be nice to have a HEADING to describe the group.

perhaps some nice UI where the HEADING stays sticky at the top of the screen when scrolling through?

and also not being able to land on a HEADING or SEPARATOR as an active field?

### LINKED ISSUE #522: parameter mapping: in/out range scaling
for midi cc mapping of parameter set:

- set input high/low (clamp beyond)
- set output range
- both can be 0-1.0 
- either side can be inverted
- saved with pset

### LINKED ISSUE #551: parameter mapping: CC ch select
It would be nice to have a channel selection UI in cc mapping mode, particularly an omni choice. I've been using dj mixer style midi controller but only half of the buttons/knobs are assignable because they share cc numbers of differing channels

### LINKED ISSUE #560: parameter mapping: support endless encoders
need to add an accumulator mode for those which send deltas (typically centered at 64 = 0)

### LINKED ISSUE #600: parameter map save by id, not index
currently saved by index, should save by id for compatibility

thanks @artfwo for spotting this

### LINKED ISSUE #735: param type: string
use `textentry` lib for interface.

use case: OSC IP destination. or naming a preset (?)

tehn:
do not merge --- PR for testing/tracking

- [x] new param type: GROUP. fixes #491 
- [x] SEPARATOR now can get a name, which makes it a heading. fixes #491 
- [x] KEY3 on SEP jumps to next SEP (wraps to top)
- [x] add SYSTEM>AUDIO params automatically to param menu (so they are mappable/etc)
- [x] new preset read/write interface
- [x] K1 hold quick access to PRESET and MAP
- [x] K1 hold + ENC2 jump SEP
- [x] error handling for attempted group overlap/nesting

sys/audio menu:

- [x] change display/id names, capitalize groups
- [x] remove SYSTEM > AUDIO
- [x] set up mix/levels page

midi mapping:

- [x] new interface
- [x] show OSC paths (id)
- [x] support endless encoders. fixes #560 
- [x] range scaling. fixes #522 
- [x] cc ch select. fixes #551 and #853 
- [x] mapping saving. fixes #600 

misc

- [x] create tTEXT param type fixes #735
- [x] "hidden" attribute for params = do not show in menu
- [x] pset "name" attribute to save/load

### interface

- basic tree for `EDIT`, `MAP`, `PSET`
- EDIT: E3=change val
- MAP: (midi mapping, show osc names): K3 = enter map edit
- PSET: E2 scrolls action, E3 slot, K2 back, K3 confirm action. star shows last active pset

quick jump: K1(alt)+K2 = PSET, alt+K3 = MAP toggle

- PMAP is saved when you save a PSET. there is one PMAP per script. PMAP is loaded on script load if found.

### groups

```lua
params:add_group("this",2)
params:add_number("num1","num2")
params:add_number("num2","num2")
params:add_number("not","not")
```

this makes a group with two params. `num1` and `num1` go in the group (the first two that followed the definition).

groups cannot be nested.

### named sep

```lua
params:add_separator("filter")
```

this prints "filter" above the line for the section (otherwise blank)

* K2 on sep jumps to next sep
* K1+E2 jumps between seps (plus top line)

### hide/show

any param can be hidden from the list. indexing by number only.

```
params:hide(4)
x = params:visible(4) -- returns false
params:show(4)
```

- by default everything is visible.
- you can hide entire subgroups.
- generally suggested that hiding happens in `init` as the menu does not manually rebuild itself unless it's entering/exiting a submenu. ie, more glue would have to be added to signal a menu list rebuild... which seems overkill.


tehn:
midi mapping now works. i apologize that this is taking me forever.

ngwese:
possible bug, for simple scripts which don't draw to the display the main menu gets redrawn on incoming midi data when parameters have been bound to midi.

setup
-----
- expressive-e touche connected as MIDI device 2
- script:
```
function redraw()
  screen.clear()
  screen.update()
end
```
- parameters mapped (and saved):
```
"reverb":"{accum=false, out_hi=2, out_lo=1, ch=1, dev=2, value=0, in_hi=127, in_lo=0, cc=17}"
"rev_eng_input":"{accum=false, out_hi=1, out_lo=0, ch=1, dev=2, value=0, in_hi=127, in_lo=0, cc=19}"
```

procedure
----------
- run script (see blank screen)
- K1 to return to MENU
- navigate to PARAMETERS
- K2 then MAP > REVERB > reverb
- K3 then LEARN
- K2 then rev_eng_input
- K3 then LEARN
- K2 up to PARAMETERS
- K1 to return to script view
- _touch midi controller_ and the last MENU screen is drawn but the key and enc handlers from the script are still in effect (it seems)
- hit K1 three more times to go into and out of script/menu mode eventually installs the menu key and enc handlers

tehn:
@ngwese i believe this fixes it

ngwese:
@tehn - confirmed, that does indeed fix the redraw bug

antonhornquist:
works great, haven't found any bugs yet!

FWIW some reflections on the UI after testing for a while:

i'm not too fond of params automapping to CC100 when using K3 in the PARAMETER MAP section. i'd prefer setting a mapping to be an explicit action (sometimes you're just navigating around the menu to browse settings with no intent to do an action)

i found it a bit weird that the menu by default (on booting the device and turning E1 right) is navigated into the `EDIT >` section (presenting the actual parameters and values: levels, reverb, compressor + engine ones) and not the top node (the one with `EDIT >`, `PSET >` and `MAP >`). i would prefer to start with the top menu node as "home". the way to get to `PSET >` and `MAP >` (by pressing K2 once) and the fact that they even exist is not evident without reading docs. a regular tree starting from the top is in my opinion simpler and more discroverable (= less support questions).

the shortcuts K1-K2 and K1-K3 work differently depending on where *in* the menu the user is at the moment. so at the menu top section (presenting `EDIT >`, `PSET >`, ´MAP >`) K1-K2 does not work while K1-K3 works like K3, but when navigated into either `EDIT >`, `PSET >` or `MAP >` they work as described in the PR. these also toggle between the ´EDIT >` and `PSET >` and `MAP >` sections (for K1-K2 and K1-K3 respectively), rather than navigate to one definite section (as in "K1-K2 always takes you to `PSET >`"). in my opinion adding modes on top of modes like that makes the UI a bit fiddly.

tehn:
thanks for testing!

1. i think i can easily make the param data save on exit of the screen, that makes sense. of course not getting cc 100 means tuning the value or using LEARN, but i get what you mean (i think?)
2. this makes sense. on bootup the first nav to the "param home" is fine
3. i agree. the problem is 3 modes, 2 keys, and a weird classification (in my brain) of the screens. EDIT and MAP are essentially the same, but with different data, so i like the idea of toggling between EDIT and MAP with the same key. but the PSET shorcut breaks down when _exiting_ that screen. it's almost like K3 should cease working (because it's a "toggle") and K2 just takes you back to whatever-was-last-used.

tehn:
ok:

- set default param menu to home on bootup
- nav keys: alt-K2 now toggles between PSET and last-used MAP/EDIT. alt-K3 toggles MAP/EDIT. this makes more sense to me, treating the pages like toggles.

tehn:
fixes #1030

antonhornquist:
tried this again, works well!

Issue #964, opened 2019-12-13
softcut buffer load incomplete


tehn:
seems something is not quite right with the frame counting.

```
softcut.buffer_read_mono (file, start_src, start_dst, dur, ch_src, ch_dst)
```

if `start_dst` is say `1.0` and `dur` is `-1` (full file) then the file load stops `1.0` early.

think it's something around here: 

https://github.com/monome/norns/blob/master/crone/src/SoftCutClient.cpp#L263

@catfact i know you're refactoring a bunch of this so i don't want to start trying to fix it if this code is different--- i'll check your new PR

(working on softcut studies while finding this)

catfact:
yes, do check against new PR as that is the very code that is changed.

but, i don't think i changed that particular arithmetic. 

so, the file load stops 1 second before the end of the file, is what you're saying?

catfact:
ah, you're right. that line is wrong. should probably be:
```
auto maxDurDst = BufFrames - frDst;
```

PR #958, opened 2019-12-01
Softcut buf worker

catfact:
this is WIP, but putting it up for early testing in case that's useful.

buffer management is now done with a dedicated worker thread; IPC interface makes requests that are placed on a job queue.

i've only implemented the `softcut/buffer/read_mono` command, will finish the rest and refactor/rewrite for disk efficiency.

i'd like as much testing coverage as possible for this since there are lots of edge cases:
- requesting source channel not present in file
- requesting position/durations  that are OOB for source/dest 
etc

PR #966, opened 2019-12-18
crone fix buffer load length

tehn:
fixees https://github.com/monome/norns/issues/964

spotted by @catfact 

tested/works


catfact:
i know you already merged this, but it does look me like the best way forward would be to make any needed fixes to PR #958 and merge that

catfact:
... see, now we have a quite unnecessary merge conflict!

tehn:
ah, apologies! i mistakenly thought #958 wasn't ready. i'll fix the conflict and approve it

Issue #841, opened 2019-06-14
fix ldoc comments for classes


catfact:
we're using `@module` when we apparently should be using `@classmod` for classes.

this will correctly differentiate in the output between methods (with colon and `self`) and static functions (with dot.)

also, a lot of class/module files are totally missing ldoc comments:
```
> ldoc .
/home/emb/norns/lua/core/clock.lua:69: no module() call found; no initial doc comment
/home/emb/norns/lua/core/hid_events.lua:752: no module() call found; no initial doc comment
/home/emb/norns/lua/core/mix.lua:167: no module() call found; no initial doc comment
/home/emb/norns/lua/core/vport.lua:28: no module() call found; no initial doc comment
/home/emb/norns/lua/core/wifi.lua:341: no module() call found; no initial doc comment
/home/emb/norns/lua/lib/beatclock.lua:160: no module() call found; no initial doc comment
/home/emb/norns/lua/lib/er.lua:25: no module() call found; no initial doc comment
/home/emb/norns/lua/lib/formatters.lua:172: no module() call found; no initial doc comment
/home/emb/norns/lua/lib/gridbuf.lua:99: no module() call found; no initial doc comment
/home/emb/norns/lua/lib/pattern_time.lua:107: no module() call found; no initial doc comment
```

fixing this should be easy, if slightly tedious

catfact:
actually i still don't really get it.

if i just change `@module` to `@classmod` at the top of `midi.lua`, ldoc interprets _all_ functions in the `Midi` table as "methods," (displays them with a colon.)

if i leave `@module` at the top, and add `@classmod` later (above the method declarations) then they all show up as class functions (with a dot.)

it seems to be all or nothing. but there's gotta be a way.

ypxk:
the ldoc thing has bothered me before. i don’t know that i will get a big enough chunk of spare time to figure out all the steps in submitting a pull request for fixing it, but i am going to try to do it by next week and will report back if i fail.

okyeron:
The [Sections part](https://stevedonovan.github.io/ldoc/manual/doc.md.html#Sections) of the ldoc docs mentions the `@type` section

>A specialized kind of section is type: it is used for documenting classes. The functions (or fields) within a type section are considered to be the methods of that class.

Does that help in this situation?

scazan:
[...]

PR #971, opened 2019-12-28
Use classmod for class, adding type annotations to most params, document some undocumented modules/classes, small typos here and there.

scazan:
Apologies for the large looking PR but I think it is hopefully a fairly small set of related changes.

This updates #841 and also adds type annotations to a number of ldoc comments (I find these very useful when referencing the docs). Since lua is prototype-based, the difference between calling a static and instance method is not differentiated except that the instance methods provide some syntax sugar for omitting an argument in the function signature. So I didn't manually differentiate the static ones from the instance ones unless it was very clear. I did make note of the constructors where it needed it and since the constructors actually generate an instance and operate upon that instance I left them as methods.

I also documented a few undocumented classes/modules.  

In most cases I only added a type annotation if I knew for sure that the type was correct. I also tried to leave most comments undisturbed as the author originally intended.

No actual code was touched.

Lastly I added in some redirects on the previous module docs to avoid any broken links from external links to the docs (hopefully I didn't miss some).

Feel free to cherry pick, update, or reject parts of this as needed.


tehn:
super super impressive, thank you! i'll have a good look by the end of the weekend.

scazan:
great! 

One note is that I didn't document a few things that I figured were maybe not meant to show in the api docs (such as fs and menu).

PR #973, opened 2020-01-02
Update to ignore Fates devices for kernel updates

okyeron:
Does this get the job done?

### DIFF
--- a/update/191230/update.sh
+++ b/update/191230/update.sh
@@ -37,7 +37,7 @@ fi
 
 # ignore kernel and samba update for Fates devices
 FATES=$(sudo cat /boot/config.txt | grep fates)
-if [[ "$FATES" == *""* ]]; then
+if [[ "$FATES" == "" ]]; then
   # samba
   SAMBA=$(dpkg -s samba | grep "install ok")
   if [ "$SAMBA" == "" ]; then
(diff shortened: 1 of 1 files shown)


okyeron:
Went to lunch and just back to test it now. Need to reflash the shield as a comparison.

tehn:
i'm going back on this in favor of my initial recommendation after some further discussion. we'd like not to add complexity, create compromises, or add technical debt to future maintainability: basically we don't want to have fates-specific code in the repo.

here's how we'd like to see fates continue:

- fork the norns repo
- change the SYSTEM > UPDATE procedure so that it does something simple like `git pull; ./waf` --- or simply remove SYSTEM > UPDATE entirely and have people manually update their norns folder.
- manually sync your fork to the master repo (you can have some other fates collaborators help you stay on top of this)

---

my earlier sentiment about not wanting a bunch of forks was the hope that people would be contributing code that benefits the project broadly instead of focusing on features that wouldn't be compatible with the main project. i think this situation where the update process is incompatible is a perfect use of forking.

okyeron:
Hmmmm. OK. 

FWIW - I just finished running the second patched version of this and it worked fine on both Fates and the Norns Shield.

Back to our previous lines conversation - to not have fates specific content in the main norns repo - I thought it would be good to check for known norns things like the kernel version (uname -r) or something similar and then skip the kernel update if it's NOT norns. That feels it would not create technical debt.

PR #977, opened 2020-01-14
Crow 1.0.3 syntax update

trentgill:
Adds a bunch of new aliased commands to norns support of crow.

These changes are untested currently, and the `done` and `running` support for output library needs proper testing. Perhaps `running` shouldn't be available on norns, and just use `done` instead.

DO NOT MERGE until `ii` support is updated.


trentgill:
It occurs to me that the `ii.<module>.get()` functions haven't been aliased on norns (except for meadowphysics). This means that the only way to query i2c devices from norns is using the `crow.send()` function. Is this an accurate analysis of the situation?

I can add this support, but it's a bit complicated because the `get` function takes varargs, as different i2c requests take different argument counts.

//

otherwise, i've updated the ii support for all the currently supported modules (it was a lot!) also updates a number of existing modules with additional commands that were added piecemeal along the way.

there is a new PR on the crow repo with the script i wrote that generates these stubs.

tehn:
re: ii get functions. indeed we should add them--- no problem if it's manual typing work.

PR #980, opened 2020-01-19
curses maiden ws compat

### LINKED ISSUE #978: revive command-line maiden
code is here: https://github.com/monome/norns/tree/master/maiden

needs to be updated to use WS instead of IPC

ngwese:
fixes #978 

this is a minimal change that allows the original curses based `maiden` to work along side the web based version. through the magic of nanomsg both the console repl and the web repl work at the same time. input is not echoed to each repl but each repl does see the same output.

_note: to compile i needed to `sudo apt-get install libncursesw5-dev`_

an outstanding issue which this PR doesn't address is the duplicate command names. perhaps it is worth reimplementing this in go and rolling it into the maiden backend command as `maiden repl`. one advantage of integrating it in the go based command would be the ability to run it on a different host (along side a non-web based editor).

### DIFF
--- a/maiden/src/io.c
+++ b/maiden/src/io.c
@@ -39,17 +39,15 @@ void *crone_rx_loop(void *x);
 struct sock_io sock_io[IO_COUNT];
 
 char *url_default[IO_COUNT] = {
-  "ipc:///tmp/matron_out.ipc",
-  "ipc:///tmp/matron_in.ipc",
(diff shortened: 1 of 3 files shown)


tehn:
super super rad! thanks for taking the time.

ngwese:
...in the end after i understood the basic structure of the code it ended up being simple. the main difference was that the `ws` sockets are bi-directional. it was probably 45 min work in the end.

PR #981, opened 2020-01-21
Bufdiskworker fixes

catfact:
small but potentially important fix for incorrect temp buffer allocation.
i haven't tested this at all yet, but opening for visibility since it's potentially important.


catfact:
this seems to be working fine. i haven't tested on norns hardware. thing to watch for is memory leaks; i'd use `valgrind --check-leaks` on crone process. but the change is pretty simple so i'm not too worried.

tehn:
will valgrind now

catfact:
no, there's a leak, sorry. (when request duration exceeds source.)
fixing...

catfact:
... better.

PR #990, opened 2020-01-28
fix softcut.defaults() (again)

AndrewShike:
sorry for double PR !

should be `state[i] = {}` not `state[i] = 0;`

attempted to clean up `level_input_cut`& `level_cut_cut`. I think this was using multiple assignment [wrong](https://www.lua.org/pil/4.1.html), the value was just `i`. made them arrays of zeros, perhaps more helpful

tested via copy and paste this time


tehn:
good catch. fix the merge conflict first?

AndrewShike:
I think I did it ... learning !

AndrewShike:
also: I'm showing `level_input_cut` defaulting to 0, but not sure if that's actually the case.

catfact:
this `defaults()` function looks extremely messed up in many additional ways

catfact:
just let me fix it please

Issue #920, opened 2019-10-16
occasional segfault on USB device removal


tehn:
seen a few times now. matron locks up or crashes on grid cable removal.

will add debug info when i get it.

may be related to https://github.com/monome/norns/pull/886 which touched a lot of device management (?)

tehn:
running `build/matron/matron`

```
grid added:	1	monome 128 m1000464	m1000464
grid added:	2	monome 128 m1000464	m1000464
grid added:	3	monome 128 m1000464	m1000464
grid added:	4	monome 128 m1000464	m1000464
grid added:	5	monome 128 m1000464	m1000464
grid added:	6	monome 128 m1000464	m1000464
grid added:	7	monome 128 m1000464	m1000464
grid added:	8	monome 128 m1000464	m1000464
grid added:	9	monome 128 m1000464	m1000464
grid added:	10	monome 128 m1000464	m1000464
Segmentation fault
```

time to add a bunch of debug prints to the device system

tehn:
to confirm, this only happens on removal.

trying to  identify the segfault location, it seems the whole removal process completes--- through weaver and monome device closure. the segfault occurs momentarily after completion of everything.

---

i just rolled back to 6551cba63bc3f91de605f780ac7517f5aa3948ed which is before https://github.com/monome/norns/pull/886 was merged, which is the last thing to touch the device code.

i did 100 usb plug cycles without a segfault. previously my max was about 30, sometimes it happened within 1-5.

@ryanlaws @artfwo @catfact could you help re-review this PR for potential problems? https://github.com/monome/norns/pull/886

ryanlaws:
Unfortunately I do not have a grid to test with. Can you repro with HID? I will test with that when I get home.

tehn:
just replicated with an HID keyboard after 20 insert/removes

ryanlaws:
Thank you. Won't be able to look at this for ~8 hours.

tehn:
no rush--- thanks for your time!

ryanlaws:
I'm unable to repro :slightly_frowning_face: 

Well... what are you thinking? Check whether `dn == NULL` before checking its `d` property?

ryanlaws:
Also what are your HID repro steps? This is what's **not** working for me:
1. Open `hid-demo/hid-keyboard-test`
2. Plug in keyboard (Leopold Otaku 101), wait for lights to come on
3. Type, confirm keystrokes visible
4. Fiddle with system menu if keys don't show up
[...]

PR #992, opened 2020-01-29
likely fix for SEGV on USB device removal

ngwese:
i spent a bit of time trying to get to the bottom of #920 - in particular looking at the changes which were made to the device management for handle multiple midi endpoints.

i believe the problematic change was the introduction of `dev_remove_node`. the `dev_remove_node` function combined existing deletion event posting, `dev_delete`, and device list node removal _with a call to `dev_lookup_path` starting from the device list node being removed/deleted_. the `dev_lookup_path` function walks the device list looking for the first node which matches the given path. in the case of `dev_remove_node` that lookup was started from the device node which was previously passed to `dev_delete`. the `dev_delete` function does several things but in particular it frees memory allocated for the device "path" string pointed to by the node. the net effect is a use after `free()` for one of the strings being fed to `strcmp(...)` and a potential source of crashing.

this change:
- factors out the call from `dev_lookup_path` from `dev_remove_node`
- renames `dev_remove_node` to `dev_list_remove_node` for clarity
- only performs exhaustive searches for multiple device nodes in the case of midi

open questions
- in the case of *midi* devices the repeated searching for matching devices always starts from the head node in the list which means that removal has O(num_endpoints * total_num_devices) complexity (i.e. exponential). in practice this didn't seem to be noticeable when hot pulling 1x grid, 2x single endpoint midi, and 1x four endpoint midi devices...
- the complexity could be reduced to O(num_endpoints) if `insque` and `remque` guaranteed that nodes wouldn't be moved in memory when the list was being modified. i think this is the case but the man page didn't make any statements regarding pointer stability.

### DIFF
--- a/matron/src/device/device_list.c
+++ b/matron/src/device/device_list.c
@@ -122,12 +122,7 @@ void dev_list_add(device_t type, const char *path, const char *name) {
     }
 }
 
-struct dev_node *dev_remove_node(
-    struct dev_node *dn, 
-    union event_data *event_remove,
-    const char *node
-) {
-    struct dev_node *next_node; 
+static void dev_list_remove_node(struct dev_node *dn, union event_data *event_remove) {
     event_post(event_remove);
 
     if(dq.head == dn) { dq.head = dn->next; }
@@ -136,10 +131,7 @@ struct dev_node *dev_remove_node(
     dq.size--;
 
     dev_delete(dn->d);
-    next_node = dev_lookup_path(node, dn);
     free(dn);
-
-    return next_node;
 }
 
 void dev_list_remove(device_t type, const char *node) {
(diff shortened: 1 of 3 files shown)


ngwese:
update: fair amount of interactive hot plug testing with multiple grids and midi devices and usb hubs - no crashing. previously with this same setup i was crashing 80% of the time i removed a usb device.

pq:
💎 💎 💎

ryanlaws:
Excellent!!!

ryanlaws:
Thank you for your testing and for fixing this. This bug made me sad whenever I thought about it and I had a hard time reproducing it. I'm going to sleep a little easier now.

okyeron:
Query - is this related to the `usb 1-1.3: urb status -32` errors I see in `dmesg` when disconnecting USB devices (on Raspi)?

ngwese:
no - this change is specifically focused on preventing `matron` from crashing when USB devices are disconnected.

the significance of the `urb status -32` and the sequence of events which leads to them is a separate line of investigation.

okyeron:
I did not see an issue for the  `urb status -32` thing so please let me know if/where I can help with that - I seem to be able to reproduce that pretty regularly.

PR #996, opened 2020-01-30
fix softcut typo and more docstrings

catfact:
fix a typo in the last PR that prevented the softcut lua module from loading.

also finish lots more docstrings for that module.


catfact:
i can confirm this at least doesn't error when the module is included. i'm gonna merge so that people's scripts can continue functioning.

PR #997, opened 2020-01-30
use softcut submodule

catfact:
since softcut is starting to have ports and there are big features coming up, i moved it to its own repo and added as a submodule.

i also tried to build and link it as a static library. but i can't actually figure it out with `waf`... (help! see comments in `crone/wscript`)

i did update the cmake build for crone and that works fine.


catfact:
works fine as source library so will just go with that for now.

only minor downside is that `crone` has to be modified if private sources of `softcut` change, and potentially needs to link different libs/standards than would otherwise be needed. 

(tested several softcut scripts on norns hardware.)

catfact:
also, ehh, no need to merge this until there are actually useful changes to the lib.

catfact:
okey dokey..

one caveat: now that there are submodules, a simple `git pull` may not suffice; `git pull --recurse-submodules`, or similar, instead...

PR #999, opened 2020-01-30
build softcut-lib from submodule

artfwo:
currently fails to build because of a few existing class redefinitions and mismatching method signatures. @catfact can you have a look?


catfact:
ah ok - so in the other branch i was trying to link `softcut` as a static lib rather than just using the sources from the submodule.

this almost worked - `softcut-lib/softcut-lib/wcript` builds the static lib, and i can link that with my test project or with `crone` if i use cmake. i just can't figure out how to make `crne.wscript` work the same as `crone/CMakeLists.txt`.

maybe that's a waste of time and a red herring, and i should compile `softcut-lib` sources directly in the parent project as you do.

artfwo:
@catfact a static lib is just an archive of object files, so there's no difference in having it built as static lib or objects later linked into crone binary (unless you want different CFLAGS).

unfortunately i can't commit into your fork, but can update this PR to include your changes to crone sources if you like.

catfact:
its no problem, i'll just biuld from source in the parent.

the motivation for making a static lib is, as you say, to encapsulate dependencies (on standards, other libraries, etc)

but you're right, in this case it's sort of pointless.

thanks for your help!

catfact:
changed PR 997 to use sources instead of stlib

PR #1000, opened 2020-01-31
softcut hotfix: re-add post_filter_fc_mod

tehn:
sorry about the ldoc noise (ldoc is not very smart)

important change is the last file.


catfact:
ohhhh

ok. the thing is, i removed that on purpose, because there is no modulation on post-filter and the inclusion of an API function for it was a mistake

i did not expect anyoe to be using it.

tehn:
no worries, thanks!

Issue #1003, opened 2020-02-06
crone commands are effectively overwritten within audio block


catfact:
commands sent to each `crone` are stored in a FIFO queue. 

each client empties its command queue at the top of each audio block and processes all pending commands.

this is fine, except in cases where commands are intended to have persistent side effects in the DSP, and they are effectively overwritten if the relevant value is changed again before the change is executed.

a good example is multiple settings of "slew" and "level" pairs.

the user sends something like (psuedocode): `slew 0; level 0; slew 10; level 1;`

the expectation is:
- immediately jump to level 0 by first setting slew time=0, then setting new level=0
- then, start a 10-second ramp to level 1 by first setting slew time=10, then level=1

but what happens is: the first [time,target] pair sets a target and a filter coefficient, but those values are immediately overwritten before any samples are processed by the filter, 
so you only get a long ramp to 1.0 from whatever the current value happened to be.

PR #1004, opened 2020-02-06
only handle one command of each type per audio block

catfact:
one possible approach to addressing issue #1003, for your consideration.

only one command of each type is processed from the command queue for each audio block.

if a command's type has been seen already, it is pushed back on the queue to be processed on the next block. processing order should be unchanged.

possible downside is if you're sending really large numbers of duplicate commands, they will end up spaced about 3ms apart in time (or whatever.) this might have weird effects for knob sweeps.

maybe more specific logic is needed, just for commands that have "side effects" for other params, requiring DSP time to execute. (e.g. slew time,) and those that are simple value setters should in fact keep the overwriting behavior.

### DIFF
--- a/crone/src/Commands.cpp
+++ b/crone/src/Commands.cpp
@@ -38,14 +38,22 @@ void Commands::post(Commands::Id id, int i, int j, float f) {
 
 
 void Commands::handlePending(MixerClient *client) {
-    CommandPacket p;
+    CommandPacket p{};
+    memset(wasHandled, 0, NUM_COMMANDS * sizeof(bool));
     while (q.pop(p)) {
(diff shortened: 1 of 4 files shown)


tehn:
aha, this is a good idea. when you're ready i can do some testing to assess usability (ie, re: knob sweeps etc)

catfact:
can be tested now for sure

catfact:
@tehn any thoughts on this? if it seems useful i will resolve conflicts. otherwise we can just close.

catfact:
i'm closing this. it's a weird thing to do and there are better ways to target the desired functionalities more directly.

PR #1005, opened 2020-02-07
expose extents

antonhornquist:
i use this in my scripts so would be good to use it from `screen` global rather than `_norns.screen_extents`.


pq:
yes please!

simonvanderveldt:
Shouldn't this be called `text_extents` like the others?

antonhornquist:
I’m all right with changing this. @tehn ?

tehn:
probably a good idea

PR #1006, opened 2020-02-08
rename function

antonhornquist:
https://github.com/monome/norns/pull/1005#issuecomment-583657182

let's call it `screen.text_extents` instead.

### DIFF
--- a/lua/core/screen.lua
+++ b/lua/core/screen.lua
@@ -165,7 +165,7 @@ Screen.text_center = function(str) _norns.screen_text_center(str) end
 --- calculate width of text.
 -- uses currently selected font.
 -- @tparam string str : text to calculate width of
-Screen.extents = function(str) return _norns.screen_extents(str) end
+Screen.text_extents = function(str) return _norns.screen_extents(str) end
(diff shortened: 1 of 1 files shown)


lua/core/screen.lua
```diff
@@ -165,7 +165,7 @@ Screen.text_center = function(str) _norns.screen_text_center(str) end
 --- calculate width of text.
 -- uses currently selected font.
 -- @tparam string str : text to calculate width of
-Screen.extents = function(str) return _norns.screen_extents(str) end
+Screen.text_extents = function(str) return _norns.screen_extents(str) end
```
simonvanderveldt: Would it make sense to rename the "internal" function (`_norns.screen_extents()`) as well?

antonhornquist:
i consider that to be a breaking change - scripts that use `_norns.screen_extents` will break. then, this is probably undocumented/non-public API so i have no problem changing this. will fix today. how about merging this and making a new PR for the `_norns.screen_extents` references (it's a bigger change since it's used in places in core right now)?

PR #1007, opened 2020-02-09
maiden-repl, rename and polish build

ngwese:
renames the curses based `maiden` to `maiden-repl` in an effort to avoid confusion with the `maiden` http server backend binary.

- adds a basic cmake build which works on linux and macos (assuming the needed dependencies are installed via homebrew)
- adds a wscript and integrates it into the top level build along side `matron`, `crone`, etc.
- provides both zero arg `CHECK(...)` macro and multi arg `CHECKV(..)` macro in order to get past more stringent compilers (clang on mscOS) with `-Werror`


ngwese:
...i realized i has already made the companion change where the `maiden repl` subcommand will exec `maiden-repl` binary.

PR #1012, opened 2020-02-17
fix for skipped remainder frames in softcut read-buffer commands

catfact:
fix a typo in read-buffer routines that resulted in post-block remainder frames being skipped with an error.


catfact:
sorry wrong branch or something

Issue #1010, opened 2020-02-14
Grid rotation function may need to update rows/cols


okyeron:
`grid.lua` grid rotation function may need to update `_norns.grid_rows` and `_norns.grid_cols` when it changes orientation.

If a script is using something like `grid_w = grid_device.cols` and `grid_h = grid_device.rows` these do not get updated when changing orientation. Thus 90 and 270 degree rotations can end up only showing half of a 128 grid.

Will do some further research.

tehn:
grid rotation still hasn't received much attention so it'd be rad if you have time to put towards improving it!

PR #1018, opened 2020-02-20
grid vports not removing rows/cols on update_devices

okyeron:
Found a situation where grid vports was reporting rows/cols values when that vport was set to `none`.

Test situation:
- with all grid vports set to none, attach grid.
- remove grid and set vport 1 back to none
- `tab.print(grid.vports[1])` then returns the following:

```
cols	16
name	none
all	function: 0x1f79248
led	function: 0x1f6bd98
rows	8
refresh	function: 0x1fa7e60
intensity	function: 0x1fa8548
rotation	function: 0x1fa8590
```
Thus `name` is set to none, but `rows` and `cols` are not reset to zero

Can also be tested with my [Grid Test script](https://llllllll.co/t/grid-test/29346) in the lines Library when manually changing grid port positions.

Proposed fix: reset rows/cols to zero in Grid.update_devices()

### DIFF
--- a/lua/core/grid.lua
+++ b/lua/core/grid.lua
@@ -145,6 +145,8 @@ function Grid.update_devices()
   -- connect available devices to vports
   for i=1,4 do
     Grid.vports[i].device = nil
+    Grid.vports[i].rows = 0
+    Grid.vports[i].cols = 0       
 
     for _,device in pairs(Grid.devices) do
       if device.name == Grid.vports[i].name then


catfact:
seems fine, but i'm also unclear on where this is actually a problem? 

if a script or system component wants to test whether a vport is really connnected, it should test `.name` and/or `.device` fields, no?

(and yea indeed, i wouldn't mind a little more documentation on vports and how these fields are intended to be used.)

okyeron:
I don't know if it's actually a "problem" anywhere just yet. 

I just thought it was odd that a device set to "none" was reporting cols/rows.

Note there quite yet, but this may come into play with resetting/reporting cols/rows for grid rotation in #1010

PR #1020, opened 2020-02-25
keystroke activated watchdog

### LINKED ISSUE #1014: clean up / document component management methods
recently there were some issues reported on the forum related to the way program lifecycle management works on norns (or doesn't work.) 

it made me realize that the use cases and the final design concept are fuzzy to me, and i guess to most users. 

here are a few things that can be used to reset parts of the norns stack:

- 1. SYSTEM > SLEEP. 
  - saves state
  - zeros audio output levels 
  - performs `sudo shutdown`

- 2. SYSTEM > RESET
  - saves state 
  - resets systemd services `norns-matron`, `norns-crone`, `norns-sclang`

- 3. issuing `;restart` to the **matron** repl in maiden.
  - resets only `norns-matron` service (i think!)
  - immediately reconnects maiden to matron websocket, catching most startup output
  - this will cause a lot of runtime problems, since matron needs metadata about running engines/polls.
  - the only time i can think of needing this, is in the diagnosis of startup errors in matron sources or system lua (and for some reason you can't or don't want to use the shell.)

- 4. issuing `;restart` to the **sc** repl in maiden.
  - resets only `norns-sclang` service (i think!)
  - immediately reconnects maiden to sclang websocket, catching most startup output
  - as far as i know, things should basically keep working after doing this, but the architecture doesn't actually guarantee it.
  - i think this is useful in the diagnosis of startup errors in supercollider, which includes any errors in custom engine classes. (though again one can just use the shell also.)

- 5. AUDIO > RESET. as far as i can tell, this is **no longer exposed** in the menu, but the underlying API calls (`_norns.reset_audio` in lua, and on down) are still available:
    - soft-resets supercollider environment by performing `Interpreter.recompile`, which also relaunches the scsynth process.
   - tells matron to reperform handshake with sclang.
   - so in a nutshell, this should guarantee that things keep working, but could of course be bugged at this point. (this was more relevant pre-2.0.)

- 6. rude power button. should result in a false `clean_shutdown` flag, meaning next launch will not run any script, restore mix levels, or (importantly) restore device mapping. **this is still the only way for a typical (non-commandline) user to get out of many bugged situations.**  (e.g. a bug in the default script, bug in our handling of vports with nonexistent devices, anything that locks up the UI.)

in other words, i guess i think we need to clearly inform users about which method to use for what circumstance. (troubleshooting, recovering, updating etc.)? and we may want to refine the offerings. for example:
- if the "audio reset" behavior is no longer relevant post-2.0, let's remove it completely.
- we may want a way for users to perform a "clean boot" without the rude power button. ideally it would be implemented at a lower level than the main application loop, e.g.:
    - if all three keys are held (say), we detect this in `hardware/gpio` module and launch a special timeout thread. 
    - if timeout thread detects a 6s hold of all keys (or whatever), matron immediately executes a soft-reset of the stack via `systemd`, without saving any application state or writing `clean_shutdown`.

what do yall think?

tehn:
further fixes #1014 

in sequence, hold K3, K2, K1. if 10s elapses, services are restart.

however, weird thing i'm seeing:

despite `systemctl restart norns-*` it seems like only matron is being restarted? ie, i don't get the startup animation and things start almost immediately. running the same command on the command line does what we want?

### DIFF
--- a/matron/src/watch.c
+++ b/matron/src/watch.c
@@ -43,7 +43,9 @@ void *watch_time(void *x) {
     if(stage==3) count++;
     if(count==10) {
       fprintf(stderr, "RESTARTING...\n");
-      system("nohup systemctl restart norns-* > /dev/null");
+      system("nohup systemctl restart norns-sclang > /dev/null");
+      system("nohup systemctl restart norns-crone > /dev/null");
+      system("nohup systemctl restart norns-matron > /dev/null");
     }
     sleep(WATCH_TIME);
   }
--- a/matron/src/watch.c
+++ b/matron/src/watch.c
@@ -1,0 +1,57 @@
+/*
+ * watch.c
+ *
+ * tracks key state, resets after timer
+ */
+
+#include <stdio.h>
+#include <stdint.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <pthread.h>
+
+#include "events.h"
+
+#define WATCH_TIME 1
+
+static int count = 0;
+static int stage = 0;
+
+static pthread_t p;
+
+void *watch_time(void *);
+
+// extern def
+
+void watch_init() {
+  if (pthread_create(&p, NULL, watch_time, 0) ) {
+    fprintf(stderr, "WATCH: Error creating thread\n");
+  }
+}
+
+void watch_deinit() {
+    pthread_cancel(p);
+}
+
+void *watch_time(void *x) {
+  (void)x;
+
+  while(1) {
+    if(stage==3) count++;
+    if(count==10) {
+      fprintf(stderr, "RESTARTING...\n");
+      system("systemctl restart norns-*");
+    }
+    sleep(WATCH_TIME);
+  }
+}
+
+void watch_key(int n, int z) {
+  if(z==0) { stage = 0; }
+  else if(stage==0 && n==3) { stage = 1; }
+  else if(stage==1 && n==2) { stage = 2; }
+  else if(stage==2 && n==1) { stage = 3; }
+}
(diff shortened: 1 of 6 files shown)


catfact:
I'll take a look later.. since matron is parent process of shell it's probably getting shut down prematurely?

matron/src/watch.c
```diff
@@ -0,0 +1,57 @@
...
+}
+
+void *watch_time(void *x) {
+  (void)x;
+
+  while(1) {
+    if(stage==3) count++;
+    if(count==10) {
+      fprintf(stderr, "RESTARTING...\n");
+      system("systemctl restart norns-*");
```
catfact: yeah, pretty sure the plain `system("systemctl ...)` won't work because as soon as matron is stopped the child shell will also be killed.

this SO question has some ideas, namely:
- `system("setsid .... &")`
- something suspiciously complicated with `fork()`, `execl()`, `waitpid()` and so on

ngwese:
Another possibly more simple option (not tested) would be to define another systemd service which watches for changes to a specific file on disk and runs the service as a result.
https://www.linux.com/tutorials/systemd-services-monitoring-files-and-directories/

...one could imagine the file watcher service simple restarts the `norns-*` units and all matron needs to do is touch a file (which the watcher service or matron could clean up after restart)

tehn:
pretty sure i fixed it with a `nohup` (i mean, testing, it works now)

tehn:
anyone see a problem with this? tested and confirmed working. i should probably try more crash-scenarios, if anyone has suggestions

simonvanderveldt:
I haven't tried this but can't we just restart `norns.target`?
(And we have a restart somewhere else which I believe just restarts a single unit and used the declared unit dependencies to restart the rest. Wasn't this the restart in the menu? Why don't we do the same here?

tehn:
changed systemctl restart messages according to discussion w/ @catfact and @ngwese

PR #1021, opened 2020-02-27
add missing dependencies and git submodules

widdly:
The instructions didn't work for me.  The ncurses libraries are missing and the softcut submodule needs to be initialized.


catfact:
thanks!

PR #1025, opened 2020-03-05
Tape branch

widdly:
tape playback automatically loops if file is longer than 1s
mono files are supported

There is a problem with playback position in GUI as it drifts over time while looping, especially on shorter files.  It should probably use a callback.


tehn:
cool!

i believe @catfact has some similar reworking on TAPE in process so i want to make sure there isn't any overlap

catfact:
hey thanks! 

i did have a larger rewrite of the tape class going. in particular it would enable pause and would insert envelopes at pause and loop points to avoid clicks. (also just cleaner and more consistent)

but this is great for now!

PR #1035, opened 2020-03-20
Possible fix buf clear region channel

catfact:
blind attempt to fix what looks like copypasta error in `oracle.c`

report: https://llllllll.co/t/cheat-codes-v1-2/30329/18?u=zebra


djstayhome:
Hey there, 

I've been experiencing a similar issue (I think) - related to this and https://github.com/monome/norns/issues/964 whereby audio samples I'm loading into a norns script that I'm currently creating, are not looping properly - there's roughly a ~10% silence at the end of the file. I'm using Claude code for the scripting, as I'm not an engineer and have spent several hours trying to pin this issue down. Have you found a way round it? 

Would appreciate any input on the below which may help 🙏

Notes as follows

(~10% missing)
Problem
softcut.buffer_read_mono is not reading the complete file - approximately 10% of the audio is missing from the end, resulting in silence before the loop point. This occurs despite commit 7d409ab (which fixed #964 in February 2020) being present in the current Norns version.
Environment

Norns version: 231114 (latest, checked for updates on 2026-03-01)
Confirmed commit 7d409ab is present (SF_SEEK_CUR → SF_SEEK_SET fix in BufDiskWorker.cpp)
Test file: ChurchBells.WAV (61.35s, 44.1kHz, stereo, 2705728 samples)

Reproduction
Code
lualocal ch, samples, rate = audio.file_info(path)
local duration = samples / rate  -- Returns 61.354376417234

softcut.buffer_read_mono(
  path,
  0,              -- start_src: read from beginning
  0,              -- start_dst: write to buffer position 0 (voice 1, no offset)
  duration,       -- dur: 61.35 seconds
  0,              -- ch_src: mix stereo channels
  1               -- ch_dst: buffer 1
)

-- Set loop points
softcut.loop_start(1, 0)
softcut.loop_end(1, 61.35)
softcut.position(1, 0)
softcut.play(1, 1)
Expected Result
File plays for 61.35 seconds, then loops back to start.
Actual Result
File plays for ~55 seconds, then 6 seconds of silence, then loops back to start.
Observations

Gap is proportional to file length (~10% for all tested files)
Loop points are set correctly (verified via debug logging)
Same file plays correctly in Audacity/VLC (no gaps)
Old buffer data persists between loads (suggests buffer isn't being cleared, but also isn't being fully written)

Debug Output
File info:
  Channels: 2
  Samples: 2705728
  Rate: 44100 Hz
  Calculated duration: 61.354376417234s

buffer_read_mono call:
  start_src = 0
  start_dst = 0.0
  dur = 61.354376417234
  ch_src = 0
  ch_dst = 1

Loop points set:
  loop_start = 0.0
  loop_end = 61.354376417234

Result: Audio plays ~55s, then 6s silence, then loops
What We've Tested
All failed to fix the issue:

❌ buffer_read_stereo instead of buffer_read_mono
❌ ch_src = 1 (left channel only) vs ch_src = 0 (mix both)
❌ Sample rate compensation (48000/44100 ratio)
❌ Adding 10-20% padding to duration
❌ Using dur = -1 (read entire file) - causes buffer overflow
❌ Testing with mono files
❌ Testing with 48kHz files

Theory
The commit 7d409ab fixed seeking within the read loop (SF_SEEK_CUR → SF_SEEK_SET), but there may be additional frame counting bugs that prevent the loop from reading the complete file.
The consistent ~10% gap suggests:

Mathematical error in total frame calculation, OR
Read loop terminates early, OR
Duration parameter is being interpreted incorrectly for non-zero start_dst (though we tested with start_dst = 0)

Questions

Was #964 completely fixed, or only partially?
Are there known issues with buffer_read_mono for files >60 seconds?
Should we reopen #964 or create a new issue?
Do other scripts (mlr, ndls) work around this somehow?

Impact
This blocks any script that loads long audio files for looping. Currently affects development of a 4-voice looper script.
Full Debug Log
Complete debugging session with all attempted fixes available here:
[Attach AUDIO_GAP_DEBUGGING_SUMMARY.md]

Can confirm: This is NOT a scripting error - loop points are correct, file info is correct, all parameters are correct. The buffer simply doesn't contain the full file after buffer_read_mono completes.

PR #1037, opened 2020-03-22
pset save flag per param

tehn:
added a param flag for saving to pset (or not)

in this case, we wanted _not_ to save mixer levels in the pset

```
params:set_state("output_level", false)
```

etc. by default now the levels are not saved. i did not extend this to reverb/comp params.

i think this is fine by default, but i suspect there's going to be a request (which is reasonable) that all params get some sort of top-level UI for save-to-pset or not (like the aleph, which was a powerful feature).


okyeron:
FWIW I would love to see this for reverb 

It seems reverb state is reset to on by default on dirty shutdown (RESET) so I’m regularly having to turn it off when reassign

tehn:
on/off is reasonable for reverb and comp, will add

PR #1040, opened 2020-03-24
docs update v200323

tlubke:
- added explicit section names to intonation.lua and ui.lua (ldoc error when nil)

I did this from my computer, and it seems that my lua version is different from the last time docs were generated. Is norns on lua v5.1? If that's the case I made need to run ldoc from my norns instead.


catfact:
thanks!

> Is norns on lua v5.1

no, it's 5.3

tehn:
i typically run this on-device--- not sure why it was giving 5.1

but this PR looks great, thanks!

PR #1041, opened 2020-03-25
lua filters module

catfact:
i thought it would be good to add some filters to the norns lua lib that would be useful for processing musical events. so here's this for consideration..

so far just added sliding mean and sliding median. easy to add a few more?


tehn:
looks great!

lua/lib/filters.lua
```diff
@@ -0,0 +1,158 @@
...
+   return y
+end
+
+--- clear a filter's history
+function f:clear() 
+   for i=1,self.bufsize do self.buf[i]=0 end
+end
+
+-- debug function to print the buffer
+function f:print_string(pre)
```
ngwese: For what it is worth if we do find value in printable representations of lua objects it might be worth consistently using implementing the `__tostring` meta method:
https://www.lua.org/pil/13.3.html

I personally see value in implementing `__tostring` in may places to make it easier to interactively debug stuff.

PR #1047, opened 2020-03-26
conditionally compile link clock bindings

artfwo:


catfact:
i think there are some other calls too?

```
matron/src/main.c.1.o: In function `main':
/home/we/norns/build/../matron/src/main.c:79: undefined reference to `clock_link_start'
matron/src/weaver.c.1.o: In function `_clock_link_set_quantum':
/home/we/norns/build/../matron/src/weaver.c:1444: undefined reference to `clock_link_set_quantum'
matron/src/weaver.c.1.o: In function `_clock_link_set_tempo':
/home/we/norns/build/../matron/src/weaver.c:1437: undefined reference to `clock_link_set_tempo'
```

maybe i'm doing something wrong... i did `waf clean`, `configure`, `build`

artfwo:
strange, in main they're already surrounded by #ifdefs

catfact:
yea sorry must be on my end

artfwo:
just reproduced on my end as well - must be some bad waf scripting from my side, working on it.

artfwo:
sorry, haven't tested the no-link configuration enough. should build fine now both with and without link.

Issue #1023, opened 2020-02-28
osc in for remote/emulated key/enc action


tehn:
to facilitate remote control and/or emulation

```
/norns/key n z
```

calls

```
key(n,z)
```

PR #1062, opened 2020-04-05
osc remote for key/enc

tehn:
fixes https://github.com/monome/norns/issues/1023

added the `/remote/` OSC pattern

- `/remote/key 2 1` --- effectively executes `key(2,1)` at the lowest level (menu or play mode)
- `/remote/enc 1 -3`

same OSC port: 10111

important to remember that with key remotes it should be treated like a state-ful toggle. ie, don't send a bunch of key-downs in a row. some menu/script logic may get super weird since a script could be tracking down-up positions.

@dndrks need some docs for this

### DIFF
--- a/lua/core/osc.lua
+++ b/lua/core/osc.lua
@@ -79,10 +79,20 @@ local function param_handler(path, args)
   end
 end
 
+local function remote_handler(path, args)
+  if path=="/remote/key" then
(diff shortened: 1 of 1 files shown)


tehn:
side note--- this could help facilitate an "emulator" for building on desktop

PR #1069, opened 2020-04-06
maiden-repl: Fix CMake compilation with split tinfo(w)

simonvanderveldt:
This fixes the CMake based compilation of `maiden-repl` for me.
Would be good if someone on another distro than mine (gentoo) could test if it still works for them.

Simple build steps are:
```
cd maiden-repl
cmake .
make
```


ngwese:
I can take a look later tonight and verify it builds on buster. Was there a specific reason to bump the minimum cmake version up? It didn't seem like the changes made would have required it.

PR #1077, opened 2020-04-09
average intervals for incoming midi clock events

### LINKED ISSUE #848: averaging intervals for midi clock source events
incoming midi clock beat events can be received at irregular intervals, causing small de-synchronisation in the scheduler.

averaging intervals (last 40 or so) could help improve scheduling with midi.

artfwo:
fixes #848


artfwo:
the tempo looks much more stable with my midi clock source - op-1, but op-1 doesn't send out the "start" byte in a controlled way, so i can only sync actual melodies by ear :) must be good to merge after more testing with real world gear.

tehn:
tested with arturia keystep = extremely improved.

PR #1082, opened 2020-04-11
Crow library refactored & auto-reset

trentgill:
Refactored the `crow.lua` file.

* On norns script load, crow table is reset & crow's virtual env is reset
* After script load, a newly connected crow's virtual env is reset when detected (this behaviour is in the user-function `crow.add` so it can be modified for more complex setups)
* Fixed typo that made the calibration library non functional
* Added some missing i2c plumbing for more recently supported device's events
* Factored auto-generated i2c stubs into separate files so generated pieces no longer need to be hand-copied

Not yet tested on device.


trentgill:
@tehn not yet, but hopefully tomorrow! I do want to make sure the metatable stuff still works correctly, and the capturing of `^^` messages.

tehn:
tested, this feels good

PR #1087, opened 2020-04-13
clock.get_beat_ms

tehn:
default arg is 1 (beat), so it also functions as a simple helper ie `clock.get_beat_ms(3/2)`

ie bpm 120 returns 500ms

mainly for use in other params that want ms values... ie softcut loop time or cut position

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -111,9 +111,9 @@ clock.get_tempo = function()
   return _norns.clock_get_tempo()
 end
 
-clock.get_beat_ms = function(x)
+clock.get_beat_sec = function(x)
   x = x or 1
-  return (60000.0 / _norns.clock_get_tempo() * x)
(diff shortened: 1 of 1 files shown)


artfwo:
softcut-lib uses seconds, maybe this can return seconds as well? i'm not sure what units we're using in softcut.lua

tehn:
you are super correct

nothing uses ms, i will make it seconds

lua/core/clock.lua
```diff
@@ -111,6 +111,11 @@ clock.get_tempo = function()
   return _norns.clock_get_tempo()
 end
 
+clock.get_beat_sec = function(x)
+  x = x or 1
+  return (60.0 / _norns.clock_get_tempo() * x)
```
artfwo, line 116:
```suggestion
  60.0 / clock.get_tempo() * x
```

PR #1089, opened 2020-04-15
filter crow/midi input

### LINKED ISSUE #1088: global clock: improve averaging function for midi/crow
potential user behavior:
- boot norns where global clock is set to `midi`
- their ext midi clock source isn’t sending signal, because they aren’t ready to play yet
- user makes a sandwich, comes back to the studio and then starts their midi clock source and try to use a script like the one below — a straight-forward sequencer that uses clock.sync(1/4) to clock a 16th pulse.
- it will take a long while to start executing steps, though -- because 20 minutes of no signal have been averaged.

to test:
- start with CLOCK set to internal , 120bpm
- let it run for 30 seconds or so
- switch CLOCK to midi (do not have your ext. midi clock source playing)
- return to script’s screen
- start your ext. midi clock source

_what i’m seeing_: counter/redraw stalls for quite a bit of time (typically 12-20 beats), even though beats are still executing

_what i expected_: clock would start averaging when it receives a start message and dump old history, taking only a beat to stabilize.

```
function init()
  task_id = clock.run(test_clock)
  count = 0
end
function test_clock()
  while true do
    clock.sync(1/4)
    count = count+1
  end
end
clock.run(function()
  while true do
    clock.sleep(1/60)
    redraw()
  end
end)
function redraw()
  screen.clear()
  screen.level(15)
  screen.move(0,10)
  screen.text(params:string("clock_source")..": "..count)
  screen.move(0,20)
  screen.text(string.format("%.2f",clock.get_beats()))
  screen.move(0,30)
  screen.text(string.format("%.2f",clock.get_tempo()))
  screen.update()
end
```

tehn:
fixes #1088

basically rewrote the averaging function somewhat needlessly, as the primary fix was tiny.

the main improvement is filtering out huge values: infer a clock stop/start.

### DIFF
--- a/matron/src/clocks/clock_crow.c
+++ b/matron/src/clocks/clock_crow.c
@@ -11,48 +11,48 @@ static double clock_crow_last_time;
 static double duration_buf[DURATION_BUFFER_LENGTH] = {0};
 static uint8_t beat_duration_buf_pos = 0;
 static uint8_t beat_duration_buf_len = 0;
+static double mean_sum;
+static double mean_scale;
 
 #define CLOCK_CROW_DIV 4.0
 
 void clock_crow_init() {
     clock_crow_counter = 0;
     clock_crow_last_time_set = false;
+    mean_sum = 0;
 }
 
 void clock_crow_handle_clock() {
-    double beat_duration;
-    double beat_duration_buf_sum = 0;
-    double beat_duration_buf_avg;
-    double current_time = clock_gettime_secondsf();
+  double beat_duration;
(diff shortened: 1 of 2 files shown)


artfwo:
@tehn, why does this skip updating reference if `clock_midi_last_tick_time_set == false`?

tehn:
> @tehn, why does this skip updating reference if `clock_midi_last_tick_time_set == false`?

it made more sense to me to wait until the second tick so that an actual time is submitted, rather than using a dummy time for the first tick.

PR #1090, opened 2020-04-15
reconnect crow clock

tehn:
if clock source set to crow:

- successfully connects on script startup if crow already connected
- sets up clock if crow connected while script running

thanks for spotting this @simonvanderveldt

### DIFF
--- a/lua/core/crow.lua
+++ b/lua/core/crow.lua
@@ -16,6 +16,14 @@ _norns.crow = {}
 _norns.crow.add = function(id, name, dev)
   norns.crow.dev = dev
   crow.add(id, name, dev)
+
+  --- enable clock-in if needed
+  if params.lookup["clock_source"] then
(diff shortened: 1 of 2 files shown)


simonvanderveldt:
:+1:
Hate to be annoying, but why do we need specific code for this (twice) compared to MIDI, which doesn't seem to need it (from my quick search in the codebase) but still works?

tehn:
@simonvanderveldt 

it's not possible without reworking a bunch of other systems.

`_norns.crow.add` is called at startup, before the param clock system default is added. so this means each plugging situation (crow attached at startup vs. crow added mid-script) needs to be dealt with separately.

crow works differently than midi, because we need to send it a command to actually start sending sync. if midi were this way we'd run into a similar situation.

simonvanderveldt:
@tehn Clear, thanks for the additional explanation!

PR #1092, opened 2020-04-18
allow passing arguments to underlying coroutine from clock.run

artfwo:


ngwese:
is there an example of how to use this? i tried to get it to work but the arguments to my clock coroutine function are getting bound to any values.

i'm trying something more or less like the following
```lua
local _routine = function (this, yield)
  while true do
    this._tick = this._tick + 1
    print(this._tick)
    yield(this.interval, this, yield)
  end
end

local t = {
  _tick = 0,
 interval = 1,
}

function init()
  clock.run(_routine, t, clock.sleep)
end
```

when i do this i die at the first reference to `this`:

```
/home/we/norns/lua/core/clock.lua:65: /home/we/dust/code/sky/test/untitled.lua:3: attempt to index a nil value (local 'this')
stack traceback:
	/home/we/norns/lua/core/norns.lua:128: in function </home/we/norns/lua/core/norns.lua:128>
	[C]: in function 'error'
	/home/we/norns/lua/core/clock.lua:65: in function 'core/clock.resume'
	/home/we/norns/lua/core/clock.lua:23: in function 'core/clock.run'
	/home/we/dust/code/sky/test/untitled.lua:15: in global 'init'
	/home/we/norns/lua/core/script.lua:103: in function 'core/script.init'
	[C]: in function 'xpcall'
	/home/we/norns/lua/core/norns.lua:129: in field 'try'
	/home/we/norns/lua/core/engine.lua:91: in function </home/we/norns/lua/core/engine.lua:89>
```

ngwese:
i believe i've located the problem. fixed via #1093

PR #1096, opened 2020-04-20
add clock.create, clock.suspend

ngwese:
allows additional control over how and when clock coroutines
execute through explicit use of suspend and resume along side
existing sleep and sync scheduling

- automatically cancels clocks when their coroutine completes
- print warning when a canceled clock is resumed

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -12,14 +12,22 @@ local function new_id()
   return id
 end
 
+--- create a coroutine to run but do not immediately run it;
+-- @tparam function f
+-- @treturn integer : coroutine ID that can be used to resume/stop it later
+clock.create = function(f)
+  local coro = coroutine.create(f)
+  local coro_id = new_id()
+  clock.threads[coro_id] = coro
+  return coro_id
+end
+
 --- create a coroutine from the given function and immediately run it;
 -- the function parameter is a task that will suspend when clock.sleep and clock.sync are called inside it and will wake up again after specified time.
 -- @tparam function f
 -- @treturn integer : coroutine ID that can be used to stop it later
 clock.run = function(f, ...)
-  local coro = coroutine.create(f)
-  local coro_id = new_id()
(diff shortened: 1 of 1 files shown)


ngwese:
marking this as do not merge for now. i'm going to experiment more deeply to validate the idea.

artfwo:
@ngwese but dead coroutines cleanup looks good to go, how about creating a separate PR for it?

tehn:
any movement on this one?

ngwese:
i’m happy with the addition however the code i was working on which used this has been simplified such that it isn’t strictly necessary.

the use cases i was imagining generally involved creating a coroutine in one part of of a script and running/suspending it another  part without the two parts needing to know about each other. suspend is useful in this context because the consumer of the preconstructed clock does have to have the knowledge to recreate it. all of this is doable by wrapping an object around the clock but that seems like more ceremony than might be desirable for small scripts. 

i have been running with these changes for two days now with no ill effects. 

i’d be inclined to merge unless there is strong concerns.

tehn:
i see the value in this, even if it's an more niche use case.

ngwese:
pushing forward with merging this.

`create` has proven useful because it allows one to pass the coroutine/clock id into the clock itself (via `resume`) and the clock can subsequently signal other parts script of changes about itself (like that it is sync'd and entering its loop or suspending itself). being able to just the clock id removes the need to create another associative table of arbitrary id => clock id in order to identity which clock is signaling.

`suspend` is useful when one wants to suspend/resume a coroutine/clock is built around a loop which is stepping via an infinite iterator (treating it like an event queue) until some external variable signals it to stop. if the queue is full of note like values one likely wants the note off to be sent before the coroutine is stopped. keeping this book keeping inside the coroutine and allowing it to suspend itself is easier than doing the coordination outside of the coroutine.

the clock abstraction is very cool and it is fun to treat clocks less as clocks and more as little actors which are sources of timed events.

PR #1097, opened 2020-04-21
refresh clock tempo param with ext input

tehn:
when midi/crow are clock source, update the tempo param with current tempo silently

tested, works for both


artfwo:
@tehn this is generally okay, but is it necessary to change the function based on current source? consider an alternative implementation here:

https://github.com/monome/norns/pull/1100 - this keeps everything in one function

PR #1100, opened 2020-04-21
update clock tempo param with ext input

artfwo:


tehn:
nice, this is cleaner than my solution.

do we want fractional tempos? ie tenths?

PR #1110, opened 2020-04-30
uncrustify -> clang-format

artfwo:


artfwo:
you don't need to do it manually with the formatter :) but unfortunately, function return types get the asterisk aligned to the function_name with this config, so i will see if this can be improved.

PR #1117, opened 2020-05-15
expose configured paths in read only table

ngwese:
- makes `_path` global table read only
- adds public `paths` table which exposes _path and state related paths
- teach tabutil.readonly optional limit which keys are exposed


ngwese:
FYI This change is as much a proposal as it is final. Certainly open to changing names or approach.

tehn:
rad--- looking into it!

PR #1134, opened 2020-06-17
some container objects

ngwese:
these container type objects are used in various places within sky but are independent/generic and are structured like traditional lua module/objects

adds:
- `Deque`
- `Observable` (value)
- `WeakTable` (a table configured with weakly held keys)
- `WatchTable` (notifies observers when a key changes)

this PR also adds a copy of `luaunit` under `norns/lua/lib/test/`


catfact:
awesome.

tehn:
this is super great.

Issue #402, opened 2018-06-06
Maintain folder/cursor position within params:add_file()


adrianfon9:
Similar to App selection Folder/Cursor, params:add_file() should remember place

tehn:
perhaps add_file() could accept a path, and then the menu system could sensibly remember the last path used, and clear this path when a new script is loaded.

tehn:
fixed by #1019

PR #1137, opened 2020-06-19
file params: navigate to last visited folder

csboling:
When editing a file param, navigates to the last folder where a file was selected, to allow quicker changing between sample files in the same folder. Suggested [here](https://llllllll.co/t/norns-sample-load-workflow-improvement/27296).


tehn:
fixes part of #402

(second part of request could be evaluated)

PR #1147, opened 2020-06-29
sky: midi pitch/channel transformers

ngwese:
consolidates midi event transforming devices into a single file and adds correlation ids to note events. correlation ids provide a way to pair related events (like a NOTE_ON and NOTE_OFF) so that the events can still be paired despite being transformed by the `Channel` or `Pitch` devices.

- adds `Channel` device
- renames `Transpose` to `Pitch`, moves from `devices/transpose` to `devices/transform`
- moves `ClockDiv` from `devices/clock_div` to `devices/transform`


ngwese:
added an improvement to the `Switcher` device such that it avoids hung notes if the output selection is modulated.

PR #1149, opened 2020-06-30
add input quantum + wrap to controlspec, add wrap to number param

### LINKED ISSUE #1144: parameter input quantization
currently, `paramset:delta` is hard-coded to increment in steps of 1/100 of the range (`maxval - minval`), times the increment. 

in scripts, its possible to scale the input to the delta arbitrarily, but in menus one is stuck with a quantum of 0.01 times the range (or 0.0005 in "fine" mode.) this is usually fine but sometimes inconvenient.

it might be nice to add an input quantum to `ControlSpec`, as well as an output quantum (the `step` field), so that this can be controlled more specifically.

csboling:
Fixes #1144 

Adds a `quantum` arg to controlspec which is used to define the fraction of the value range that `params:delta` adjusts by, rather than this being fixed at 1% of the range.

It is also sometimes useful to have params with values that wrap around rather than clamping, e.g. angles -- this is the case for the script I'm currently working on which is why I started messing with this. Added a `wrap` option to both `control` and `number` params.

Since this winds up being a lot of args on `ControlSpec.new` now I added a `ControlSpec.def` method which accepts a table -- I looked around a bit but I'm not sure what if any naming convention exists for positional vs keyword-argument style constructors. The table args for `ControlSpec.def` are `min` and `max` for consistency with `params:add`, rather than `minvalue` and `maxvalue` as the positional args are named. Perhaps the positional args should be changed to `min` and `max` for clarity of the luadocs?

Test script for all the new functionality:
```lua
function init()
  params:add{
    type='control',
    id='ctl',
    name='ctl',
    controlspec=controlspec.def{
      min=0,
      max=10,
      warp='lin',
      step=0.01,
      default=0,
      quantum=0.05,
      wrap=true,
    },
  }
  params:add{
    type='number',
    id='num',
    name='num',
    min=0,
    max=3,
    wrap=true,
  }
end

function enc(n, d)
  if n == 2 then
    params:delta('ctl', d)
    redraw()
  end
  if n == 3 then
    params:delta('num', d * 0.1)
    redraw()
  end
end

function redraw()
  screen.clear()
  screen.move(0, 8)
  screen.text('ctl: ' .. params:get('ctl'))
  screen.move(0, 16)
  screen.text('num: ' .. params:get('num'))
  screen.update()
end
```

### DIFF
--- a/lua/core/controlspec.lua
+++ b/lua/core/controlspec.lua
@@ -61,8 +61,10 @@ ControlSpec.__index = ControlSpec
 -- @tparam number step quantization value. output will be rounded to a multiple of step
 -- @tparam number default a default value for when no value has been set
 -- @tparam string units an indicator for the unit of measure the data represents
+-- @tparam number quantum input quantization value. adjustments are made by this fraction of the range
+-- @tparam boolean wrap true to wrap around on overflow rather than clamping
 -- @treturn ControlSpec
-function ControlSpec.new(minval, maxval, warp, step, default, units)
+function ControlSpec.new(minval, maxval, warp, step, default, units, quantum, wrap)
   local s = setmetatable({}, ControlSpec)
   s.minval = minval
   s.maxval = maxval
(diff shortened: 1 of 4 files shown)


csboling:
> renaming minval and maxval to min and max for consistency

Addressed in `287a526` and added default values for these (0 and 1) since the table passed to `ControlSpec.def` might not set them.

PR #1152, opened 2020-07-04
sky: norns key/enc event use n to match arc

ngwese:
previously the "norns" enc events used `num` for the encoder number field while "arc" enc events used `n`. this patch changes the "norns" enc event to use `n` (matching the arc event and bringing it more inline with the native norns api/docs).

fixes a wart noted by @csboling


ngwese:
just realized this also changes `key` events to use `n` which is good since that aligns key events better as well.

PR #1155, opened 2020-07-05
localize smoother type, remove debug print

ngwese:
two minor changes to `filters.smoother`
- makes the `smoother` type/object definition local to the module
- removes debug print in `calc_coeff`

without the type being localized to the module i was running into weird undefined method errors for the `clear(..)` method on the metatable. i'm guessing this might have been caused by globals colliding.

whitespace changes are due to my editor removing trailing whitespace.

### DIFF
--- a/lua/lib/filters.lua
+++ b/lua/lib/filters.lua
@@ -14,8 +14,8 @@ local function wrap_inc(x, max)
 end
 
 --- clear a filter's history
-function f:clear() 
-   for i=1,self.bufsize do 
+function f:clear()
+   for i=1,self.bufsize do
       self.buf[i]=0
    end
 end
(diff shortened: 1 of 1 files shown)


lua/lib/filters.lua
```diff
@@ -81,18 +81,18 @@ setmetatable(median, { __index=f })
 --- @param bufsize: window size, cannot change after creation
 function median.new(bufsize)
    local new = setmetatable({}, median)
-   
+
    new.buf = {}
    if bufsize==nil then bufsize=17 end
    -- only odd buffer sizes are allowed!
    if (bufsize) == 0 then bufsize = bufsize + 1 end
```
catfact, line 88: nothing to do with your PR, but this looks like a typo where bufsize is supposed to be coerced to (2*N)+1, but the modulo comparison is missing

Issue #1151, opened 2020-07-03
non-realtime buffer manipulation commands


catfact:
things like "copy/mix" would be useful additions to buffer "clear, read, write" &c. functionally very straightforward.

additions required for glue layers:

on the crone/c++ side:

(note that for this we _don't_ need to add to `Commands` enum because there is no work to do on the audio thread.)

- [`BufDiskWorker` request/perform methods](https://github.com/monome/norns/blob/main/crone/src/BufDiskWorker.h#L93) (where the work actually happens)
- [`SoftcutClient` method](https://github.com/monome/norns/blob/main/crone/src/SoftcutClient.h#L86)
- [`OscInterface` lambda](https://github.com/monome/norns/blob/main/crone/src/OscInterface.cpp#L677)

and then on the lua/matron/C side:
- [`weaver.c` function](https://github.com/monome/norns/blob/main/matron/src/weaver.c#L178)
- [`softcut.lua` function](https://github.com/monome/norns/blob/main/lua/core/softcut.lua#L296)

tehn:
@catfact did you have additional commands you wanted added? or did @csboling cover this?

catfact:
i can imagine plenty of other things, but this is not a priority. i like it as a "good first issue" to encourage potential contributors to get their feet wet with the audio cpp code in a non-realtime environment. if you're trying to purge the issue list of non-bugs, then i guess close it by all means. i have no personal desires here.

tehn:
i agree it's a good first issue!

i was suggesting that if we were to indicate a few NRT processes we wanted it'd be a better first issue.

ie, NRT gain changes? or normalization?

not trying to simply purge the list of non-bugs... just making an attempt to establish a more focused roadmap

catfact:
gain change and normalization would be obvious ones that have been explicitly mentioned

PR #1161, opened 2020-07-12
softcut: routines for copying and grabbing content from buffer regions

csboling:
Refs #1151 

* Implements a BufDiskWorker routine for copying content from one point in a softcut buffer to another, with optional fade time and an argument to reverse the content while copying. As with `buffer_clear_region`, plumbing includes mono and stereo variants.
* Implements a "render" routine for capturing the contents of an arbitrary portion of a softcut buffer, sampled at a caller-specified number of points, using the procedure described in [this](https://llllllll.co/t/norns-2-0-softcut/20550/186) post. If the total number of frames in the requested region is less than the requested number of samples, frames are returned unaltered. Otherwise a simple peak detection scheme is used, skipping frames if the requested region is sufficiently large.

For debugging and demonstration I wrote [this](https://github.com/csboling/waveviz) simple waveform visualizer script, presented with apologies for possible line ending issues and definite script-level bugs.

An assorted mash of remarks / questions / insecurities:

* Nomenclature and method signatures - not sure about the choice of OSC paths in particular. 
* I think the algorithm(s) are right? I can hear fades working and stuff looks reasonable in the visualizer. Not sure how best to more thoroughly verify this kind of thing, I guess probably I should export WAVs and pull them up in Audacity side-by-side or something. Most likely needs some more thorough testing for edge cases.
* Peak-finding downsampling is probably too aggressive, needs tuning for accuracy / perf tradeoff. I just cooked up a value that seemed to work OK to avoid xruns for large regions.


crone/src/OscInterface.cpp
```diff
@@ -681,6 +681,65 @@ void OscInterface::addServerMethods() {
...
+            return;
+        }
+
+        int ch = argv[0]->i;
+        if (argc > 3) {
+            sampleCt = argv[3]->i;
+        }
+
+        softCutClient->renderSamples(ch, argv[1]->f, argv[2]->f, sampleCt,
+                                     [=](float secPerSample, float start, size_t count, float* samples) {
```
csboling, line 737: Here and [here](https://github.com/monome/norns/pull/1161/files#diff-78e28548e3d76018659322684a08402aR28) I demonstrate that I only pretend to know how to do Modern C++ things. All the other callbacks use regular function pointers and I understand that `std::function` is more costly, but I could not get this to compile with a function pointer when this lambda needs to value-capture the `ch` index. Other places in this file use `addServerMethod` with some capturing going on (I think?) but the compiler is not happy about this -- maybe because this is a lambda defined inside another lambda?

tehn:
holy smokes!! this is huge--- looking forward to digging in--- i was just thinking the other day about buffer ops, and waveforms have been in the back of my head since the projects' inception...

crone/src/OscInterface.cpp
catfact: i think the overhead of the lambda is totally fine here. i don't think there are any cases of capturing elsewhere - there is just a lot of static storage.

csboling:
* Changed fade curve to raised-cosine.
* Realized I was only fading in the beginning of the copied content and not fading out its end 😬 
* For doubling an existing softcut loop exactly I realized that you probably want to just mix the existing content in the buffer with the copied content and not do an additional crossfade since the loop content will already have its endpoints faded -- you just want to sum the faded ends. Therefore added a "preserve" option to keep / attenuate the existing content you overwrite with the copy, rather than eliminating it entirely, which should allow you to use something like (I think) 
  ```lua
  softcut.buffer_copy_stereo(0,             -- start of copy source
                             length,        -- start of copy destination
                             length + 0.1,  -- source length including post-roll, so s/0.1/fade_time/
                             0.0,           -- no crossfading during copy
                             1.0)           -- mix existing material 1-to-1 with copied material
    ```
  I tried editing Samsara along these lines and it seems anecdotally to work, but I am still getting a pop occasionally when doubling the buffer while recording. Need some more time digging to know why but I thought I would post these updates in case someone else gets a chance to try this patch.
* For my own application I think I also want to add a fade time option to clearBuffer to allow emptying a buffer region without introducing discontinuities. This would involve adding an argument to the OSC methods, not sure if that kind of thing constitutes a problematic API change. May be best to make a follow-up PR for this.

crone/src/BufDiskWorker.cpp
```diff
@@ -128,6 +147,14 @@ int BufDiskWorker::secToFrame(float seconds) {
     return static_cast<int>(seconds * (float) sampleRate);
 }
 
+float BufDiskWorker::raisedCosFade(float unitphase) {
+    return 0.5f * (cosf(M_PI * (1.f + unitphase)) + 1.f);
+}
+
+float BufDiskWorker::mixFade(float x, float y, float a, float b) {
+    return a * x + b * y;
```
csboling, line 155: Split this out to a function so the mixing approach could be tweaked. Not sure if this should use the same mixing calculation as [ReadWriteHead::mixFade](https://github.com/monome/softcut-lib/blob/4b535d9823e59e62437a56fbef6ce382bbcf050f/softcut-lib/src/ReadWriteHead.cpp#L178), I don't think I understand how this works so I left it linear here. Also maybe need to think about clipping behavior?

tehn:
>     * For my own application I think I also want to add a fade time option to clearBuffer to allow emptying a buffer region without introducing discontinuities. This would involve adding an argument to the OSC methods, not sure if that kind of thing constitutes a problematic API change. May be best to make a follow-up PR for this.

this could be pulled off without a breaking change on the lua API side because a default arg could be assumed. i think we can change the OSC layer presently without treating it like an API layer--- this of course could change in the future if there was an alternative front-end to crone (instead of matron) but i don't think that's going to happen--- if anything there's been talk of changing the IPC from OSC to something else, though that's not super high priority.

but yes, clear with fades with be a great feature. i'm very enthusiastic about these buffer ops! hoping to test this week, thank you for your efforts!

catfact:
>  fade time option to clearBuffer

re: OSC/lua API for this, i agree:

- lua API can use extra arg with default value, non-breaking
- in OSC layer, i haven't figured out a way to use optional arguments. `clearBuffer` could take a new arg (breaking) or make a new `clearBufferWithFade` (non-breaking.) AFAIK i'm the only one using `crone` without `matron`, so either way is fine. the former is more consistent with the style of existing commands, the latter probably would have been smarter in the first place. (oh well.)

crone/src/BufDiskWorker.cpp
catfact: i actually think it's fine to exceed unity range in the buffer.

crossfade options are a whole huge topic that i'd love to discuss...

- for the (generally good[*1]) assumption that real-world signals are uncorrelated, we want the sum of each signal's power (squared amplitude) to remain constant.
- $sin^2 + cos^2 = 1$ is a convenient relation for building curves with that property.
- in crone/softcut we've used both "raised cosine" (for a gradual fade) and "half cosine" for a fast fade.
- tbh, neither is really ideal and don't create "perceptual linearity." [*2]


so anyways maybe we can raise a TODO issue for updating the crossfades perfectly.


[*1] when signals are perfectly correlated, you want "equal voltage" instead of "equal power." it's prohibitive to be determining this in real time for all mix points of course; but you could actually do it for an offline operation.
[*2] it's even weirder in context of softcut xfades, because overdub and record levels are also crossfading, so correlation with signals from the past comes into play.

tehn:
excellent.

i'll do more testing tonight with live input, and attempt some high-stress (queue up a ton of buf workers) but i suspect this is probably good to go. there will be caveats as to how this functionality can be used, but i think that's reasonable.

tehn:
ok, i'm feeling good about this and feel it can be merged to main. that way we might get a few additional testers prior to a release. @csboling @catfact any concerns?

csboling:
Testing last night I finally got it through my head that you get weird behavior if the source region and target region overlap, so it needs to use the `memmove` trick (changing which direction the copy loop goes depending on the relative positions of the source / target). Going to work on this this morning and then I think it's good to go.

tehn:
good catch. i'll try to analyze similar edge cases

csboling:
- Fixed handling of overlapping regions in the non-reversed copy case. This allows for loop-doubling without discontinuities, I tried this out with Samsara and it (finally) works as expected 🔁🔁 
- When reversing content, there's no way to avoid garbling the overlap without allocating additional storage the size of the overlap and doing another copy (i.e. potentially the size of the whole softcut buffer(s)) so this is left unhandled. Added a remark in the Luadoc that this is not handled when `reverse` is set.
- Use the complementary-sines mixing approach for crossfades for consistency with `ReadWriteHead`. Avoid the trig overhead when not crossfading and just use a linear mix since it's mathematically equivalent.

PR #1162, opened 2020-07-15
Fixed: fine param value control after entering file select

tomwaters:
Hope this is OK, I'm just getting familiar with lua and the norns source.

I noticed that if a script has file and number params, when you select a file parameter and come back out of the file selector the number params are much harder/slower to change. Fine value control is enabled when you go into the file selector but it isn't disabled when you come out.

This change disables fine control when coming out of the file selector.

### DIFF
--- a/lua/core/menu/params.lua
+++ b/lua/core/menu/params.lua
@@ -198,8 +198,9 @@ m.key = function(n,z)
         m.mpos = 1
         m.mode = mMAPEDIT
         m.pm = pm
+      else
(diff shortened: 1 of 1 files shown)


tomwaters:
Having slept on it, i'm not sure this is the right fix. The problem isn't that we should be clearing fine coming out of the file selector, but that we're setting fine too broadly.

Should fine only be set on key 3 down if mode is mMAP?

PR #1167, opened 2020-07-21
add params/toggle

### LINKED ISSUE #1166: add params/toggle
proposing a new param type exactly like `trigger`, but stores a boolean and toggles on key press in the menu

certainly `option` can accomplish a similar thing - my particular use case is working on a screen UI builder tool and wanting to let the user input params - in this case I'd need to draw `option` different from `toggle` so there'd be an awkward workaround to detect  which case is which (& maybe this helps clear the way for midi button / foot switch mapping in the menu!)

andr-ew:
Fixes #1166 

I think I got it...?

here's my test script:

```
function init()
  params:add_toggle("test", "test", true)
  params:set_action("test", function(v) print(v) end)
end
```


lua/core/menu/params.lua
```diff
@@ -198,9 +203,8 @@ m.key = function(n,z)
         m.mpos = 1
         m.mode = mMAPEDIT
         m.pm = pm
-      else
-        m.fine = true
       end
+      m.fine = true
     elseif n==3 and z==0 then
```
catfact, line 208: this seems unintentional (we don't want to enter fine mode for all cases)

catfact:
check L207, otherwise looks ok to me :eyeglasses: 

i'm not convinced `toggle` needs a `formatter`, but i guess it doesn't hurt anything.

if i were to add any additional options, it would probably be a "gate mode" flag which changes the delta behavior, such that (d>0) -> true, (d < 0) -> false, (d == 0) -> no change. (this is like an `option` or `number`, but having both behaviors in one object could be more convenient for customization, especially if part of the point is having a custom formatter for the true/false values.)

andr-ew:
I think 207 was just a vim goof

(I honestly wasn't thinking too much about the formatter thing when I copied it over, oh well)

lua/core/menu/params.lua
tehn: line 201-202 are supposed to stay, but perhaps they should be made more explicit (this case should happen on number and control types, but we're just doing a catch-remaining else)

lines 204-205 (key 3 up) turn off fine

tehn: but correct, line 207 was wrong

tehn:
besides breaking the fine tune (easy fix) this looks good, thanks!

andr-ew:
does this look right ? truly I do not know how I managed to bork those lines as it wasn't even close to where I was editing

andr-ew:
just tested fine tune and it seems good, lol

tehn:
ugh, `formatter` is not a tparam. just param. this busts the CI integration (luadoc generation, see below)

honestly i'd just rip out all of the formatter stuff?

andr-ew:
ah gotcha, easy to do !

PR #1169, opened 2020-07-24
give UI.Slider direction

ericmoderbacher:
@markwheeler What do you think about this feature?  If there is something you want slightly different just let me know I can work on it.

### DIFF
--- a/lua/lib/ui.lua
+++ b/lua/lib/ui.lua
@@ -314,8 +314,9 @@ UI.Slider.__index = UI.Slider
 -- @tparam number min_value Minimum value, defaults to 0.
 -- @tparam number max_value Maximum value, defaults to 1.
 -- @tparam table markers Array of marker positions.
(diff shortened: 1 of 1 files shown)


markwheeler:
Awesome! Something I'd been meaning to add.

One thought/question: did you consider having an `orientation` param that takes `horizontal` or `vertical` given that the 'direction' can already be flipped using minval/maxval? Might simplify the implementation, unsure if it's clearer to users?

ericmoderbacher:
I did consider that but swapping the min and max val gets a little odd when we call linlin because it doesnt like when f is larger than the max.
https://github.com/monome/norns/blob/7e1cb1aa379d6eae53953f3701cf59b5bd217c78/lua/lib/util.lua#L132

I thought direction made more sense, but sometimes I think about things weird.

markwheeler:
Ah cool, sounds good, Thanks for checking!

PR #1170, opened 2020-07-27
fixed clock documentation - stop should be cancel

tomwaters:
Documentation for the clock module lists a 'stop' function but it should be 'cancel'


tehn:
hi @tomwaters ! thanks for the PR. the docs are auto-generated by luadoc, so the change actually has to happen in `lua/lib/clock.lua` instead

there's a chance this is already fixed, but the luadocs need to be regenerated.

PR #1175, opened 2020-08-02
[WIP] Include boost

catfact:
collected boost headers needed for `boost::lockfree`, removed trivial boost dependencies (here and in submodules)

should make it unnecessary to ship boost on norns for development

compiles, runs, but needs testing


catfact:
... i decided it would be better to use a totally different lockfree SPSC implementation and remove boost altogether...

so... will keep this open and add to it

catfact:
closing this in favor of removing boost and using a different lockfree queue implementation

Issue #1182, opened 2020-08-21
add MAP support for trigger, toggle, momentary  


andr-ew:
the intention here is to support midi foot-switch devices or midi buttons in PARAMS/MAP. the primary thing for me would be supporting cc messages (usually 0 for off, 127 for on, to my knowledge) but I suppose midi keys could be supported someday too.

the MAP submenu would only need to show `cc`, `ch`, `dev` (no `in`, `out`, `accum`)

again, happy to PR though I know this one is less trivial maybe

tehn:
i'm not seeing a clear path on this one, can you try to elaborate?

for example, mapping a TRIG, would basically any incoming cc value call the action?

for toggle, since it's basically a TRIG with internal state, would each subsequent cc input simply flip the state and call action?

as i said in just-closed issue, you get MOMENTARY for free with OPTION already, with cc mapping.

also, do you have specific gear that has foot switches/etc? i'm wondering about standardization of cc/note behavior, and program-ability of the source device.

andr-ew:
the specific implementation I'm imagining is essentially just mirroring the state of the norns key with the state of the button :

for `TRIG`, if cc > 63 call `action` 
for `TOGGLE`, if cc > 63  flip state 
for `MOMENTARY`, if cc > 63 set state = 1, else set state = 0

re: hardware I'm building a teensy thing soon but I just researched some commercial options (KMI softstep, morningstar MC6, novation launch control, akai mpk mini) and it seems like they generally ship with a config tool and offer the gated cc option and toggles (sometimes you can set the cc number for on/off, sometimes not (?)). certainly there will be weird inconsistencies as there's not really a standard, but the 63 boundary thing is used by ableton & seems like a good place to start.

catfact:
i'm sorry but i'm also not seeing the benefit of these specializations compared to `control` and so on. i don't think the selection of parameter type is the place to define behavior and add state.

andr-ew:
I am really truly wishing to develop these features for a new implementation of [wrms](https://github.com/andr-ew/wrms/blob/master/wrms.lua) that uses the params in a particular way (to store all states and configure the UI library) - that might clear it up ?

catfact:
i mean can you post a small piece of code

catfact:
@andr-ew i'll take a look right now at the behaviors currently accessible and see if we can't clarify that way.

[...]

PR #1185, opened 2020-08-28
replace params/toggle with params/binary

andr-ew:
#1182 

replaces `toggle` with @catfact's suggested more capable `binary` type (value is `0` or `1`), with three `behavior` strings (`trigger`, `momentary`, `toggle`) affecting the behavior of the param within the `EDIT` menu, and later when some midi mapping support is added

just wanted to try this out real quick and get feedback. to me this feels like a clean way to solve the switch mapping issue from a UI angle - it feels clear how mapping the param to a midi switch would behave in the different cases


andr-ew:
here's my test script

https://gist.github.com/andr-ew/9183b0fb55dd58ffaee9235e98b03ca6

tehn:
apologies for the long delay. getting ready for an update so i'll test this for inclusion.

tehn (approved):

PR #1186, opened 2020-08-28
Remove boost

catfact:
scrub boost dependencies from softcut and crone.

the only significant functionality we used from boost was `lockfree::spsc_queue`. this has been replaced with moodycamel's `ReadWriteQueue`, a much lighter dependency and possibly more performant.


catfact:
only to make room for more powerful and insidious new evils

PR #1190, opened 2020-09-04
ER: return empty table when zero pulses are requested

### LINKED ISSUE #1189: er.gen(0, 16) == er.gen(1, 16)
- expected behavior: providing er with a value of `0` for pulses returns a table where all booleans are false.
- actual behavior: providing er with a value of `0` for pulses returns a table where the first boolean is true.

if this is actually an issue and not intended behavior i can make a pr to resolve it.

catfact:
fixes #1189


catfact:
ech sorry,

PR #1191, opened 2020-09-04
ER module: correctly handle request for empty table; add shift param

### LINKED ISSUE #1189: er.gen(0, 16) == er.gen(1, 16)

catfact:
fix #1189

### DIFF
--- a/lua/lib/er.lua
+++ b/lua/lib/er.lua
@@ -11,6 +11,8 @@ function er.gen(k, n)
    local r = {}
    for i=1,n do r[i] = false end
 
+   if k<1 then return r end
+
    -- using the "bucket method"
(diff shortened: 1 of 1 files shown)


catfact:
just to make this a little more interesting, also added a shift parameter to ER.gen()

tyleretters:
@catfact amazing. i just built this in arcologies. definitely will be used. https://github.com/tyleretters/arcologies/blob/main/lib/Cell.lua#L138

tehn (approved):

PR #1194, opened 2020-09-09
musicutil scales update

markwheeler:
Combine a few identical scales, remove some problematic naming that @tyleretters flagged.

**Note:** This will break saved presets as the scales index has changed.

I wonder if we might want to revisit this scale data in a larger way at some point, it's certainly very western-perspective.


tyleretters:
1. would it be helpful to stub in duplicate scales at those index positions so as not to break old presets?
2. i'd be very interested in helping revisiting scale data to be more inclusive

markwheeler:
Re: 1. I think it'd be difficult to do this without causing duplicates to show up in the params menu (at least my own scripts I've used the lib in this way). Unless I'm missing a way to do it cleanly?

Re: 2. That'd be awesome, this honestly isn't a topic I'm knowledgable about (I made the lib to avoid thinking about it in every script 😉)

tyleretters:
1 - yeah good point. there'd be ways to do it but it probably isn't worth the effort for value.

2 - same. i just watched this last night so i'm pretty into this topic right now: https://www.youtube.com/watch?v=Kr3quGh7pJA

tehn (approved):

PR #1203, opened 2020-09-28
sky: unstable device for polyperc engine

ngwese:
nothing major, just some initial glue to allow use of `PolyPerc` in sky device chains similar to the existing `PolySub` glue.


ngwese:
I'm going to go ahead and merge this as it will not affect anyone other than myself and is it going into the unstable sky device directory.

PR #1206, opened 2020-10-14
new midi input handler and virtual rawmidi port support

ngwese:
this PR is intended to be an alternative to https://github.com/monome/norns/pull/1204

this PR is different in two aspects:
- it does not depend on third party code from fluidsynth, it was implemented by referencing the existing midi input handler and the published midi 1.0 specification alone
- it takes a different approach to integrating a virtual rawmidi device into the matron device class. 

virtual rawmidi devices do not have a device file unlike other physical devices. the initial implementation used a faux device file (`/dev/null`) and added a `virtual` boolean to all device types (which was ignored by all device types except midi) in order to handle the special case during midi device initialization. this PR instead adds a new device type which re-uses most of the existing midi device implementation and allows devices to be created with NULL as the device file path.

specific details to note:
- `dev_midi_start` (and related functions) is less minimal than before. reading from a buffer and supporting running status required enough logic to keep track of midi message boundaries and to differentiate between different status bytes (system real time messages do not affect running status logic unlike other message types) that one has ~50% of a midi parser now implemented in C.
- incoming midi sysex data is segmented (matron) midi event values and passed through to the lua layer. this is of limited initial use but opens the door to responding to midi device which send sysex messages when certain physical controls are manipulated (DSI Evolver desktop was used during testing).

the virtual device support was (lightly) validated by setting up `rtpmidi` and sending midi data over wifi from AUM on an ipad as well as sending midi data over ethernet from ableton live (running on macos 10.14).

 cc/ @patriciogonzalezvivo


tehn:
code looks great. despite tracing initialization i remain slightly confused if/what needs to be done on the linux side?

ngwese:
> code looks great. despite tracing initialization i remain slightly confused if/what needs to be done on the linux side?

...regarding the usage of the virtual midi port?

apologies for not more clearly explaining the scope of the change. this change just adds an ever present `virtual` midi device to matron. it does nothing to make that port usable in a meaningful way without a user shelling into the device and connecting that virtual device to some other program running on the device sends or receives midi via `aconnect`.

the original motivation for the virtual device was to combine it with the `rtpmidi` daemon to allow midi data to be received over the network. in the prior PR concern was expressed about ensuring `rtpmidi` was accessible to a user of the device without the need to tweak things at the linux level. in my testing i found `rtpmidi` to be a bit flaky so i stopped short of working to integrate that as well. this change is the bare minimum to open the door to further experimentation.

a significant amount of time was spent trying to figure out if it was possible to set descriptive names for the virtual device which gets constructed at the alsa level. in the end it took reading of the alsa code to discover that the static names given to the virtual device are hard coded into alsa unfortunately.

tehn:
thank you! my brain is catching up with the previous ideas, but this is an excellent summary.

simonvanderveldt:
>  this change just adds an ever present `virtual` midi device to matron

Won't this be confusing to users, also keeping in mind most of them probably won't be using this functionality?

ngwese:
without a dash of documentation i could see some initial moments of confusion.

one thought that just occurred to me is that i think there is some logic which sets the initial `vport` assignment if nothing has been configured. i think it would be good to investigate excluding the virtual interface from the candidate list so that it doesn’t end up implicitly selected (only explicitly by user action)

ngwese:
@simonvanderveldt - i've added an additional change to `midi.lua` which avoids auto-assigning the virtual port thus preserving the current behavior or auto-assigning physical vports. the end result is a much cleaner user experience so thank you for the extra push.

Issue #1200, opened 2020-09-22
FR: add a "favorite" mechanism to the SELECT menu


dndrks:
a recent support ticket surfaced a nice feature request for on-norns management of which scripts the SELECT menu presents to the user.

**current behavior**: all scripts, regardless of direct need (eg. I want to use a script which requires the Molly engine from the Molly project but I won't be using the Molly _script_), surface in the SELECT menu.

**desired behavior**: user wants to manage, from the norns hardware alone, which scripts the SELECT menu presents.

----

<s>I put in some thought on this today and realized that currently, holding key 1 in the SELECT menu doesn't execute anything (whereas a press toggles back to the currently running script). so, maybe we can endow a key 1 hold in SELECT with special powers?</s>

<s>**proposed behavior**:
- user navigates to SELECT and highlights an entry they want to hide
- user holds key 1 and presses key 3 to hide the entry from SELECT
     - this two-key gesture adds the filepath for the entry to an ignore.txt
- when [building the visible list](https://github.com/monome/norns/blob/main/lua/core/menu/select.lua#L7-L29), omit any paths from the ignore.txt file
- when the user holds key 1, the hidden entries are shown again but dim (screen level 2, maybe?)
- while holding key 1, the user can navigate to a hidden entry with the encoders
     - if they press key 3 while holding key 1, the entry is removed from ignore.txt and will display in the SELECT scroll again
     - if they release key 1 while on a hidden entry, they are re-focused onto the next visible entry</s>

**proposed behavior** (thank you @pq + @tyleretters for refinement!)
- SELECT has two sections: `favorites` + `all`
- e3 CW on an `all` entry defines it as a `favorite` (adds a * to the right in the `all` view)
- `favorites` are surfaced at the top of the SELECT scroll (by adding to / querying a favorites.txt file?)
- as the user scrolls down past `favorites`, they hit a line after which `all` entries are shown (here, `favorites` have a *)
- e3 CCW on a `favorite` in the `all` section will remove the * and pull it out of the `favorites` section of the SELECT scroll (user cannot accidentally remove a favorite in the `favorites` section

pq:
yay!  some script management would be awesome.  for conversation, i wonder about an inverse affordance... a while ago we briefly discussed "favorites" (https://github.com/monome/norns/issues/987) but maybe this is even better?

dndrks:
oh! @pq , i really like framing this as a positive endowment, rather than a negative.

ok, so what if:
[...]

PR #1207, opened 2020-10-14
First pass at adding 'favorites'.

zzsnzmn:
Related to #1200 I would like to add a separate selection at the top right now, but this will at least add an asterisk next to any script selected as a favorite.

### DIFF
--- a/lua/core/menu/select.lua
+++ b/lua/core/menu/select.lua
@@ -4,7 +4,6 @@ local m = {
   pos = 0,
   list = {},
   favorites = {},
-  selected_script = "",
   len = "scan"
 }
 
@@ -109,7 +108,6 @@ m.redraw = function()
         local line = m.list[i+m.pos-2].name
         if(i==3) then
           screen.level(15)
-          m.selected_script = m.list[i+m.pos-2]
         else
           screen.level(4)
         end
@@ -123,15 +121,15 @@ m.redraw = function()
 end
 
 m.add_favorite = function()
-  if not contains(m.favorites, m.selected_script) then
-    table.insert(m.favorites, m.selected_script)
+  if not contains(m.favorites, m.list[m.pos+1]) then
+    table.insert(m.favorites, m.list[m.pos+1])
     tabutil.save(m.favorites, "/home/we/dust/.favorites")
   end
 end
 
 m.remove_favorite = function()
   for i, v in pairs(m.favorites) do
-    if v.file == m.selected_script.file then
+    if v.file == m.list[m.pos+1].file then
       table.remove(m.favorites, i)
       tabutil.save(m.favorites, "/home/we/dust/.favorites")
       return
--- a/lua/core/menu/select.lua
+++ b/lua/core/menu/select.lua
@@ -17,6 +17,13 @@ local function menu_table_entry(file)
 end
 
 local function sort_select_tree(results)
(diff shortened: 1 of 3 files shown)


lua/core/menu/select.lua
```diff
@@ -70,16 +111,31 @@ m.redraw = function()
...
       end
     end
   end
   screen.update()
 end
 
+local function noop(_)
+end
+
+m.add_favorite = function()
```
zzsnzmn, line 132: both of these are kinda mega hacks -- if it'd be preferable to do this is a lua native way i'm totally happy to go that route.

ngwese: I haven’t looked at the code in detail but at first glance I’d consider maintaining favorites in a table and just using `tab.save` and `tab.load` functions in the `tabutil` module.

https://github.com/monome/norns/blob/main/lua/lib/tabutil.lua#L93

zzsnzmn: ah yeah, that's way better. i'll get that pushed up in a few

lua/core/menu/select.lua
```diff
@@ -1,37 +1,65 @@
+tabutil = require "tabutil"
+
 local m = {
   pos = 0,
   list = {},
-  len = "scan" 
+  favorites = {},
+  selected_script = "",
```
zzsnzmn, line 7: I'm fine with tossing this -- it just felt a little weird calling `m.list[i+m.pos-2]` all over the place.

zzsnzmn:
okay so i think this should be good to go. lmk if there are any suggestions about how it works. i'm happy to provide screenshots if necessary.

ngwese:
Cool - I’ll give this a spin on my device!

lua/core/menu/select.lua
```diff
@@ -1,37 +1,64 @@
...
+  end
+  return false
+end
 
 m.init = function()
-  m.len = "scan" 
+  m.len = "scan"
   m.list = {}
+  m.favorites = {}
+  m.favorites = tabutil.load("/home/we/dust/.favorites")
```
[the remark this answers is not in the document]
ngwese: UPDATE: after some thought is would be valuable to call this file `system.favorites` so that it is visible in maiden and sits nicely along side the existing `system.state` and `system.pset` files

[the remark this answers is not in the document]
tehn: agreed with @ngwese it'd be great to have this file be visible

otherwise, this is great!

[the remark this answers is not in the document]
zzsnzmn: That all makes sense to me, I'll try and get a patch submitted this evening or tomorrow morning.

zzsnzmn:
Just factored the paths into the paths table, also added its removal to the 'reset' menu option.

dndrks:
oh, excellent! will bring onto my test machine tomorrow!! 💞💞💞

dndrks:
wowowow, @zzsnzmn , this is fantastic! really elegant, lovely implementation that feels minimal and almost background, but is super powerful. thank you!!

tehn (approved):

tehn:
this works/looks great! @ngwese yeah?

ngwese (approved):

ngwese:
Yup - Looks good. Thanks @zzsnzmn! Very excited to use this feature.

Issue #1209, opened 2020-10-20
negative resonance settings crash softcut


lylepmills:
I ran the following on a whim to see what would happen -

```
softcut.post_filter_rq(1, -0.01)
```

In doing so I managed to crash softcut badly enough that (after some brief, very loud, self-oscillation) it didn't make any more sound. Nothing short of a hard reset by pressing the button on the back of the norns was enough to get it back to a good state (I tried things like `softcut.reset()`, soft resets, etc).

It seems like there should be some protection against this?

tehn:
fixed by #1211

PR #1211, opened 2020-10-21
Clamp softcut parameters

catfact:
alternate fix for #1209 

this is completely untested! (it compiles.)


tehn:
rad! will test on device

tehn (approved):

PR #1213, opened 2020-10-23
clamp more parameters, tidy and refactor

catfact:
clamp more softcut parameters, including Rate and voice/buffer indices. (in theory, i think it was possible to crash or smash memory by sending OOB indices.)

also some small refactoring for DRY and to avoid magic numbers.

like the last PR, this is not really tested except for compilation.

### DIFF
--- a/crone/src/SoftcutClient.cpp
+++ b/crone/src/SoftcutClient.cpp
@@ -9,19 +9,19 @@
 #include "SoftcutClient.h"
 
 // clamp to upper bound (unsigned int)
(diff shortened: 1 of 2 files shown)


tehn (approved):
ah, this is a nice update, thank you. code looks great. i'll test tomorrow and slip it into the update.

catfact:
woops, big error, wait 1...

catfact:
...there we go.

tehn:
tested and merged as #1214

Issue #955, opened 2019-12-01
Differentiating ttyACM devices


okyeron:
I started down this roard with issue #525 and it's [come back around with crow](https://github.com/monome/norns/pull/839#issuecomment-501969281), so I'm starting a new issue to find a good way forward for future ttyACM devices.

**Problem:** 
DIY devices using Teensy or some other microcontrollers use `/dev/ttyACM*` 
but [device_monitor.c] (https://github.com/monome/norns/blob/master/matron/src/device/device_monitor.c) is currently hard coded to treat any ttyACM device as a crow. There is a handshake for crow, but it just [ignores anything not a crow](https://github.com/monome/norns/blob/ef4900316d9257267d3f16984be777aebf0c2dd7/matron/src/device/device_crow.c#L61) (and it really seems like we need to differentiate between devices *before* it gets to the handshake stage).

I've done some hacking on this, but I could really use some guidance.

Possible Solution:
Simplify the watchers by subsystem - to only "tty", "sound", "input". For tty, check the node and then add additional checks for ttyACM devices. 

In [handle_device()](https://github.com/monome/norns/blob/ef4900316d9257267d3f16984be777aebf0c2dd7/matron/src/device/device_monitor.c#L204) we could look at device information from udev and get product ID, model, vendor, etc. (which is similar to what libmonome does I think). Would those details be adequate for differentiating devices?

```
udev_device_get_property_value(dev, "ID_MODEL"); 
udev_device_get_property_value(dev, "ID_VENDOR");
udev_device_get_property_value(dev, "ID_SERIAL_SHORT");
```
`udevadm info --query=all --name=/dev/ttyACM0` is very handy to check those names/values

At this point, i think the device could be added in [device_list.c](https://github.com/monome/norns/blob/master/matron/src/device/device_list.c) as it stands (or cases added for new device types).

Thoughts? Suggestions? Help?

I don't have a crow, so i can't test against that.

catfact:
ok cool, then seems to me like vid/pid is a possibly simple way to identify crow without handshake+fallback logic.

druid does this:
https://github.com/monome/druid/blob/master/src/druid/crowlib.py#L14

that said, i would not mind structuring the device detection stuff more intelligently anyway.

simonvanderveldt:
Model and vendor are what's normally used for determining if a device is what you think it is, so unless there are any blockers that would prevent us from using or setting)them it should be the correct way to achieve this.

tehn:
[...]

PR #1219, opened 2020-10-28
avoid monitoring device types without device files

### LINKED ISSUE #1217: 201023: crow showing as a MIDI device, not usable
reported at https://llllllll.co/t/norns-update-201023/37646/25.

to repro:
- plug a crow into norns while monitoring in maiden
- crow does not trigger the typical "telephone line" message in maiden
- crow shows up as a MIDI device under DEVICES > MIDI: `crow: telephone line`
- subsequent plug/unplug shows as `crow: telephone line 2`, `crow: telephone line 3`, etc
- crow is not addressable by norns via CLOCK > crow output

@ngwese , unsure if this is related to the virtual MIDI work?

ngwese:
fixes a bug in cac8479fcc where crow devices were
mapped to the virtual midi device type (which has no
corresponding device file to watch for) because the
device pattern watch list did not match the order of
the `device_t` enum values.

fixes #1217

### DIFF
--- a/matron/src/device/device_common.h
+++ b/matron/src/device/device_common.h
@@ -9,13 +9,17 @@ typedef enum {
     DEV_TYPE_HID = 1,
     // raw midi devices
     DEV_TYPE_MIDI = 2,
-    DEV_TYPE_MIDI_VIRTUAL = 3,
     // usbmodem (crow)
-    DEV_TYPE_CROW = 4,
+    DEV_TYPE_CROW = 3,
+    // place all virtual devices (devices without device files to monitor) below
+    DEV_TYPE_MIDI_VIRTUAL = 4,
     // counter - unused, don't remove
     DEV_TYPE_COUNT,
     DEV_TYPE_INVALID
 } device_t;
+
+// maximum device_t value for devices which have corresponding device files
+#define DEV_TYPE_COUNT_PHYSICAL 4
 
 struct dev_common {
     // device type
--- a/matron/src/device/device_monitor.c
+++ b/matron/src/device/device_monitor.c
@@ -41,13 +41,14 @@ struct watch {
 
 // watchers
 // FIXME: these names / paths are really arbitrary.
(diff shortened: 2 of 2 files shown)


matron/src/device/device_common.h
```diff
@@ -9,14 +9,18 @@ typedef enum {
...
+    DEV_TYPE_CROW = 3,
+    // place all virtual devices (devices without device files to monitor) below
+    DEV_TYPE_MIDI_VIRTUAL = 4,
     // counter - unused, don't remove
     DEV_TYPE_COUNT,
     DEV_TYPE_INVALID
 } device_t;
 
+// maximum device_t value for devices which have corresponding device files
+#define DEV_TYPE_COUNT_PHYSICAL 4
```
artfwo, line 22: i think DEV_TYPE_COUNT_PHYSICAL could be also declared in the device type enum.

artfwo (approved):

matron/src/device/device_common.h
ngwese: I wasn't sure of that; I thought if I did it would throw off the implicit count which `DEV_TYPE_COUNT` is set to.

ngwese: ...to be clear i'm not loving this fix, it is more fragile than i'd like but it seemed the least bad of the options i tried.

the thing which broke when i changed the enum around was related to this static array:

okyeron: FWIW - I worked around that static array on Line 44 in my hacks to device_monitor for teensy (ACM) devices - [mentioned in this issue](https://github.com/monome/norns/issues/955#issuecomment-561963162)

I will test tomorrow if this works with the virtual midi devices or not. 

Don't know if this helps or not, but thought I'd mention it again.

ngwese:
i'm going to roll with this change for now so that we can get a fix out soon. @okyeron - it would be great to hear if teensy base midi devices behave with this change.

Issue #1218, opened 2020-10-27
shield-specific behaviors


tehn:
system awareness of shield vs. not-shield

for shield:
- remove i2c headphone errors
- "sleep" needs to be clarified because the device doesn't shut down
- remove battery displays/etc
- anything with wifi??

catfact:
how would we actually detect shields?

it's the same kernel and everything, right?

can we see something different in `/proc/cpuinfo` or etc?

ericmoderbacher:
Maybe this thread will help:
https://www.raspberrypi.org/forums/viewtopic.php?t=192748

Looks like you can get a cpu revision.

catfact:
thanks!

capturing that: seems like both these paths should have something usable
`/proc/cpuinfo`, indeed (maybe messier)
`/sys/firmware/devicetree/base/model` (human-readable string built by rpi bootcode)

tehn:
sorry I should've mentioned this is already figured out because we have to id the chip for updates. we do it in update.sh which is in the update folder.

other issues however remain, will post them here when i get a minute

catfact:
gotcha. so, here:
https://github.com/monome/norns/blob/main/update/update.sh#L56-L62

should also, i dunno, write a config file that matron can use.

csboling:
I think it would probably work to configure all these things in the [matronrc.lua](https://github.com/csboling/norns/blob/decouple-fb/matronrc.lua) file my screen-decoupling branch (for #1187) loads at startup to decide which type of screen / input source to use.

tehn:
using `matronrc.lua` is a good idea--- but i'm wondering where/how the config step should happen. at some point matron will have to self-detect if it's on a CM3 or a shield, and that info can be written to the rc file... so should there always be a user config file?

for example, if the user config file doesn't exist, first copy the default, then append some extra information re: detecting architecture? i suppose since desktop is possibly an option it should check that as well?

tehn:
fixed by #1250

Issue #1187, opened 2020-08-29
support multiple screen backends


csboling:
To facilitate running norns on a desktop or other hardware, it is desirable to have matron be able to render to different types of displays besides the linux `/dev/fbX`. This issue is intended for discussing the design of this functionality and implementation approaches. My prototype branch is [here](https://github.com/monome/norns/compare/main...csboling:decouple-fb).

Use cases include:
- working on matron and lua without norns hardware
- having a virtual interaction mechanism for scripting and demos -- you could capture a screen recording on your PC demonstrating a script UI
- running norns on a PC for connecting to your DAW through JACK, etc?

Design issues include:
- what back ends are important? SDL and "in the browser" seem like good candidates
- how to configure what screens / options to use? @catfact made the suggestion of spinning up a separate lua VM for evaluating lua configuration and stopping it before weaver starts, which sounds quite broadly useful in matron.
- how best to structure the browser interaction? My thought right now is having matron spawn its own websocket server as a UI control interface that the maiden browser app can interact with in a similar fashion to the matron / supercollider ws-wrapper REPLs (I have the maiden side of this working on a branch here somewhere). Another option would be to have matron push data to/from the maiden go server and then from there to the browser somehow.

Interacts with https://github.com/monome/norns/pull/1052

Prosperoh:
I started working on some modifications to make it more easy to use on my local machine, picking up from where @csboling started, [branch here](https://github.com/Prosperoh/norns/tree/decouble-fb-modif). For now, the possibility to render on a bigger SDL window, and some fixes to the events.

tehn:
fixed by #1444

PR #1221, opened 2020-11-01
multiple screen backends with runtime lua configuration

csboling:
This implements #1187 and the approach taken is relevant to #1218.

- Restructure the `screen` code to allow drawing to an SDL window (for instance) instead of a framebuffer device.
- Restructure the input polling code (keys and encoders) for similar configurability. This polling code was moved to `matron/src/hardware/input/gpio.c` since it is only one type of input source that might be used (versus e.g. SDL input). 
- At startup, use a Lua configuration file to determine which hardware devices should be used. This is `~/matronrc.lua` if found, otherwise `~/norns/matronrc.lua` loads the default norns hardware configuration.
- Remove the `-f` framebuffer selection arg to matron as it is no longer used.
- To include SDL in the build, `waf` must be run with `--desktop` to avoid adding a library requirement to norns.
- Note that SDL input does not currently cooperate well with the norns main event loop, further work required to get a fully functional matron-on-desktop-linux working.


ngwese:
cool, will take a look as i'd like to extend the current screen backend to support off screen rendering and i suspect this might make that simpler.

tehn:
wow, sam, this looks great.

i read through everything, and while i should sit with it longer, my immediate confusion is the code duplication in `fd.c` and `gpio.c` (ie `enc_gpio_poll`)

this also reminds me that i could use further guidance in completing https://github.com/monome/norns/pull/1052 as it's a real drag that screen operations currently are in the same thread as lua, which creates timing lags with heavy-animated scripts

csboling:
>  code duplication in fd.c and gpio.c

Oops, this was just a botched rename at some point -- `fd.c` was not actually included in the `wscript` and was not being compiled. Removed in `49da71c` and merged in the latest changes from `main`.

ngwese:
@tehn - i agree on the note about #1052. it would be good to think about how these changes might affect future on device improvements (if at all)

tehn:
just a heads up @csboling that i haven't forgotten about this, i apologize for the delay

ngwese:
same for me. i started to look through and had some minor suggestions but wanted to give it a proper look before commenting!

tehn (commented):
this looks great! (will need a merge `main` first, sorry!)

also a good introduction (for me) to SDL, and the simple lua vm instantiation for the config is very nice. the implementation feels clean, though it does add a level of complexity--- i just learned about `TAILQ` :)

i'll defer to @ngwese and @catfact for further critique

tehn:
resolved and merged #1444

PR #1228, opened 2020-11-02
no double free for CLEAR

### LINKED ISSUE #1215: CLEAR script function can freeze sound from SuperCollider
steps to repro:
- load boingg, which uses PolySub
- press K2 to start a bouncing ball
- navigate back to the top-level menu
- while the voice is active (not on its decay), hold K1 and CLEAR the script
- sound will drone until a new script is added

expected: CLEAR silences all output

tehn:
quick fix to make sure the engine is never double-cleared

fixes #1215 better

### DIFF
--- a/lua/core/menu/home.lua
+++ b/lua/core/menu/home.lua
@@ -23,6 +23,7 @@ m.key = function(n,z)
       _menu.set_page(m.list[m.pos])
     else
       norns.script.clear()
+      _norns.free_engine()
       _menu.locked = true
     end
   end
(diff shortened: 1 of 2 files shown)


tehn:
ps, tested :)

catfact:
give me a moment and i will show what i mean

catfact:
dunno why, but i cannot seem to fetch this branch. will open a new one with my additional fixes to SC side to avoid SC errors from running this with no engine loaded.

catfact (commented):
because the SC handler for this command is kinda stale, it will throw errors. some engines may also fail to free resources if they rely on the availability of synchronization functions. (which CroneEngine promises for init / deinit methods.) it can also throw more benign errors when performed repeatedly. 

i'll just open a new PR that buttons up these cases.

tehn:
still wanted to remove this line in script.lua

Issue #850, opened 2019-06-28
separate read/display png functions


okyeron:
While looking at the png functions today I wanted to revisit the idea of separating read and display functions for pngs.

Am I on the right track with something like the following?

```
void screen_read_png(const char *filename){
	image = cairo_image_surface_create_from_png (filename);
	if(cairo_surface_status (image)) { 
		fprintf(stderr, "display_png: %s\n", cairo_status_to_string (cairo_surface_status(image))); 
		return; 
	} 
}
		
void screen_show_png(double x, double y){
	int img_w, img_h;
	img_w = cairo_image_surface_get_width (image);
	img_h = cairo_image_surface_get_height (image);

	cairo_set_source_surface (cr, image, x, y);
	cairo_rectangle (cr, x, y, img_w, img_h);
	cairo_fill (cr);
	cairo_surface_destroy (image);
}

void screen_display_png(const char *filename, double x, double y){
	screen_read_png(filename);
	screen_show_png(x, y);
}
```

artfwo:
@okyeron the code above will destroy the image after you first display it. it would be better to return an image object from `screen_read_png` (with a finalizer that will call `cairo_surface_destroy` in C). `screen_show_png` can then accept the image object as first parameter, so you can load and display images separately.

PR #1231, opened 2020-11-07
PNG read/show WIP

okyeron:
This is a work in progress PR posted in hopes of getting help with the lua/C glue.

Way back last year I started on trying to separate the read and display functions for pngs (issue #850). The idea being to be able to load a png into memory and display it multiple times without having to re-read from disk. 

I've finally revisited this and I'm looking for some assistance in finishing it up.

In the PR I've got the C side of things working in `screen.c` to separate "display_png" into read/show/destroy functions. After that I'm a bit stuck on how to glue this together in `weaver.c`. I've started some placeholder code in `weaver` but don't know where to go next.

How would someone go about getting a reference for the cairo_image object and passing that between lua and C?

### DIFF
--- a/matron/src/hardware/screen.c
+++ b/matron/src/hardware/screen.c
@@ -5,41 +5,38 @@
  */
 
 #include <assert.h>
-#include <cairo-ft.h>
-#include <cairo.h>
-#include <fcntl.h>
-#include <linux/fb.h>
+#include <unistd.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/ioctl.h>
+#include <fcntl.h>
+#include <linux/fb.h>
 #include <sys/mman.h>
-#include <unistd.h>
+#include <sys/ioctl.h>
+#include <cairo.h>
+#include <cairo-ft.h>
 
 #include "args.h"
 
 // skip this if you don't want every screen module call to perform null checks
 #ifndef CHECK_CR
-#define CHECK_CR      \
-    if (cr == NULL) { \
-        return;       \
-    }
-#define CHECK_CRR     \
-    if (cr == NULL) { \
-        return 0;     \
-    }
+#define CHECK_CR if (cr == NULL) {return; }
+#define CHECK_CRR if (cr == NULL) {return 0; }
 #endif
 
 #define NUM_FONTS 67
 static char font_path[NUM_FONTS][32];
 
-static float c[16] = {0,   0.066666666666667, 0.13333333333333, 0.2, 0.26666666666667, 0.33333333333333,
-                      0.4, 0.46666666666667,  0.53333333333333, 0.6, 0.66666666666667, 0.73333333333333,
-                      0.8, 0.86666666666667,  0.93333333333333, 1};
+static float c[16] =
+{0, 0.066666666666667, 0.13333333333333, 0.2, 0.26666666666667,
+ 0.33333333333333,
+ 0.4, 0.46666666666667, 0.53333333333333, 0.6, 0.66666666666667,
+ 0.73333333333333, 0.8, 0.86666666666667, 0.93333333333333, 1};
 
 static cairo_surface_t *surface;
 static cairo_surface_t *surfacefb;
 static cairo_surface_t *image;
+static cairo_surface_t *pngimage;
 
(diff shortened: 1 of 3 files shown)


tehn:
perhaps reconsider the end goal and you're basically done. ie, what if instead of trying to load a bunch of images, why not just have a single img buffer that you load into, and then a display function which just draws the buffer. yes, you can only have one img at a time, but that is very likely sufficient for the norns ecosystem, and the syntax doesn't require passing references to images into the lua layer.

okyeron:
Perhaps yes. 

My initial goal was implementing "sprites" of some sort - thinking that multiple images could be loaded and displayed without hitting the disk every time.

Perhaps that could be approached by loading a single "sprite sheet" of multiple images in a single png, and then getting sections of the loaded image when needed. (I believe someone on lines was talking about getting pixel values from a cairo surface - so maybe that could apply?) I'll do some further looking at the cairo docs to see if that's doable.

tehn:
are sprites supported by this?

https://github.com/p3r7/p8/blob/main/README.md

okyeron:
I'll get with @p3r7 (@eigen on lines) about that and see

catfact:
to just answer the question about passing ~~references~~ pointers to lua (whether or not it is a good idea), it is accomplished by the "lightuserdata" lua type. this is how lua represents an opaque pointer to something. (the "light" variety just means that comparison of two LUDs with the same pointer address returns `true`.)

we use this in a couple places, notably with USB devices.

e.g., when adding a MIDI device:
- from USB driver, an event of type `EVENT_MIDI_ADD` is pushed to the loop, and it's payload is a device pointer
- in main loop, these events are handled by calling `w_midi_add` with the pointer from the payload,
- and pushing the pointer to the lua stack with `lua_pushlightuserdata()`

back in lua, the userdata is passed to `Midi.new()` and stored with the device table.

and when lua calls back to C using the device:
- `_midi_send()` is called in weaver
- stack content is checked for type `LUA_TLIGHTUSERDATA`,
- retreived with `lua_touserdata()`
- and then cast/assigned to pointer of appropriate type to be used in further C function calls.

HTH

p3r7:
Hi there.

**About p8 / PICO-8 sprites**

No, [p8](https://github.com/p3r7/p8) doesn't support PICO-8 sprite functions (`spr`, `sspr`, `sset`, `sget`) as of writing.

I initially thought of supporting those in pure Lua, using a in-memory table to represent PICO-8's fixed length spritesheet.

Some tweetcarts populate the spritesheet dynamically (with `sset`) and then display part of it but that's an uncommon technique so it wasn't very high on my TODO list.

Displaying a sprite (area of the spritesheet) would be somewhat inefficient, by looping the sub-table and drawing pixels one by one. 


**About sprite support on norns**

A C / cairo approach would be much more efficient as compared to the pure Lua I initially thought of.

But I now understand that the issue with supporting an arbitrary number of sprites being loaded is that we have to have to `free` them, with a corresponding Lua-side call which is against the user-friendliness provided by Lua (thanks notably to garbage collection).

> a simpler and more limited approach would just be to set aside a judicious amount static memory for images / spritesheets. (old skool)

That looks like the best solution.

norns is limited by design (low res screen, few buttons / encoders) so having a fixed-length spritesheet fits into its philosophy.

The hard part would be:
- determining the appropriate sprite sheet size
- determining how it fits alongside the existing `screen.display_png`

p3r7:
I’ve just re-read the previous posts.

As I understand @okyeron and @tehn comments, the idea would be to restrict the loading of only one png/spritesheet into memory at a time. A new call would free up any existing in-memory buffer before loading the new one. Am I right?

This way arbitrary sprite sheet size would be supported.

catfact:
i'm sorry for the slow feedback on this, it somehow dropped off my radar. is it still being considered?

i think the idea of a single function to load a PNG, which always frees the old memory, seems safe enough.

catfact:
so, this is an interesting feature discussion, but regarding this specific PR i would like to close it.
- it's stale and conflicted
- there are a few serious code issues (added coupling between weaver/screen, unusued/unfinished functions)
- the road to desired functionality is not really clear to me, from here

maybe i can roll the spritesheet idea into the next round of breaking changes to `screen` (required by new multithreading system)

PR #1233, opened 2020-11-11
add midi mapping to binary param

schollz:
i wanted to be able to map triggers to midi keys from a keyboard. @tehn suggested building off @andr-ew's "binary" parameter which does exactly what i need (and more!). it was so lovely to discover that this was extermely easy to do - only took about 45 minutes (including learning the codebase)! 

so here is my patch, which accomplishes this mapping to midi using the binary parameter. built off of and tested with script from @andr-ew (below).

i've never submitted a patch to norns before, i'm happy to follow guidance to make this kosher.

```lu
-- param test test

cs = require 'controlspec'

px = 48
py = 16

function init()
  params:add{ type='binary', id='tog', behavior='toggle', allow_pmap=true, action=function(v) print('tog: ' .. tostring(v)) redraw()  end }
  params:add{ type='binary', id='mom', behavior='momentary', action=function(v) print('mom: ' .. tostring(v)) redraw()  end }
  params:add{ type='binary', id='trig', behavior='trigger', action=function(v) print('trig: ' .. tostring(v)) redraw() end }

  redraw()
end

function key(n, z)
  if n==1 and z==1 then
    params:delta('trig', z)
  elseif n==2 and z==1 then
    params:delta('tog', z)
  elseif n==3 then
    params:delta('mom', z)
  end
  redraw()
end

function redraw()
  screen.move(py + px, py)
  screen.level(15)
  screen.text('trig')
  screen.move(py, py * 2)
  screen.level(params:get('tog') == 1 and 15 or 2)
  screen.text('tog')
  screen.move(py + px, py * 2)
  screen.level(params:get('mom') == 1 and 15 or 2)
  screen.text('mom')
  screen.update()
end
```

### DIFF
--- a/lua/core/menu/params.lua
+++ b/lua/core/menu/params.lua
@@ -179,13 +179,6 @@ m.key = function(n,z)
           params:set(i)
           m.triggered[i] = 2
         end
-      elseif t == params.tBINARY then
-        if m.mode == mEDIT then
-          params:delta(i, 1)
-          if params:lookup_param(i).behavior == 'trigger' then
-            m.triggered[i] = 2
-          else m.on[i] = params:get(i) end
-        end
       elseif m.mode == mMAP and params:get_allow_pmap(i) then
         local n = params:get_id(i)
         local pm = norns.pmap.data[n]
@@ -193,7 +186,7 @@ m.key = function(n,z)
           norns.pmap.new(n)
           pm = norns.pmap.data[n]
           local t = params:t(i)
-          if t == params.tNUMBER or t == params.tOPTION then
+          if t == params.tNUMBER or t == params.tOPTION or t== params.tBINARY then
             local r = params:get_range(i)
             pm.out_lo = r[1]
             pm.out_hi = r[2]
@@ -439,6 +432,7 @@ m.redraw = function()
           screen.move(127,10*i)
           if t ==  params.tNUMBER or
               t == params.tCONTROL or
+              t == params.tBINARY or
               t == params.tOPTION then
             local pm=norns.pmap.data[id]
             if params:get_allow_pmap(p) then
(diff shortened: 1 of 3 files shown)


andr-ew:
awesome ! i'll test this out tomorrow with my foot pedal !

tehn:
rad! can also test later today and get it rolled into this week's update

andr-ew:
nice - footswitch mapping works great for me. I programmed a teensy to send 127 on rising edge and 0 on the falling edge, which seems standard from my quick research. the edit menu looks like it's broken though, there should be little squares mapped to K3 just like trigger, are you seeing them @schollz ? I could've fudged my git fetch

schollz:
@andr-ew ooh i forgot to check whether those squares were still alive! i guess not. i can look into it

schollz:
ok! its updated and works now. i didn't realize i broke setting the triggers from the menu but thats fixed now too.

one thing i did that doesn't seem kosher to me is this:

https://github.com/monome/norns/pull/1233/files#diff-9119f34c10f85e20b617221886c7538b95a3c1bdfc9b30b3a20bd81aaeaf18b5R620-R628

basically i was unsure how to select the right parameter to update the `m.on`/`m.triggered` to update showing the square....so i update all of them if the menu is active. is there a way to determine which index for `m.on`/`m.triggered` has been selected from inside that function?

tehn (approved):

tehn:
this works well, tested

the mapping UI is a little weird, in that it doesn't make a ton of sense what it means to have scaled in/out ranges, and accum i imagine is pretty weird and almost no applicable (trying to imagine using a infinite encoder to map toggle... mapping CW/CCW doesn't require accum even...)

however i think this is adequate, so i'm going to merge it.

PR #1234, opened 2020-11-13
+6db levels

tehn:
- adjust controlspecs for all six levels (out, in, monitor, engine, cut, tape) to allow +6db
- update MIX menu with lines for unity gain (see pic)
- remove `mix.lua` which was obsoleted, along with references to it within other files

IMG_20201113_145620


tehn:
some update files (changelog, version) slipped in, sorry.

catfact:
looks fine from here, but i won't be able to check on device til later this evening. (i'm sure it's fine, and cleaning up the dead code is great.)

tehn:
the change is trivial, and tested on hardware. i was more looking for your review re: design choice with adding "digital gain" to all channels. i think the uniformity makes sense, though the docs need to be updated to convey the ramifications of boosting noise floors, in addition to clipping (easier to do now).

alternative implementation option (that i like less) is a on/off +12db boost for only the input section.

PR #1237, opened 2020-11-15
Fix util.round_up

ryanlaws:
Shouldn't add 0.5 when using `math.ceil` - this causes e.g. 3.9 to get rounded up to 5.

### DIFF
--- a/lua/lib/util.lua
+++ b/lua/lib/util.lua
@@ -188,7 +188,7 @@ function util.round_up(number, quant)
   if quant == 0 then
     return number
   else
-    return math.ceil(number/(quant or 1) + 0.5) * (quant or 1)
+    return math.ceil(number/(quant or 1)) * (quant or 1)
   end
 end
 


ngwese (approved):
looks good, thanks for the fix

catfact (changes requested):
agree that this needs changing, but IMHO the right fix is to replace `ceil` with `floor`, as seems intended

ryanlaws:
> agree that this needs fixing, but IMHO the right fix is to replace `ceil` with `floor`, as seems intended

This sounds a little confusing to me. As far as I know, `ceil` will always round up when there is a non-zero mantissa which seems to be precisely the behavior that the name `round_up` implies, and there's already a `round` function for the floor with 0.5 offset. Can you explain this a little more? 

I apologize if I'm overlooking something obvious here. I'm happy to file another PR if needed, I just need to understand the feature a little better.

catfact:
no, sorry you are totally right, reading comprehension fail

ryanlaws:
No sweat, thanks for clarifying!

PR #1242, opened 2020-11-20
api for sending arbitrary lua

schollz:
maybe this pr is not needed but i wasn't aware of any alternative to send norns arbitrary lua via matron. if there is a way please let me know.

this commit is motivated by making a "remote controlled norns": https://llllllll.co/t/norns-online-crowdsource-your-norns/ basically i'm using this patch to be able to communicate directly to matron with arbitrarily setting keys/encoders when those inputs are manipulated remotely: https://github.com/schollz/norns.online/blob/main/main.go#L142

though i made a remote norns for fun i've been messaged that it would be useful in a classroom setting so i'd like to eliminate some of the barriers for others to run it. if this pr can be merged (or alternative suggested!) i'd like to make this "remote norns" an app that is easy to get from maiden.

### DIFF
--- a/matron/src/weaver.c
+++ b/matron/src/weaver.c
@@ -217,6 +217,7 @@ static int _sound_file_inspect(lua_State *l);
 
 // util
 static int _system_cmd(lua_State *l);
+static int _system_cmd_lua(lua_State *l);
 
 // reset LVM
 static int _reset_lvm(lua_State *l);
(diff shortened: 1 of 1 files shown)


ngwese:
unless i’m completely miss reading what is trying to be accomplished it seems like just calling Lua’s built in `loadstring` would be sufficient no?

https://www.lua.org/pil/8.html

alternatively if the goal is to send code from a remote process to matron to be evaluated the existing web socket which the maiden repl uses could also be an option.

schollz:
> alternatively if the goal is to send code from a remote process to matron to be evaluated the existing web socket which the maiden repl uses could also be an option.

@ngwese you're absolutely right lol. i made this more complicated then i needed! i am already using the websocket and didn't realize i didn't need to use a special api to do this. it works without this pr, so i will close it. thanks :+1:

PR #1250, opened 2020-11-25
platform checking

tehn:
facilities to query the host hardware:
- CM3
- Pi3 (shield)
- "other" (not tested, but feasibly detectable)

this prevents irrelevant error messages from the battery and i2c on the shield, which though harmless looks alarming.

also allows customization of some menus:
- shield, don't show battery
- shield, give better shutdown message

lua helpers:
```
norns.platform -- 2=norns 3=shield
norns.is_norns -- boolean
norns.is_shield -- boolean
```

this could've been done with a config file, but i'd rather not have the user even worry about having the right config.

ps. i'll submit a followup pr which scrubs the hardware/device formatting, it's a little messy

### DIFF
--- a/matron/src/hardware/platform.c
+++ b/matron/src/hardware/platform.c
@@ -1,0 +1,24 @@
+#include <unistd.h>
+#include <stdlib.h>
+
+#include "platform.h"
+
+#define PLATFORM_PATH "/sys/firmware/devicetree/base/model"
+
+platform_t p;
+
+void init_platform() {
+  if(access(PLATFORM_PATH, F_OK ) != -1) {
+    if(system("sudo cat /sys/firmware/devicetree/base/model | grep 'Compute'")) {
+      p = PLATFORM_PI3;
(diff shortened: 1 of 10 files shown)


matron/src/hardware/platform.c
```diff
@@ -0,0 +1,22 @@
+#include <unistd.h>
+#include <stdlib.h>
+
+#include "platform.h"
+
+platform_t p;
+
+void init_platform() {
+  if(access("/sys/firmware/devicetree/base/model", F_OK ) != -1) {
+    if(system("sudo cat /sys/firmware/devicetree/base/model | grep 'Compute'")) {
```
csboling, line 10: Is `sudo` needed here / potentially problematic? The `access` call should have already confirmed that you have read access to the file.

tehn: ah it's probably unnecessary, i was doing some unconscious cut-paste. will test/fix

PR #1252, opened 2020-11-27
add screen rotate, translate, save, & restore

tyleretters:
implementation example: https://github.com/northern-information/athenaeum/blob/main/rotate.lua

this works fine and good but i think we'll also want bunch of "rotate_rel" functions that cleanup after themselves?

```
screen.move(10, 10)
screen.rotated_text("i'm rotated ninety degrees", 90)
screen.move(10, 20)
screen.text("i'm not")
screen.move(30, 30)
screen.rotated_rect( ...
```

### DIFF
--- a/lua/core/screen.lua
+++ b/lua/core/screen.lua
@@ -233,5 +233,19 @@ end
 -- @tparam number y y position
 Screen.display_png = function(filename,x,y) _norns.screen_display_png(filename,x,y) end
 
+--- rotate
+-- @param number radians
+Screen.rotate = function(r) _norns.screen_rotate(r) end
+
+--- move origin position
+-- @tparam number x position x
+-- @tparam number y position y
+Screen.translate = function(x, y) _norns.screen_translate(x, y) end
+
+--- save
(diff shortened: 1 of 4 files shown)


tehn:
cool! good to see the screen lib getting extended.

maybe a couple helper functions (ie for text) could be useful, but generally i think low-level like your example makes sense. we should try to keep the drawing lib close-ish to the cairo style (?)

also `deg_to_rad` would be a good addition to `util.lua`

tehn (approved):

tehn:
i also want to point out that exposing `save` and `restore` make scripts even more prone to totally busting the menu system in some ways.

ie; `save` must be followed immediately by `restore` (ie, not separated by some long-timed event) otherwise if attempting to enter the menu (which does a `save`, the `restore` when re-entering PLAY) we'll hose the cairo stack if there are any drawing calls (save/restore) outside the `redraw` function (which is blocked when inside the menu).

just to say... keep all drawing inside `redraw`!

Issue #1251, opened 2020-11-26
higher-level keyboard input module


catfact:
@tehn expressed desire for a higher-level keyboard interface in lua, which would automatically be used when a device supporting keyboard interface is connected. whereas now, we don't distinguish between different HID classes. 

i think such a distinction would have to happen at the C level

tehn:
my goal (i think this is a worthy goal?) is to make the keyboard a special case rather than a generic HID. this way we could manage its behavior (ie, to control the menu/passwords/etc) with a focus-sharing layer, and also provide the keycode decoding by default (right now keyboard-centric scripts all have redundant libs for managing the keyboard).

some backnotes:

> it's pretty easy to grab the keyboard via `evdev` similarly to how we collect key/encoder data in matron presently. devices can be queried for names, and "keyboard" shows up in that name, so i could autoselect. weirdly (as seen in HID menu also) keyboards enumerate twice, and it's the first instance that sends data (tested with only two keyboards, however). succeeded in testing similarly on the laptop... keyboard shows up fine.
>
> one issue with usb keyboards is a disconnection (ie, dev path vanishes) will hose the thread, flooding repeat data. so i'm not sure if a path monitoring thread is also required, to kill the process when disconnected.
>
> here's a good working example: https://github.com/darkelement/simplistic-examples/blob/master/evdev/evdev.c           
>
> trying to get more info from udev brought me back to considering smarter device detection (ie, check manufacturer strings) which is weirder than i'd hoped... because you have to traverse the parent nodes until you get a device with the manufacturer attribute. for a crow it's two steps back, but for FTDI devices the manufacturer string is something like 4 parents up, so i'd need to write a smarter traversal mechanism.
>
> the attr tree can be read such: `udevadm info --name=/dev/ttyUSB0 --attribute-walk`
>
> but back to the keyboard... might be able to check an attr for device class, rather than trying to match the word "keyboard" (i mean, that's how the kernel would do it, right...)

catfact:
IMHO: this does not have to be a complicated change or a parallel device type in C.

yes: we can check an `evdev` device attribute to see whether a given HID device supports the keyboard interface class. we can pass this flag on to lua when the device is added and take it from there.

catfact:
(i have code for this somewhere. will see if i can get the interface-detection step working this evening.)

catfact:
ok... actually, code i have is for lower level USB host driver. there we can inspect interface subclass and proctocol (keyboard and mouse.)

[...]

PR #1255, opened 2020-11-28
Detect hid device class

catfact:
first step in addressing issue #1251 

- event type and code tables are now copied to HID device object on construction. [note 1]

- added small module `hid_device_class` to query event types and codes for specific interface profiles. currently this just detects whether a device is an ASCII keyboard or a mouse. [note 2]

- add `isAsciiKeyboard` and `isMouse` flags to the device object. these could be used for all kinds of purposes: change whether/how device is shown in menu, add keycode->charcode conversion and callback for key input, etc.

- add a pretty trivial utility function to `tabutil.lua`, which inverts keys/values in a shallow table. (if one is going to repeatedly test whether various values are present in the table, this is way more efficient than looping over it many times. maybe having it as a util function will encourage that optimization.)

- minor fixes to luadoc comments in `hid.lua`.

----

[1] might not really be necessary / desirable to copy all this data for every device, if flagging the interface type suffices.

[2] easy enough to add other profiles like tablets and gamepads, but i don't have any on hand ATM


tehn (commented):
this looks great--- and super glad this can all be done in lua (i was in a misguided `evdev` wormhole)

i'll test this on device before merging, and then set upon some menu/global logic to handle focus.

thank you millions!

tehn:
re #1 (copying data table) i don't think it's a big deal...

ngwese:
super minor note. the ‘isAsciiKeyboard’ and friends should probably be `is_ascii_keyboard` (snake case) to match other APIs

tehn:
tested on device--- works great.

catfact:
renamed functions and fields to snake case, and removed some wayward debug prints. 

seems ok to me know? if approved, i can move forward with adding a text input class and a locale-switching mechanism.

tehn (approved):

Issue #1248, opened 2020-11-24
PARAMETERS: mapping improvement springboards


dndrks:
I've been thinking a lot about how the mapping function works on PARAMETERS, so I just wanted to collect some reflection here in case it's helpful. if this feels like noise, just lmk.

### multi-map
- _current_: re-mapping a control to another parameter removes it from its prior definitions
- _proposed_: map many parameters to a single control
- _challenges_: provide user a way to quickly un-link the control from all mappings

### range ([see this thread](https://llllllll.co/t/norns-param-map-screen/32352))
- _current_: a mapped control will access the full range of the parameter
- _proposed_: allow the user to refine the range of the parameter, so that (for example) a slider can focus on a subset of the parameter's values. could lead to really expressive control when combined with multi-map.

### arc
- _current_: an arc cannot control parameters unless the loaded script explicitly supports it or if the arcify library is used
- _proposed_: unless a script defines arc interaction, arc could default to a mapping mode that allows the first three encoders to be mapped to any parameter. the fourth encoder could switch between multiple banks, so an arc could control up to 12 parameters across 4 banks (for example)

### midi callback
- _current_: if a user maps a controller that has LED feedback (thinking specifically of Midi Fighter Twister or OP-Z), LEDs do not display the loaded state. user has to turn the encoder all the way to the extreme in order to align the values. if a script sets params values directly, this decouples things pretty quickly.
- _proposed_: provide a method to report relative parameter values as a "midi echo" back to the transmitting device on a specified channel, with a specified type (CC, note on, etc), mapped 0-127. this would also open up a potentially dumb but fun-sounding creative opportunity to map a parameter's changes to MIDI notes lol
- _challenges_: this might be too niche, but could be a nice MIDI QOL improvement

tehn:
re: range, an individual issue should be made for this, it's a bug

p3r7:
for future reference regarding OSC feedback:
- Zack implemented a small [OSC feedback mod](https://discord.com/channels/765746584582750248/870538207748571186/873656361475538965) that allows forwarding all incoming OSC events.
- i've implemented the [osc-cast](https://github.com/p3r7/osc-cast) mod that sends an OSC messages for all param changes. it's more suitable for bidirectional param editing.

those mods might be sufficient for the time being. midi feedback could be implemented in a similar fashion.

sixolet:
[...]

PR #1257, opened 2020-12-02
Allow mapping control params over non-integer ranges

### LINKED ISSUE #1249: parameter mapping: 'out' range doesn't facilitate floats
from [this thread / post](https://llllllll.co/t/norns-param-map-screen/32352/5):

a mapped control's `out` range is definable by integers only, which works well for certain types of parameters (options, binary on/off, etc). however, parameters with float ranges between 0 and 1 or which have musical needs for floats (cutoff frequency, etc) are not able to be mapped in useful ways.

**proposed behavior**: K3 + E3 on `out` range could add the same fine-tune precision offered on the standard parameter screen, allowing users to set a range of floats to which their `in` values map.

synthetiv:
This PR allows users to map `control` params with 'out' min/max values between 0 and 1 -- plus, it displays the output range using the controlspec for the param being mapped, so the user can choose scaled/warped values that match what they'd see in the params menu, instead of raw 0.0-1.0 values.

Resolution/precision is determined by the controlspec's `quantum` property, just like `params:delta()`.

mapping screen

`number` params work just like they did before.

I've been running this code since last weekend with no problems so far.

Fixes #1249, as mentioned in #1248.

### DIFF
--- a/lua/core/menu/params.lua
+++ b/lua/core/menu/params.lua
@@ -341,10 +341,22 @@ m.enc = function(n,d)
       elseif m.mpos==7 then
         pm.in_hi = util.clamp(pm.in_hi+d, pm.in_lo, 127)
         pm.value = util.clamp(pm.value, pm.in_lo, pm.in_hi)
-      elseif m.mpos==8 then
-        pm.out_lo = pm.out_lo + d
-      elseif m.mpos==9 then
-        pm.out_hi = pm.out_hi + d
+      elseif m.mpos==8 or m.mpos==9 then
+        local param = params:lookup_param(n)
+        local min = 0
+        local max = 1
+        if t == params.tCONTROL then
+          d = d * param.controlspec.quantum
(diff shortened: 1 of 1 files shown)


synthetiv:
oh, I wrote this code before seeing @dndrks's K3 idea. I still think control params should use float/controlspec-based mapping by default, but the added fine control is a good idea too, I'll see if I can add that later tonight.

synthetiv:
that should do it!

tehn (approved):
wonderful!

synthetiv:
@tehn I actually just found & fixed a bug in the fine control (although that's not merged yet) -- and added support for `taper` params on my branch. Want a separate PR for that?

tehn:
separate PR is fine, and good catch on the bug. i didn't test on device, the code looked good :)

synthetiv:
@tehn done! #1258

PR #1262, opened 2020-12-04
allow params:read to read silently

schollz:
i've been doing workarounds to get various params to act silently when reading new parameters. basically i have parameters that trigger reading parameters, and so it can get into loops when that parameters is banged upon reading. i implemented my own params_read using `silent` and it works fine, but i thought it might be worth adding in an option to the core to read parameters silently since it is already built into the `set` command.


tehn (approved):
make sense, thanks!

PR #1263, opened 2020-12-05
continue if params:read throws error on parameter

schollz:
i'm working on scripts to share between users and one thing i'm running into is if users have different versions of a script and thus might be missing a parameter in their script when loading from someone else's parameter file. currently, using `params:read` will cause an error and the reading will be canceled in this case. 

if this is the behavior that you want (perhaps to prevent downstream errors) - that's totally fine and i can make a function to work around it. 

however, just in case, i thought i'd make a pr to continue reading, even if parameters aren't able to be set when reading, and set the parameters that are available. (this might be amended to print something to the console about it too, but i'd like to know if this is of interest first).

### DIFF
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -407,13 +407,13 @@ function ParamSet:read(filename, silent)
 
           if index and self.params[index] then
             if tonumber(value) ~= nil then
-              self.params[index]:set(tonumber(value), silent)
+              pcall(function() self.params[index]:set(tonumber(value), silent) end)
             elseif value == "-inf" then
-              self.params[index]:set(-math.huge, silent)
+              pcall(function() self.params[index]:set(-math.huge, silent) end)
             elseif value == "inf" then
-              self.params[index]:set(math.huge, silent)
+              pcall(function() self.params[index]:set(math.huge, silent) end)
             elseif value then
-              self.params[index]:set(value, silent)
+              pcall(function() self.params[index]:set(value, silent) end)
             end
           end
         end


tehn:
doesn't line 408 catch and prevent trying to read a nonexistent param? i guess i am missing something?

definitely the intention is for a param read to complete even if there are irrelevant entries

schollz:
i doubt your missing something - i'm probably missing something. i'll look into it some more. but a bit of context: i have an optional menu. i tried saving a parameter set with and without this optional menu. i then turn the menu off and try loading both parameter sets

loading the parameter set that i saved with the menu off gives no problem:

```
load_name: /home/we/dust/data/awake/names/AB
loading AB
pset >> read: /home/we/dust/data/awake/AB/parameters.pset
<ok>
```

but then, when the optional menu is off and i load a parameter set that was saved with the optional menu on i get a problem:

```
load_name: /home/we/dust/data/awake/names/AB
loading AB
pset >> read: /home/we/dust/data/awake/AB/parameters.pset
lua: /home/we/norns/lua/core/params/option.lua:34: bad argument #1 to 'floor' (number expected, got string)
stack traceback:
        [C]: in function 'math.floor'
        /home/we/norns/lua/core/params/option.lua:34: in function 'core/params/option.set'
        /home/we/norns/lua/core/paramset.lua:416: in function 'core/paramset.read'
        /home/we/dust/code/awake/lib/sharer.lua:181: in global 'backup_load'
        /home/we/dust/code/awake/lib/sharer.lua:55: in field 'action'
        /home/we/norns/lua/core/params/file.lua:60: in function 'core/params/file.bang'
        /home/we/norns/lua/core/params/file.lua:35: in function 'core/params/file.set'
        (...tail calls...)
        /home/we/norns/lua/core/menu/params.lua:282: in function 'fileselect.callback'
        /home/we/norns/lua/lib/fileselect.lua:50: in function 'fileselect.exit'
        /home/we/norns/lua/lib/fileselect.lua:133: in function 'fileselect.key'
        /home/we/norns/lua/core/menu.lua:139: in function </home/we/norns/lua/core/menu.lua:116>
<ok>
```

here's the difference in the two parameter files:

```
diff parameters_menuon.pset parameters_menuoff.pset 
21,24d20
< "share_upload": /home/we/dust/data/awake/names/
< "share_download": /home/we/dust/data/norns.online/virtualdir/awake/
< "share_refresh": 0
< "share_message": 
```

somehow this latter case gets past the `self.params[index]` check...let me see if i can figure out why.

schollz:
ok, I'm on norns 3b4d7dac37a7c9aef85c364d1763e3a6cb750926 (latest) and can reproduce with clean awake/data folders:

```
> cd ~/dust/code && git clone https://github.com/tehn/awake
> rm -rf ~/dust/data/awake
```

Run awake and save `PSET`. Then open the `awake-01.pset` and add this line seemingly anywhere:

```
"share_message":
```

then awake won't load anymore as it gets stuck trying to load that file. I don't get why it gets stuck, as you say, it should be checking and seeing that `share_message` does not exist as a parameter. unless `share_message` is a special string?

```
# script init
starting halfsecond
pset >> read: /home/we/dust/data/awake/awake-01.pset
### SCRIPT ERROR: init
attempt to compare string with number
stack traceback:
        /home/we/norns/lua/core/norns.lua:138: in function </home/we/norns/lua/core/norns.lua:138>
        [C]: in function 'math.max'
        /home/we/norns/lua/lib/util.lua:99: in function 'util.clamp'
        /home/we/norns/lua/core/params/number.lua:40: in function 'core/params/number.set'
        /home/we/norns/lua/core/paramset.lua:416: in function 'core/paramset.read'
        /home/we/norns/lua/core/paramset.lua:429: in function 'core/paramset.default'
        /home/we/dust/code/awake/awake.lua:268: in global 'init'
        /home/we/norns/lua/core/script.lua:105: in function 'core/script.init'
        [C]: in function 'xpcall'
        /home/we/norns/lua/core/norns.lua:139: in field 'try'
        /home/we/norns/lua/core/engine.lua:91: in function </home/we/norns/lua/core/engine.lua:89>
```

tehn:
i may be reading this incorrectly, but is the error getting thrown because
of a data mismatch?

ie, the param index exists, but has a different data type? (ie, if you had
param name "super" which is a number, but a pset data tries to load a
string... yes, this error isn't currently covered...) so your fix would
correct this. though instead of pcalling, perhaps we should just add data
type checking?

schollz:
ok, I'm on norns 3b4d7dac37a7c9aef85c364d1763e3a6cb750926 (latest) and can reproduce with clean awake/data folders:

```
> cd ~/dust/code && git clone https://github.com/tehn/awake
> rm -rf ~/dust/data/awake
```

Run awake and save `PSET`. Then open the `awake-01.pset` and add this line seemingly anywhere:

```
"share_message":
```

then awake won't load anymore as it gets stuck trying to load that file. I don't get why it gets stuck, as you say, it should be checking and seeing that `share_message` does not exist as a parameter. unless `share_message` is a special string?

```
# script init
starting halfsecond
pset >> read: /home/we/dust/data/awake/awake-01.pset
### SCRIPT ERROR: init
attempt to compare string with number
stack traceback:
        /home/we/norns/lua/core/norns.lua:138: in function </home/we/norns/lua/core/norns.lua:138>
        [C]: in function 'math.max'
        /home/we/norns/lua/lib/util.lua:99: in function 'util.clamp'
        /home/we/norns/lua/core/params/number.lua:40: in function 'core/params/number.set'
        /home/we/norns/lua/core/paramset.lua:416: in function 'core/paramset.read'
        /home/we/norns/lua/core/paramset.lua:429: in function 'core/paramset.default'
        /home/we/dust/code/awake/awake.lua:268: in global 'init'
        /home/we/norns/lua/core/script.lua:105: in function 'core/script.init'
        [C]: in function 'xpcall'
        /home/we/norns/lua/core/norns.lua:139: in field 'try'
        /home/we/norns/lua/core/engine.lua:91: in function </home/we/norns/lua/core/engine.lua:89>
```


also, weird to me is that you can restart norns `~/norns/stop.sh; sleep 1; ~/norns/start.sh` after which it will load just fine.

schollz:
@tehn, nvm, this is so strange and complicated. i can only seem to reproduce this after i run my version of awake, then deleting everything and then cloning your version. does something save within norns between restarts about each script's parameters? please don't spend anymore time on this, i'll try to figure out whats going on..its probably not affecting anyone else.

schollz:
yeah confirming it reproduces by running my version of awake, then saving something, then deleting and running your version of awake and then saving something, and then adding that `share_message` line to that save and then it won't open.

schollz:
i'll close this until i can figure it out some more!

PR #1267, opened 2020-12-09
ws-wrapper prints

catfact:
this reverts PR #1028

yea or nay?

original problem:
- weird recursive duplicated output when you try and log the `matron` service from lua
- too many logs?

new problem:
- not enough logs?

### DIFF
--- a/ws-wrapper/src/main.c
+++ b/ws-wrapper/src/main.c
@@ -177,5 +177,6 @@ int main(int argc, char **argv) {
         printf("usage: ws-wrapper WS_SOCKET BINARY <child args...>");
     }
 
+    setvbuf(stdout, NULL, _IONBF, 0); 
     launch_exe(argc, argv);
 }
--- a/ws-wrapper/src/main.c
+++ b/ws-wrapper/src/main.c
@@ -66,6 +66,7 @@ void *loop_tx(void *p) {
         nb = read(pipe_tx[PIPE_READ], buf, PIPE_BUF_SIZE - 1);
         if (nb > 0) {
             buf[nb] = '\0';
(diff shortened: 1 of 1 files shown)


tehn (approved):

catfact:
i'm having a weird time with this for some reason after re-adding the prints.

as expected, i see output both on the shell and on the websocket when i manually execute from ssh/bash:

```
/home/we/norns/build/ws-wrapper/ws-wrapper ws://*:5556 /usr/bin/sclang
/home/we/norns/build/ws-wrapper/ws-wrapper ws://*:5555 /home/we/norns/build/matron/matron
```

but somehow, same output is not showing up in service logs after reboot using:
```
journalctl -ef -u norns-matron
```

i remember this working before and am probably doing something wrong.

simonvanderveldt:
> but somehow, same output is not showing up in service logs after reboot using:
> 
> ```
> journalctl -ef -u norns-matron
> ```
> 
> i remember this working before and am probably doing something wrong.

@catfact I'm seeing the same thing and remember this being the case since ws-wrapper was introduced. I don't think I've ever had the logs going into ws-wrapper show up in the journal for me.

In any case, I think this is caused by buffering. AFAIK on interactive terminals everything is usually line buffered but this isn't when it's redirected to something like the journal. I'll do some tests.

simonvanderveldt:
@catfact I don't think the changes to `crone/softcut` are intended?

catfact:
ah jeez

catfact:
sorry i completely forgot about this. reverted the unintentional change to softcut submodule revision. i think this is good to merge.

Issue #1260, opened 2020-12-03
FR: show all 16 slots in system > devices > midi


dndrks:
## current behavior

according to MAP's `dev` selector, norns can host up to 16 devices. however, only 4 slots are shown in SYSTEM > DEVICES > MIDI, which makes device management difficult for users.

since norns doesn't "forget" disconnected devices, difficulty can be compounded -- a user may end up in a situation where the only _presently_ connected MIDI device isn't displayed without manual management.

this assumed four device limit has also become canonical, since many scripts mirror this in their midi i/o -- tambla, for instance, only allows MIDI input or output for slots 1-4.

## proposed behavior

we could show all 16 slots in the MIDI list, so users don't have to worry about manually managing their devices. this would allow users to "set and forget" a bit more, where all of their devices would already be registered + allocated.

I believe this would also help ensure that MIDI mapping and script i/o settings in PSETs would be recalled to the appropriate devices, because the device list would be less subject to change.

dndrks:
closed via #1268

PR #1268, opened 2020-12-10
PMAP autosave + 16 MIDI vports

### LINKED ISSUE #1261: parameters > edit screen can be made inaccessible after manual script clear
## repro steps

- connect to maiden for error monitoring
- load a script with parameters (awake works well)
- navigate to PARAMETERS > EDIT
- use E2 to scroll down deep into the script's params (eg. on awake, nav down to `probability`)
- turn E1 CCW navigate to main menu and CLEAR script with K1+K3
- try to turn E1 CW to navigate back to PARAMETERS > EDIT

**experienced behavior**
- screen will seemingly stay stuck on main menu (I think it does move to PARAMETERS > EDIT, but can't render due to below)
- maiden prints:
```lua
lua: /home/we/norns/lua/core/paramset.lua:357: invalid paramset index: 56
stack traceback:
	[C]: in function 'error'
	/home/we/norns/lua/core/paramset.lua:357: in function 'core/paramset.lookup_param'
	/home/we/norns/lua/core/paramset.lua:301: in function 'core/paramset.t'
	/home/we/norns/lua/core/menu/params.lua:410: in function 'core/menu/params.redraw'
	/home/we/norns/lua/core/menu/params.lua:608: in field 'event'
	/home/we/norns/lua/core/metro.lua:169: in function </home/we/norns/lua/core/metro.lua:166>
```
- turn E2 CCW and you'll suddenly see the PARAMETERS > EDIT menu
- turn E2 CW, but you won't be able to navigate past REVERB
- if you try to press K2 to back out to parent PARAMETERS menu with REVERB selected, you won't be able to and maiden will print additional errors about an `invalid paramset index`
- you'll need to turn E2 CCW to resurface the params and select LEVELS
- you can now press K2 to back out to parent PARAMETERS menu
- select EDIT and everything is now sorted

dndrks:
### params.lua
**previous behavior**:
- midi mappings (PMAPs) were only saved when a PSET was saved

**new behavior**:
- PMAPs are saved after they're assigned, so they persist between script loads without needing to save a PSET

## 16 MIDI vports

**previous behavior**:
- norns could host up to 16 MIDI devices, but only assigned devices to 4 MIDI vports
- SYSTEM > DEVICES > MIDI reflected only 4 MIDI vports
- users had to remove/reassign vport-assigned devices more often
- this removing/reassigning disrupted accuracy of mapped controls in PMAPs (which restores based on vport device ID)

**new behavior**:
- norns can now assign up to 16 MIDI vports to match its hosting capability
- users can manage all 16 vports via SYSTEM > DEVICES > MIDI
- 16 device history means users will have to do much less device management/rearranging
- restoring static vport device IDs via PMAPs is less problematic, unless the user purposefully rearranges their device

SYSTEM > DEVICES > MIDI:
mididev1
mididev2

assigned a connected device to a vport (up to 4 pages can generate, no indicator if only 1 page):
mididev3

### midi.lua
- create 16 MIDI vports (addresses #1260)

### pmap.lua
- expect assignments from all 16 MIDI vports

### state.lua
- remember devices from all 16 MIDI vports

### devices.lua
- add menu redraw handling for 16 MIDI vports
- add `m.last_pos` to remember which port was last highlighted when switching back to port list from device list

### clock.lua
- extended clock params table to include all 16 ports
- added naming (with abbreviations) to help make it easy for users to see port assignments
midiout
nb. of course, names are refreshed only at boot, but i think this is helpful regardless. if a device's name is none, it just shows the port number.

### params.lua
**previous behavior**: mapping a midi controller to a parameter showed only its port number
**new behavior**: midi map menu now shows the device's name (abbreviated if longer than 6 characters, eg. "Midi Fighter Twister" becomes "MFT"), which helps with clarity

new-map
new-map2

### home.lua
- added parameter menu reset

### util.lua
- util.acronym() will return an acronym from a given string

eg. `util.acronym("Midi Fighter Twister")` returns `MFT`

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -234,9 +234,9 @@ function clock.add_params()
     end)
   params:set_save("link_quantum", false)
   local clock_table = {"off"}
-  for i = 2,17 do
-    local short_name = string.len(midi.vports[i-1].name) < 12 and midi.vports[i-1].name or util.acronym(midi.vports[i-1].name)
-    clock_table[i] = "port "..(i-1)..""..(midi.vports[i-1].name ~= "none" and (": "..short_name) or "")
+  for i = 1,16 do
+    local short_name = string.len(midi.vports[i].name) < 12 and midi.vports[i].name or util.acronym(midi.vports[i].name)
+    clock_table[i+1] = "port "..(i)..""..(midi.vports[i].name ~= "none" and (": "..short_name) or "")
   end
   params:add_option("clock_midi_out", "midi out",
       clock_table, norns.state.clock.midi_out)
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -233,9 +233,13 @@ function clock.add_params()
       norns.state.clock.link_quantum = x
     end)
   params:set_save("link_quantum", false)
-
+  local clock_table = {"off"}
+  for i = 2,17 do
+    local short_name = string.len(midi.vports[i-1].name) < 12 and midi.vports[i-1].name or util.abbreviate(midi.vports[i-1].name)
+    clock_table[i] = "port "..(i-1)..""..(midi.vports[i-1].name ~= "none" and (": "..short_name) or "")
+  end
   params:add_option("clock_midi_out", "midi out",
-      {"off", "port 1", "port 2", "port 3", "port 4"}, norns.state.clock.midi_out)
+      clock_table, norns.state.clock.midi_out)
   params:set_action("clock_midi_out", function(x) norns.state.clock.midi_out = x end)
   params:set_save("clock_midi_out", false)
   params:add_option("clock_crow_out", "crow out",
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -235,7 +235,7 @@ function clock.add_params()
   params:set_save("link_quantum", false)
   local clock_table = {"off"}
   for i = 2,17 do
-    local short_name = string.len(midi.vports[i-1].name) < 12 and midi.vports[i-1].name or util.abbreviate(midi.vports[i-1].name)
+    local short_name = string.len(midi.vports[i-1].name) < 12 and midi.vports[i-1].name or util.acronym(midi.vports[i-1].name)
     clock_table[i] = "port "..(i-1)..""..(midi.vports[i-1].name ~= "none" and (": "..short_name) or "")
   end
   params:add_option("clock_midi_out", "midi out",
--- a/lua/core/menu/devices.lua
+++ b/lua/core/menu/devices.lua
@@ -145,7 +145,8 @@ m.redraw = function()
       if m.section == "midi" then
         screen.move(0,20)
         screen.level(4)
-        screen.text("[ "..util.round_up(m.pos/4).."/"..util.round_up(#m.options[m.section]/4).." ]")
+        local page_indicator = util.round_up(#m.options[m.section]/4)
+        screen.text(page_indicator > 1 and "[ "..util.round_up(m.pos/4).."/"..page_indicator.." ]" or "")
         local positions = {{1,4},{5,8},{9,12},{13,16}}
         for j = positions[util.round_up(m.pos/4)][1],positions[util.round_up(m.pos/4)][2] do
           if(j==m.pos) then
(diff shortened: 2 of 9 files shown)


tehn:
rad! will check it out.

lua/core/menu/home.lua
```diff
@@ -3,6 +3,8 @@ local m = {
   list = {"SELECT", "SYSTEM", "SLEEP"}
 }
 
+local p = require 'core/menu/params'
```
tehn, line 6: `params` is globally imported i don't think you need to do this?

lua/core/menu/home.lua
```diff
@@ -24,6 +26,7 @@ m.key = function(n,z)
     else
       norns.script.clear()
       _norns.free_engine()
+      p.reset()
```
tehn, line 29: can just do `params.reset()`

lua/core/clock.lua
```diff
@@ -233,9 +233,13 @@ function clock.add_params()
       norns.state.clock.link_quantum = x
     end)
   params:set_save("link_quantum", false)
-
+  local clock_table = {"off"}
+  for i = 2,17 do
```
tehn, line 237: maybe more clear to do `i=1,16` and then below, `clock_table[i+1]` and then get rid of all the `-1`

tehn (commented):
this is great! some minor inline comments to address but tested it out, nice attention to detail.

one big change i'd like you to tackle first, however, is you introduce a new scrolling/paging style for the device midi menu. i'd much prefer you make it conform the the existing style which is followed by the SELECT and PARAM lists: basically six lines displayed, highlight on line 3. for device/midi the minor mod to this would be MIDI on line 1, line 2 blank, then have lines 3-6 scrolling while keeping line 3 the highlight.

awesome work!

lua/core/menu/home.lua
dndrks: ah, excellent -- couldn't grok how to get the params reset in there

lua/core/menu/home.lua
dndrks: 🙏🏻

lua/core/clock.lua
dndrks: lol, YES, _absolutely_

ryanlaws: @dndrks this was the thing I had the question about, rad 👍🏻

lua/core/menu/devices.lua
```diff
@@ -75,8 +76,8 @@ m.key = function(n,z)
         hid.update_devices()
       end
       m.mode = "list"
-      m.len = 4
-      m.pos = 1
+      m.len = m.section == "midi" and 16 or 4
```
ryanlaws, line 79: Doesn't this mean the same thing as #L38 and #L60? Just wondering why the logic was flipped here. When I was skimming this at first they looked like weird magic numbers but I get it now (e.g. grid vs. MIDI).

lua/lib/util.lua
```diff
@@ -217,4 +217,15 @@ function util.rads_to_degs(radians)
   return radians * (180 / math.pi)
 end
 
+--- convert string to acronym
+-- @tparam string name
+-- @treturn string acronym
+function util.acronym(name)
```
ryanlaws, line 223: Nice, I might steal this!

lua/core/menu/devices.lua
dndrks: good eye! just changed mid-code, no reason for the break. will unify :)

dndrks:
@tehn + @ryanlaws , thank you for the detailed feedback!

i've aligned the nav mechanics of the DEVICES > MIDI device menu + submenu with the scrolling template of SELECT + PARAMETERS. i also touched up some of the legibility, as highlighted by your q's and comments.

lmk if this feels good and tests well!

tehn:
code looks good, will test on device this afternoon!

tehn (approved):

lua/core/menu/home.lua
```diff
@@ -24,6 +24,7 @@ m.key = function(n,z)
     else
       norns.script.clear()
       _norns.free_engine()
+      params.reset()
```
tehn, line 27: little late--- this throws an error, and i'm wondering what the original intention was?

dndrks:
mmm. dang. it was to avoid a screen redraw issue after clearing the script
and returning to parameters (maybe to adjust the clock or fine-tune a
global param in between script loads).

i can take a peek later this week, sorry for the trouble!

tehn:
ah i remember now! this is an easy fix i think. i'll get it

PR #1269, opened 2020-12-10
move key watcher to separate application

tehn:
tested, works. in order, K3-K2-K1 held for 10s.

now we can reboot when matron crashes.

systemctl service file in https://github.com/monome/norns-image/pull/82


simonvanderveldt:
Are there any potential issues with two applications reading the button states?

tehn:
afaik, no? it's just two processes both watching a file for events.

simonvanderveldt:
OK :)

catfact:
i'm ok with it, can't test for a few hours

tehn:
it occurred to me that this process could actually monitor is `matron` is running, and relaunch if not.

then the K123 would only be needed for a locked/unresponsive matron.

ie, it might be nice to know matron broke right away (by having it restarting) instead of wondering for some time if it'd recover (when clearly it won't, if the process has died)

tehn:
of course this could be annoying for system dev and debugging, but the watcher process is easily stopped with systemctl

catfact:
yeah that is what i had in mind. if watcher knows matron `pid` it could potentially also create stack traces for us.

simonvanderveldt:
> it occurred to me that this process could actually monitor is `matron` is running, and relaunch if not.

How would it do so? And why don't we let systemd take care of that?

catfact:
i imagine it would be done by making `matron` a child process of `watcher` and using ptrace or something

but i am a doofus and don't know much about systemd. if there's a way to say "do some specific thing when this process dies unexpectedly" then that seems like it would be very nice

simonvanderveldt:
So the way to make use of systemd's watchdog functionality is to enable it in the unit and make sure the application you're running enables its watchdog functionality based on the `NOTIFY_SOCKET` env var. It should then send a simple datagram every half of the configured interval (defined via the `WATCHDOG_USEC` env var) to the process specific socket (the one listed in the `NOTIFY_SOCKET` env var). If no response is received in time the process can be restarted (or any other action that's possible similar to when the process dies)
I'm not aware of a standalone application that we could use as glue, but the interface is very simple, so I think we can just simply implement this in matron itself. There's a library that's part of systemd called [sd_notify](https://www.freedesktop.org/software/systemd/man/sd_notify.html) that can be used for this. Alternatively we could hack something together using [systemd-notify](https://www.freedesktop.org/software/systemd/man/systemd-notify.html) (I don't think this is a very good idea/elegant solution) or send the datagrams ourselves to the correct socket. An example of how that works can be found here, as you can see from this link it's really very simple https://github.com/stigok/sd-notify/blob/master/sd_notify.py#L33

Some more details can be found here http://0pointer.de/blog/projects/watchdog.html and a hacky way using systemd-notify is described here https://www.medo64.com/2019/01/systemd-watchdog-for-any-service/

PR #1276, opened 2020-12-22
add pull to “fetch”

pq:
useful for me but i won't be offended if this stretches the original intent to (and past) a breaking point.

tl;dr: if a script has already been cloned (or if it looks like it has because there's an existing directory with its name), do an actual `"fetch"`.

/cc @tehn

### DIFF
--- a/lua/core/norns.lua
+++ b/lua/core/norns.lua
@@ -179,10 +179,10 @@ norns.fetch = function(url)
   local script = parts[#parts]
   local checkout = paths.code.."/"..script
 
-  -- fetch if already downloaded.
+  -- pull if already downloaded.
   if (util.file_exists(checkout)) then
-    print(script.." already cloned; fetching instead...")
-    command = "cd "..checkout.."; git fetch"
+    print(script.." already cloned; pulling instead...")
+    command = "cd "..checkout.."; git pull"
   end
 
   local status = os.execute(command)
--- a/lua/core/norns.lua
+++ b/lua/core/norns.lua
@@ -172,7 +172,20 @@ end
 
 -- fetch (git clone)
 norns.fetch = function(url)
-  local status = os.execute("cd "..paths.code.."; git clone "..url)
+  -- default to clone.
+  local command = "cd "..paths.code.."; git clone "..url
+
+  local parts = tab.split(url, "/")
+  local script = parts[#parts]
+  local checkout = paths.code.."/"..script
+
+  -- fetch if already downloaded.
+  if (util.file_exists(checkout)) then
+    print(script.." already cloned; fetching instead...")
+    command = "cd "..checkout.."; git fetch"
+  end
+
+  local status = os.execute(command)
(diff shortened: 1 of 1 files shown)


pq:
and needless to say, there just has to be a more elegant way to do this.  happy to be schooled!

ngwese:
wouldn’t it make more sense to do `pull` instead of `fetch` ? ...fetch alone won’t change the files on disk.

also fwiw, any git repo in the scripts dir will be treated as a project in the `maiden` installed page so clicking update on that page will do the equivalent of a `git pull`

pq:
> wouldn’t it make more sense to do pull instead of fetch

umm, yeah?... 😬 i think i meant to do a fetch and then a pull but that's sure not what it has here.  thanks!

> also fwiw, any git repo in the scripts dir will be treated as a project in the maiden installed page

i noticed that.  super cool.  if that's the preferred way to update (and we don't want to land this) maybe we can update the output to nudge folks to use the install page flow?

pq:
ok, well.  it should at least _work_ now.  not super invested either way.

thanks for the review!

simonvanderveldt:
> also fwiw, any git repo in the scripts dir will be treated as a project in the maiden installed page

Doesn't this also mean that updating would work with the cli as well?

tehn (approved):
excellent, thanks!

tehn:
> Doesn't this also mean that updating would work with the cli as well?

yes, but you can do this via the maiden interface for repos in the community collection.

the primary use case for this command was for people to be able to get repos not in the community collection (listed in maiden), ie, something brand new that someone posts to the forum

simonvanderveldt:
> yes, but you can do this via the maiden interface for repos in the community collection.
> 
> the primary use case for this command was for people to be able to get repos not in the community collection (listed in maiden), ie, something brand new that someone posts to the forum

@tehn I was thinking that if a user has added a script using fetch they can already update it using the catalog on maiden or using the maiden cli on norns so there's no need to add this functionality.

It would probably be better to incorporate the fetch functionality into the maiden cli: if you pass a repo URL to `install` it will install it for you the same as it does for repos from the catalog.

pq:
> Tbh it would probably be better to incorporate the fetch functionality into the maiden cli: if you pass a repo URL to install it will install it for you the same as it does for repos from the catalog.

ah, and i hadn't even though of this and had entirely forgotten about the maiden CLI and was just in the repl and wanting to stay in one flow.  i do think there's something appealing about that.  (one less thing to remember.)  on the other hand, having two ways to do the same thing has it's down-sides.

maybe we should explore the "install an unpublished script" user journey in an issue and hammer out a more uniform solution?

simonvanderveldt:
> maybe we should explore the "install an unpublished script" user journey in an issue and hammer out a more uniform solution?

Yeah, sounds like a good idea :+1:
Might also inform us about how well the mainden CLI currently works/is used :)

PR #1277, opened 2020-12-23
constrain a positive number to a positive min/max range

tyleretters:
i end up writing a function like this for just about every script. it is most often used to keep track of transport steps, but has other applications.

```lua
-- "cycle" back around to 1
util.cycle(17, 1, 16) -- 1 (first beat of 2nd measure)

-- recursive to handle "over cycling"
util.cycle(33, 1, 16) -- 1 (first beat of 3rd measure)
util.cycle(37, 1, 16) -- 5 (fifth beat of 3rd measure)
util.cycle(1024, 1, 16) -- 16 (last beat of 16th measure)

-- adjustable min if you want to confuse yourself
util.cycle(13, 2, 7) -- 6
util.cycle(13, 4, 10) -- 7

-- no change to a value within the range
util.cycle(1, 1, 16) -- 1
util.cycle(13, 1, 16) -- 13
```

### DIFF
--- a/lua/lib/util.lua
+++ b/lua/lib/util.lua
@@ -228,4 +228,19 @@ function util.acronym(name)
   return (name:gsub("%s+", ""))
 end
 
+--- constrain a positive number to a positive min/max range
+-- @tparam number value
+-- @tparam number min
+-- @tparam number max
+-- @treturn number cycled value
+function util.cycle(value, min, max)
+  if value > max then
+    return util.cycle(value - max, min, max)
+  elseif value < min then
+    return util.cycle(max - value, min, max)
+  else
+    return value
+  end
+end
+
 return util
--- a/lua/lib/util.lua
+++ b/lua/lib/util.lua
@@ -228,12 +228,12 @@ function util.acronym(name)
   return (name:gsub("%s+", ""))
 end
 
---- wrap a positive number to a positive min/max range
--- @tparam number value
+--- wrap a number to a positive min/max range
+-- @tparam number n
 -- @tparam number min
 -- @tparam number max
 -- @treturn number cycled value
-function util.wrap(value, min, max)
+function util.wrap(n, min, max)
   local y = value
   local d = max - min + 1
   while y > max do
@@ -243,6 +243,15 @@ function util.wrap(value, min, max)
     y = y + d
   end
   return y
+end
+
+--- wrap a number to a positive min/max range but clamp the min
+-- @tparam number n
+-- @tparam number min
+-- @tparam number max
+-- @treturn number cycled value
+function util.wrap_max(n, min, max)
+  return util.wrap(util.clamp(n, min, max), min, max)
 end
 
 return util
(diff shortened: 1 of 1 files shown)


lua/lib/util.lua
```diff
@@ -228,4 +228,19 @@ function util.acronym(name)
...
 end
 
+--- constrain a positive number to a positive min/max range
+-- @tparam number value
+-- @tparam number min
+-- @tparam number max
+-- @treturn number cycled value
+function util.cycle(value, min, max)
+  if value > max then
+    return util.cycle(value - max, min, max)
```
catfact, line 238: the recursion trick is neat. but... it is pretty inefficient compared to a simple while-loop here. compare this version:

```
function util.cycle2(value, min, max)
    local y = value
    local d = max - min + 1
    while y > max do
      y = y - d
    end
    while value < min do
      y = y + d
    end
    return y
  end
  ```
  
  as the input gets really large, the second version runs much faster because it isn't adding function calls to the stack each time. (and eventually, the recursive version will eat all the stack; whereas with the non-recursive version we can add a counter for safety.)
  
  so this is one instance where i think using recursion isn't needed and isn't quite the right call.
  
  this is nitpicking, but its sort of important for a fundamental operation like this.

**Review by catfact — CHANGES_REQUESTED** [review]
even though it is clever, i'd lose the recursion in this case

File: lua/lib/util.lua (hunk 1 above)
Comment (tyleretters) [reply lua/lib/util.lua]: oh wow this is way better

**Comment by tyleretters** [comment]
fixed while loop bug @simonvanderveldt discovered

**Comment by tyleretters** [comment]
renamed to `wrap`

**Review by tehn — APPROVED** [review]

**Comment by tyleretters** [comment]
is it naughty to call other `util` functions?

```
--- wrap a number to a positive min/max range but clamp the min
-- @tparam number n
-- @tparam number min
-- @tparam number max
-- @treturn number cycled value
function util.wrap_max(n, min, max)
  return util.wrap(util.clamp(n, min, max), min, max)
end
```

seems worse to duplicate the whole algorithm again

catfact:
seems 100% a-ok to me

simonvanderveldt:
> is it naughty to call other `util` functions?

I've been super conscious of doing as little computation as possible where I can for my scripts, especially the sequencing part. That's why I generally don't use any or only very little abstractions in them. Now this might be over-optimizing but my thinking is we're already a ways off from hard realtime, any small thing I can do to get it as close as possible is a gain. Or phrased differently, it's easy to use some additional CPU cycles here and there but in the end they do add up.

Problem of course is the difficulty of measuring things like this given that the code is running in an interpreter that's running in a general purpose OS and some (or all) of these (micro)optimizations might get dwarfed by some other process eating up the CPU. :man_shrugging:

tyleretters:
i fixed a bug and cut a the function calls.

---

not calling other functions is a good heuristic. if that hasn't been done until this point (quick scan says pretty much: no) i don't want to start. this is maintainable enough that the duplication is probably ok.

tehn:
@catfact is this good to go?

catfact (approved):

PR #1279, opened 2020-12-24
Fix nil value error in Script.load

kevinkyyro:
I was exploring maiden and ran into an error. I created a simple `print("hello there")` script at `home/we/dust/untitled.lua` and got the following output when running it in maiden:

```
norns.script.load("untitled.lua")
lua: 
/home/we/norns/lua/core/script.lua:131: attempt to concatenate a nil value (field '?')
stack traceback:
	/home/we/norns/lua/core/script.lua:131: in function 'core/script.load'
	(...tail calls...)
```

So, I dug into the `Script.load` function in `script.lua` and got the general idea of what it was doing, but it was a bit confusing with the "hack for 3-deep subfol[d]ers" and all, so I tried to clean it up and got it working for my simple script. I could be missing some assumptions or cases the previous code was handling, but this seems correct.

I verified that various apps seem to work (or at least load without error when I wasn't sure how to _use_ them), including cyrene and orca, which have multiple levels of subfolders.

### DIFF
--- a/lua/core/script.lua
+++ b/lua/core/script.lua
@@ -117,8 +117,8 @@ Script.load = function(filename)
     name = norns.state.name
     path = norns.state.path
   else
-    filename = string.sub(filename,1,1) == "/" and filename or _path["dust"].."/"..filename
-    name = filename:match("^.*/(.+)$")
+    filename = string.sub(filename,1,1) == "/" and filename or _path["dust"]..filename
+    name = filename:match("^.*/(.+)$"):match("^([^.]+).*$")
     path = filename:match("^(.*/).+$")
   end
 
--- a/lua/core/script.lua
+++ b/lua/core/script.lua
@@ -118,8 +118,16 @@ Script.load = function(filename)
     path = norns.state.path
   else
     filename = string.sub(filename,1,1) == "/" and filename or _path["dust"]..filename
-    name = filename:match("^.*/(.+)$"):match("^([^.]+).*$")
-    path = filename:match("^(.*/).+$")
+    path, scriptname = filename:match("^(.*)/([^.]*).*$")
+    name = string.sub(path, string.len(_path["code"]) + 1)
+    
+    -- append scriptname to the name if it doesn't match directory name in case multiple scripts reside in the same directory
+    -- ex: we/study/study1, we/study/study2, ...
+    if string.sub(name, -#scriptname) ~= scriptname then
+      name_parts = tab.split(name, "/")
+      table.insert(name_parts, scriptname)
+      name = table.concat(name_parts, "/")
+    end
   end
 
   print("# script load: " .. filename)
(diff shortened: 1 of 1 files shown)


tehn:
sorry for that messy code. i had something very specific in mind for how to handle scripts inside of subfolders.

can you confirm it still works? try the monome/we scripts as a start. if they load and save psets correctly i think we're good. i can test this later this weekend otherwise.

thanks!

simonvanderveldt:
I don't think this `home/we/dust/untitled.lua` is supported, scripts should live in their own directory. We might want to return a more useful error message for this case though?

ngwese:
This is most likely happening (in the `untitled.lua` case) due to a bug in `maiden` https://github.com/monome/maiden/issues/160

tehn:
ah, good point.

i've been using `dust/untitled.lua` without any trouble editing remotely w/o maiden (and launching with hardware menu)

kevinkyyro:
> can you confirm it still works? try the monome/we scripts as a start. if they load and save psets correctly i think we're good. i can test this later this weekend otherwise.

Good tip, turns out I did break psets, but managed to get them working again.

> i had something very specific in mind for how to handle scripts inside of subfolders.

I think I see the intent a bit more clearly. Am I missing anything here?
- `filename` is the absolute path to the lua script
- `path` is the absolute path to the directory of the lua script
- `name` is the relative path from `/home/we/dust/code` to the file
   - special case: to support multiple scripts in a single directory, append the scriptname when it doesn't match the directory name

I updated the code to handle these semantics (and fixed psets). It's a bit less simple than it was before, but it handles the `untitled.lua` case and is a bit more explicit about the intent.

> I don't think this `home/we/dust/untitled.lua` is supported, scripts should live in their own directory. We might want to return a more useful error message for this case though?

I guessed it might have been the odd file placement, but from my initial read of the code there didn't seem to be a clear intent to disallow having scripts above `/home/we/dust/code`, so I just tried fixing the loading. 

Now, I see that there were assumptions in the code about where the script would be. I tried to make those assumptions more explicit and hopefully easier to follow.

I can see things potentially getting a bit whacky with psets and audio if the loading isn't made more strict, but it might be a smoother experience if monome/maiden#160 is fixed before making the loading strict? Even with the strictness, the `name` handling still benefits from some clarity I think.

lua/core/script.lua
```diff
@@ -111,30 +111,23 @@ end
...
-	  name = t[#t-1].."/"..t[#t]
-	end
-  if #t==4 then name = t[2].."/"..name end -- dumb hack for 3-deep subfolers
-	path = string.sub(_path["dust"],0,-2)
-	for i = 1,#t-1 do path = path .. "/" .. t[i] end
-	--print("name "..name)
-	--print("final path "..path)
+    filename = string.sub(filename,1,1) == "/" and filename or _path["dust"]..filename
+    path, scriptname = filename:match("^(.*)/([^.]*).*$")
+    name = string.sub(path, string.len(_path["code"]) + 1)
```
kevinkyyro, line 122: The `+ 1` assumes that `_path["code"]` doesn't have a trailing slash. If that's an uncomfortable assumption, I can make it check.

tehn: `_path["code"]` is hard-defined so we're safe

tehn (approved):

Issue #1280, opened 2020-12-25
softclock


tyleretters:
here's a first draft of [softclock](https://github.com/tyleretters/softclock/blob/main/lib/Softclock.lua
), based on @catfact's [mockup](https://gist.github.com/catfact/614fcea7bd1c6d1fe1cbdd57b4211580). [usage here](https://github.com/tyleretters/softclock/blob/main/softclock.lua).

- uses norns system clock
- allows for multiple softclocks to be instantiated
- allows for pulses to be changed on the fly
- allows for any subclock to be toggled on or off
- allows for global softclock to be toggled on or off
- allows for user definable callbacks for both "advance" and "pulse" events
- encapsulates the `clock.run` and `clock.cancel`

tyleretters:
- added an optional `meter` param
 ---
is adding this file to the `norns/lua/lib` as simple as dropping it in?
https://github.com/monome/norns/blob/main/lua/core/config.lua#L11

tehn:
yes, drop in `lib` and add to `startup.lua` if it's going to be a global.

i like the usability introduced here.

i'm slightly concerned that we have too many clock approaches:

- metro
- clock
- softclock

is there any sense to consider unifying these in any way?

@artfwo any perspectives on this?

artfwo:
`clock` provides the **coroutine-based** interface, that enables multiple entry points in functions (via sleep/sync calls with arbitrary multipliers of the base tempo). `clock` can also sync to midi, link, and the internal metronome.

Both `metro` and `softclock` (i've only briefly looked at the latter) provide **callback-based** interfaces. i don't know the benefits of callbacks, but the scripts that use either of these two modules will require additional state data and logic for the callbacks.

For instance, the [softclock example](https://github.com/tyleretters/softclock/blob/main/softclock.lua) linked above looks very terse with nothing but print statements in the callbacks. However, using this approach for something less trivial (looping musical phrases) will require additional state variables (such as playhead positions) in the global namespace - a problem that `clock` tries to solve by encouraging using coroutines to the full extent.

The additional benefit of using coroutines is having multiple sync points inside a single function, which allows to describe musical processes in a really elegant functional way. In some cases it can also make the code more intuitive.

That said, I'd suggest deprecating `metro`, since `clock` can do 100% of what `metro` can do and much more. As for `softclock`, I don't really understand what problem it solves. @tyleretters what's the use-case for it?

I also see a minor problem with the API design of softclock - string identifiers for subclocks, e.g.
```lua
[...]

PR #1286, opened 2020-12-29
add lattice

tyleretters:
based on #1280 !

example usage: https://github.com/tyleretters/lattice

```lua
lattice = require("lattice")

-- basic lattice usage
my_lattice = lattice:new()

-- make some tracks
track_a = my_lattice:new_track(1/2, function() print("half notes") end)
track_b = my_lattice:new_track(1/4, function() print("quarter notes") end)

-- optionally add a track that is disabled
track_c = my_lattice:new_track(1/8, function() print("eighth notes") end, false)

-- start the lattice
my_lattice:start()

-- user definable callbacks
my_lattice.downbeat = function(transport)
  print("downbeat", transport)
end

-- another one, but in in 5/4 and 128 ppqn
my_other_lattice = lattice:new(5, 128)

-- global lattice controls
my_lattice:stop()
my_lattice:start()
my_lattice:toggle()
my_lattice:destroy()
my_lattice:set_meter(7)
my_lattice:set_ppqn(48)

-- individual track controls
track_a.stop()
track_a.start()
track_a.toggle()
track_a.destroy()
track_a.set_division(1/7)

```

### DIFF
--- a/lua/lib/lattice.lua
+++ b/lua/lib/lattice.lua
@@ -1,0 +1,106 @@
+--- module for creating a multitrack sequencer based on a single fast "superclock"
+--
+-- @module Lattice
+-- @release v1.0.0
+-- @author tyleretters & ezra
+
+local Lattice = {}
+
+--- instantiate a new lattice
+-- @tparam[opt] number meter number of quarter notes per measure, defaults to 4
+-- @tparam[opt] number ppqn the number of pulses per quarter note of this superclock, defaults to 96
+-- @treturn table a new lattice
+function Lattice:new(meter, ppqn)
+  local s = setmetatable({}, { __index = Lattice })
+  s.meter = meter ~= nil and meter or 4
+  s.ppqn = ppqn ~= nil and ppqn or 96
+  s.downbeat = nil -- downbeat callback
+  s.transport = 0
+  s.is_playing = false
+  s.superclock_id = nil
+  s.tracks = {}
+  return s
+end
+
+--- start running the sequencer
+function Lattice:start()
+  if not self.superclock_id then
+    self.superclock_id = clock.run(self.pulse, self)
+  end
+  self.is_playing = true
+end
+
+--- stop the sequencer
+function Lattice:stop()
+  self.is_playing = false
+end
+
+--- toggle the lattice
+function Lattice:toggle()
+  self.is_playing = not self.is_playing
+end
+
+--- destroy the lattice
+function Lattice:destroy()
+  self:stop()
+  clock.cancel(self.superclock_id)
+  self.tracks = {}
+end
+
+--- set the meter of the lattice
+-- @tparam number meter the meter the lattice counts
+function Lattice:set_meter(meter)
+  self.meter = meter
+end
+
+--- set the ppqn of the lattice
+-- @tparam number ppqn the pulses per quarter note
+function Lattice:set_ppqn(ppqn)
+  self.ppqn = ppqn
+end
+
+--- advance all tracks in this lattice a single by pulse
+-- @tparam table s this lattice
+function Lattice.pulse(s)
+  while true do
+    clock.sync(1/s.ppqn)
+    if s.is_playing then
+      s.transport = s.transport + 1
+      if s.downbeat ~= nil and s.transport % (s.ppqn * s.meter) == 1 then
+        s.downbeat(s.transport)
+      end
+      for id, track in pairs(s.tracks) do
+        if track.is_playing then
+          track.phase = track.phase + 1
+          if track.phase > (track.division * s.ppqn * s.meter) then
+            track.phase = track.phase - (track.division * s.ppqn * s.meter)
+            track.event(track.phase)
+          end
+        end
+      end
+    end
+  end
+end 
+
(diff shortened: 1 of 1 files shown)


tehn:
example/api looks great! will check out the code later today.

ngwese:
Looking at this only at a code level I would recommendation using a metatable with `track` as well so the method invocation was consistent (`.` vs `:`)

tyleretters:
that's a great point. i'll get that going. on a similar note, i dislike `my_lattice.downbeat = ...`. should we perhaps pass in the function with `new()`?

tyleretters:
or this `my_lattice:set_downbeat(function(transport) print("downbeat", transport) end)`

ngwese:
Sorry for the piecemeal comments, haven’t found a moment to provide detailed feedback. One additional thought to consider, `track` => `player`, in the sense that a player drives a sequence or plays a track.

tyleretters:
i like the spirit of the idea but don't love the word `player`.

maybe `lattice` is too poetic and we could go with `sequencer` and `sequence`?

pq:
at first blush i really like this.  cheers!

a quick API thought... perhaps consider using tables for passing (named) params?  that would make things more explicit and i think more legible.  for example:

```lua
 lattice:new(5, 128)
```

could become:

```lua
 lattice:new{meter = 5, ppqn = 128}
```

(doing away with a need for explanatory comments like  these:


```lua
  -- another one, but in in 5/4 and 128 ppqn
  -- my_other_lattice = lattice:new(5, 128)
```
)

`new_track` would benefit from this style especially imo.

tyleretters:
@pq thanks! awesome idea to use a table for the args. refactored. this solved a couple other style issues, too!

also how about `track` -> `pattern` ?

updated usage examples: https://github.com/tyleretters/lattice/blob/main/lattice.lua

```lua
-- basic lattice usage (uses defaults)
my_lattice = lattice:new()

-- named params, each optional
my_lattice = lattice:new{
  meter = 4,
  ppqn = 96,
  callback = function(t) print("transport position:", t) end
}

-- make some patterns
pattern_a = my_lattice:new_pattern{
  division = 1/2,
  callback = function() print("half notes") end
}
pattern_b = my_lattice:new_pattern{
  division = 1/4,
  callback = function() print("quarter notes") end
}
pattern_c = my_lattice:new_pattern{
  division = 1/8,
  callback = function() print("eighth notes") end,
  enabled = false
}

-- start the lattice
my_lattice:start()
```

ryanlaws:
This is a really cool feature and overall I like the API. A few suggestions:
- `clock.sync` is a perfectly reasonable default as `pulse`'s timing source, but exposing this for overriding would enhance its utility, enabling e.g. LFOs, modulated clock sources, arpeggiators with irregular patterns. 
- It is a little confusing to me that both `Lattice` and `Pattern` provide callbacks. They're definitely valuable in `Pattern` but they seem extraneous in the parent `Lattice`. The use case of keeping track of `s.transport` may be a non-issue as it is public to the instantiator of the `Lattice`.
- The type check on like 97 could be replaced with a simple `or` since the default is `false`, and that type checking could be moved to `Pattern:set_callback`. This seems computationally cheaper (which may be significant in this time-critical scenario) and would also reduce nesting in `Lattice.pulse`.
- Removing a pattern which isn't the last pattern could cause duplicate pattern IDs and potentially orphan a pattern based on the size-based ID generation method on L123. It's probably safer to initialize a per-`Lattice` ID counter and increment it for every new `Pattern`.
- Requiring passing the pattern itself to remove it as in `Lattice:destroy_pattern` feels weird to me, especially since it's just grabbing its `id` property. It does look like `Lattice.patterns` is public and so could be used to enumerate the patterns. Maybe removing a pattern by passing its `id` would feel less weird.

tyleretters:
lots of refactoring: https://github.com/tyleretters/lattice

@ryanlaws

- interesting. i broke out `.pulse` into `.auto_pulse` and `:pulse`. you can now pass a flag in to disable the automatic behavior and then call `my_lattice:pulse()` manually if you wish. cool idea.
- good feedback on the dual callbacks. i removed the `Lattice` one and it actually smoked out a minor inconsistency with timings.
- good feedback on the type checking. updated.
- good point on the ID issues. i added a counter as you suggested. it still feels somewhat loose to me.
- yeah, i don't like passing the pattern to lattice to destroy it either. i had it the other way (with the ID) but there were some issues with that, too. I HAVE A NEW SOLUTION: `pattern_a:destroy()`
- i also removed the `lattice:set_ppqn()` method - if you need to update `ppqn` in real time this module isn't for you.

tyleretters:
should i take a look at anything else on this one?

tehn:
i'll give it another review!

tehn:
i think the broad usefulness of this lib greatly improved with the addition of the manual pulse.

API suggestion: we've been avoiding the term "callback" in preference of "action" (ie in params), so it seems that should continue here?

tyleretters:
yeah! @ryanlaws was right about that one for sure.

yes. i've renamed callback to action.

tehn (approved):

Issue #777, opened 2019-03-25
separate cairo thread


tehn:
screen drawing currently blocks

cairo would have its own event queue

PR #1289, opened 2021-01-04
Screen events

catfact:
addresses issue #777 

- adds dedicated event queue for screen drawing in `screen_events.h`, `screen_events.c`
- most screen functions in `weaver` now post to the screen Q instead of blocking

this seems to have the desired results...

test script with heavy drawing, midi handling and metro:
https://github.com/catfact/norns-tests/blob/master/screen_timing.lua

previously, reported time spent in `redraw` was ~26ms, and increasing frame rate above 15FPS would noticeably affect timing of metro and midi handling, and eventually soft-lock the system. 

now, reported redraw time is ~14ms, and FPS can be pushed to 50 without soft-locking. (timing does begin to degrade, much less noticeably.)

questions:
- screen event API uses varargs instead of the weird stuff with unions in the main event Q. is this better?
- screen event Q is a statically allocated circular buffer, instead of a dynamically allocated linked list. allocations are only made for strings. is this better? 

if the answer to both is "yes" then i might refactor the main event q to follow suit.

caveats:
- needs fuller testing of all drawing commands. (i ran a few community scripts with heavy drawing, but have not systematically tested the full drawing API.)
- needs checking for leaks with valgrind/massif.
- `screen_poke` still blocks, because i didn't work out a way to pass the data buffer through the Q.


matron/src/screen_events.c
```diff
@@ -0,0 +1,199 @@
...
+    dst->type = src->type;
+    dst->data_count = src->data_count;
+    for (int i=0; i<src->data_count; ++i) {
+        dst->data[i] = src->data[i];
+    }
+    if (src->text != NULL) {
+        int nbytes = strlen(src->text) + 1;
+        dst->text = malloc(nbytes);
+        memcpy(dst->text, src->text, nbytes);
+        free(src->text);
```
ngwese, line 79: given that event structures are reused the src text pointer should probably be set to NULL to prevent a double free when reused

matron/src/screen_events.c
```diff
@@ -0,0 +1,199 @@
...
+    data->data_count = data_count;
+    if (text != NULL) {
+        int nbytes = strlen(text) + 1;
+        data->text = malloc(nbytes);
+        memcpy(data->text, text, nbytes);
+    } else {
+        data->text = NULL;
+    }
+    va_start(args, data_count);
+    for (int i=0; i<data_count; ++i) {
```
ngwese, line 49: recommend bounds checking `data_count` against the max number of `data` values accommodated in the event struct (currently 6). in my book I’d probably just use an assert at the top of the function.

matron/src/screen_events.c
ngwese: alternatively consider avoiding the malloc, memcpy, and free here and simply assign the text pointer value to the destination event then setting src text to NULL - since then free is being handled in `screen_event_loop`

matron/src/screen_events.c
```diff
@@ -0,0 +1,199 @@
...
+        memcpy(data->text, text, nbytes);
+    } else {
+        data->text = NULL;
+    }
+    va_start(args, data_count);
+    for (int i=0; i<data_count; ++i) { 
+        data->data[i] = va_arg(args, double);
+    }
+    va_end(args);
+    screen_q_wr = (screen_q_wr + 1) & SCREEN_Q_MASK;
```
ngwese, line 53: if the write position in the queue were to pass the read position in the queue (and the drawing thread was lagging) it seems like there is the potential to leak memory associated with the text value in the event. not sure if that is a big enough concern to warrant checking.

ngwese (commented):
looks like a pretty nice addition to me. easy to follow logic.

to the question of varargs vs union structures. personally, while i like the compactness of the varargs style I’d probably still lead toward the explicit union of structs because the casting is happening in one place and if debugging in gab say one could/would perform the same cast then be able to print the event contents clearly.

matron/src/screen_events.c
ngwese: now that i’m thinking about it it might be nice to provide a `screen_event_free(...)` to encapsulate the deallocation of the text pointer if set and setting it to NULL.

catfact:
thanks @ngwese, great feedback.

pushed some changes:
- bounds checking on data_count: OK

- cleanup `event_data.text` management , skip extra copy/alloc: OK

- check for queue wrapping: ended up putting this at the end of `screen_event_push`, where write pointer is incremented. (seemed like the only reliable possibliity.)

this raised some issues; turns out the event queue is getting full during the stress test with high FPS. (so, it was leaking memory before.) question in my mind is what to do about this:
  - we could crash, but that seems a little harsh and maybe unhelpful for script development unless we at least add some very obvious feedback about what happened.
   - we can drop the oldest event easily enough; this is what happens now.
   - additionally, we're now printing an error message. this is actually not ideal because the print itself takes too long and soft-locks the system, kinda self-defeating. maybe we should try and fire an event back to lua so we can at least put a warning on screen....?
   
- i didn't rewrite the varargs stuff... i don't disagree with the assessment, but at the moment it will be a bit of a PITA to make that change, and the benefit seems minimal.

- still not sure what to do about `screen_poke`. my confusion lies in understanding how the pixel data is supposed to be encoded in the string argument to the lua call; API comments aren't enlightening me. we can certainly use the same buffer as for strings but might need to explicitly store the buffer size rather than relying on `strlen` (maybe that's a good idea anyway.)

tehn:
> we can drop the oldest event easily enough; this is what happens now.
>
> additionally, we're now printing an error message. this is actually not ideal because the print itself takes too long and soft-locks the system, kinda self-defeating. maybe we should try and fire an event back to lua so we can at least put a warning on screen....?

i think this approach is very smart.

tehn:
`screen_poke` seems to be simply using the string arg as an array of unsigned char bytes. the decoding is here: https://github.com/monome/norns/blob/main/matron/src/hardware/screen.c#L527

the values will be constrained from 0-15 (i believe).

the buffer size could be deduced by the height and width args, but perhaps just treating it like a string is fine?

---

otherwise, this PR looks great, thank you! and thanks @ngwese for the detailed review

catfact:
i see how the poke data is decoded. what i'm missing is an example of the lua side API usage. this is kinda important, because under the hood, `lua_checklstring` calls `lua_tolstring` which actually coerces whatever is on the stack into a C string - and i believe the `len` is inferred from location of null byte just like in `strlen`.

in that context, the `>=` here gives me pause:
```
if (buf && len >= w * h) {
```

----------

maybe my question can be boiled down to: can the buffer contain `\0`s?

if not, i can just cast it to a string.

if so, i'll refactor all the buffer handling to use an explicit byte count.

tehn:
ah yes that is a very good point.

@csboling could you clarify the lua API?

ngwese:
random thought - on the topic of event queue overflow what about dropping _everything_ on the queue and inserting a SCREEN_CLEAR ahead of the overflowing event instead of just dropping the last event? 

it seems desirable to be able to clearly see visually when the overflow occurs. whether dropping the last event is visible on not on screen depends a lot on what the event is. if the event was a position move or a blending mode change or rotation, subsequent drawing could land off screen  or otherwise be hard to spot. explicitly clearing the screen and ending up with partial draw erase;to might make it more obvious what it going on from a user perspective.

catfact:
i love that idea.

tehn:
agreed

csboling:
>  i believe the len is inferred from location of null byte just like in strlen.

This is not actually the case, while the [API doc](https://www.lua.org/manual/5.3/manual.html#lua_tolstring) says it "Converts the Lua value at the given index to a C string." it is using perhaps an idiosyncratic definition of "C string" since it continues: "lua_tolstring returns a pointer to a string inside the Lua state. This string always has a zero ('\0') after its last character (as in C), but can contain other zeros in its body." The [implementation](https://github.com/lua/lua/blob/e2ea3b31c94bb3e1da27c233661cb2a16699c685/lapi.c#L381) seems to, in the case where the value on the stack is a Lua string, simply set `*len` to the stored string length `vslen(o)` and returns the byte array without modifying it or seeking for a null terminator. I wrote a quick C extension:

```C
static int _checklstring_len(lua_State *l) {
    lua_check_num_args(1);
    size_t len;
    luaL_checklstring(l, 1, &len);
    lua_pushinteger(l, len);
    return 1;
}
```

and confirmed that;

```lua
> _norns.checklstring_len("123")
3
> _norns.checklstring_len("\x00");
1
> _norns.checklstring_len("\x00\xff\x00\xff\x00")
5
```

---

Indeed the buffer is expected to possibly include null bytes and the encoding stores one byte per pixel of the peeked screen region, each byte in the range 0 to 15 inclusive -- I should definitely have remarked on this in the ldoc, will make a PR presently. The contents of the encoded buffer get converted to the screen's encoding [here](https://github.com/monome/norns/blob/d0dd1d1266af60d10a34a91f28e8bcc156576c7c/matron/src/hardware/screen.c#L529) by just duplicating this value 3 times into a 24-bit value -- I am somewhat puzzled by this since the Cairo data surface is actually configured as having an RGB565 pixel format [here](https://github.com/monome/norns/blob/d0dd1d1266af60d10a34a91f28e8bcc156576c7c/matron/src/hardware/screen.c#L153), but I believe when I tried reconstructing the pixel data in 565 order the results were incorrect.

catfact:
thanks so much for setting me straight there. 

ok, the path forward seems clear; will try and push remaining changes tonight.

catfact:
i'd consider this PR mergeable, it just has limitations: `poke` still blocks the main thread. 

i did make an attempt to refactor, clean up, support `poke`, and drop the varargs... this is branch `screen-events-2` and can be seen in PR #1295. i'm experiencing crashes with the newer version and could use any assistance spotting the concurrency bug.

catfact:
... i take it back. this is susceptible to the same crash as the refactor, which i now think is evidence of cairo's un-threadsafe-ness.

PR #1290, opened 2021-01-05
break norns.shutdown() out into a function

tyleretters:


lua/core/menu/sleep.lua
```diff
@@ -4,18 +4,8 @@ m.key = function(n,z)
   if n==2 and z==1 then
     _menu.set_page("HOME")
   elseif n==3 and z==1 then
-    print("SLEEP")
-    --TODO fade out screen then run the shutdown script
     m.sleep = true
-    _menu.redraw()
```
tehn, line 10: _menu.redraw() should be left in sleep.lua i think

tyleretters:
moved `_menu.redraw()` back. the TODO came with, too.

tehn (approved):

PR #1292, opened 2021-01-05
API help via REPL

tehn:
opening this as a proof of concept, looking for suggestions on formatting:

- what terminal char width should be encouraged? the horizontal dashes are 80 chars wide.
- table design: currently not aligning the descriptions, wrapping will be tight
- format of functions: not sure why i added vars for example return value functions
- should there be examples?? or save that for the quick-reference scripting page?


tehn:
i feel like i've landed on a reasonable convention, but still up for suggestions:

```
help()
--------------------------------------------------------------------------------
help(topic): grid, clock
--------------------------------------------------------------------------------
<ok>
help(grid)
--------------------------------------------------------------------------------
grid.connect( port )          create a grid table using device [port]
                                default [port] 1 if unspecified
                              (returns) grid table
.key( x, y, z )               function called with incoming grid key event
                                this should be redefined by the script
.led( x, y, level )           set LED at [x,y] to [level]
                                [level] range is 0..15
.all( level )                 set all grid LED to [level]
                                [level] range is 0..15
.refresh()                    update the grid LED state
--------------------------------------------------------------------------------
-- example
lx,ly,lz = 0,0,0
-- connect grid
g = grid.connect()
-- key function
g.key = function(x,y,z)
  print(x,y,z)
  lx = x
  ly = y
  lz = z*15
  draw_grid()
end
-- simple draw function
draw_grid()
  g.all(0)
  g.led(lx,ly,lz)
  g.refresh()
end
--------------------------------------------------------------------------------
```

merging this so it serves as a template to complete more docs

PR #1295, opened 2021-01-11
screen events, 2nd attempt

catfact:
this is an attempt to finish, refactor, and address review for #1289 .

unfortunately, *this version is crashing*. 

- the crash is easily reproducible with `arcologies`, but it happens eventually with other scripts. 
- have not narrowed it down to a specific drawing call
- it is a concurrency issue; becomes less frequent with less optimization and more debug prints.
- it looks like smashed stack memory and manifests in a segfault in one or another cairo call (usually `text_extents`.)

have been staring at this for a while but cannot spot the thing, so i'm putting this up here in case someone else has time to look...

---------------

update: as detailed below, crash seems to be due to simultaneous access to the main `cairo_t` instance in `screen_text_extents()`. adding locks around `cr` fixes this. 

`peek` and `poke` do not directly access `cr`, only data in the main drawing surface, so they do not crash. however i've also added locks to these in order to prevent partial reads/writes to the surface data.


catfact:
oh hey... maybe this is because we now have multiple threads making cairo calls. (in particular, main thread still calls `screen_peek` and `screen_text_extents`)

catfact:
.. yes, that seems to be it. (replacing those fn's with no-ops seems to prevent the crash.)

so this now requires a callback mechanism... ideally i guess that would be a function argument in the lua call, stored in the [lua registry](https://www.lua.org/pil/27.3.1.html) or something.

tehn:
thanks for the deep dive-- let's get on a call about this when you've got time

catfact:
ok, the main workaround i can come up with now is to protect access to the main cairo surface behind another lock, and block on acquiring the lock for screen functions that must return values to lua.

this of course would work against the optimization goal by blocking the main thread. but entering an uncontested lock has negligible cost, so the ultimate outcome (and the real lesson here, i suppose) is that `peek()` and `text_extents()` are relatively expensive calls from the point of view of main thread timing. the design of a drawing-heavy script should take this into account where possible...

catfact:
added locks around the cairo instance. 

i can't notice any performance impact on this stress-test without `text_extents`:
https://github.com/catfact/norns-tests/blob/master/screen_timing.lua

but if anyone wants to try turning off the locks, the macro/flag is here:
https://github.com/monome/norns/blob/screen-events-2/matron/src/hardware/screen.c#L26

_arcologies_ now works without crashing again. but i do notice some text-position glitching on the main page that i think is new...

tehn:
will do some testing later, thanks!

tehn:
tested a few scripts, feeling good. haven't done a code review yet.

it'd be good to get a few more people testing, particularly those experiencing any sort of issue with heavy screen scripts (calling discord people)

here's how to fetch/compile:
```
cd norns
git fetch origin pull/1295/head:1295
git checkout 1295
./waf configure
./waf -j 2
systemctl restart norns-matron
```

ngwese:
i will make an effort to do some testing later today/tomorrow.

tehn (approved):

ngwese:
fwiw -`tambla` is fairly glitchy w/ this change in place (text is bouncing back and forth on the screen). i have a feeling it is due to the aggressive redraw speed (which i need to fix) and probably the use of text_extents() to right justify things.

catfact:
yeah that sounds like the same text-bouncing glitch as in arcologies.

i can't specifically figure out what is causing this. going to try refactoring `screen_text_extents` to just return a struct instead of a pointer to static data.

ngwese:
curiously dropping the frame range down low didn't change the behavior but swapping any use of `text_center` or `text_right` to be just `text` looks to resolve the jumping for the parameters i can still see. i'll see if i can narrow it down more.

tehn:
wondering if there's actually a better way to do text alignment in cairo w/o text_extends

catfact:
yeah, glitch looks like this:

```
extents: 45, 5
extents: 45, 5
extents: 45, 5
extents: 170, 19
extents: 45, 5
extents: 45, 5
extents: 45, 5
extents: 45, 5
```

i think i get what's happening. `text_extents` is synchronous with the lua main thread (as is `peek`) - but all unidirectional commands are not. 

so, commands are potentially being evaluated out of order. this is going to reveal itself whenever multiple font sizes / faces are used together.

hypothesis: this is only going to show up as a problem if there are commands in 

`text_extents` is definitely a big headache and it would be sort of nice if it were not relied on so heavily.

ngwese:
yeah - looking at the code i came to a similar conclusion. tambla uses multiple font sizes for sure.

in the case of `text_extents` i was thinking of the same solve, just move the implementation of `text_center` and `text_right` into C and avoid the need to return values back to lua.

catfact:
ok, i can make that change later today. 

for `peek` i think a callback would make sense, and fortunately i doubt many users would be affected by breaking the API for it.

catfact:
ok unfortunately there are a not-insignficiant number of scripts using `text_extents`:

```
~/norns-all-scripts$ ack text_extents
r/util/norns_engine_tester.scd
559:	var screen_text_extents = { |instance, text|
635:		screen.text_extents_ = { |self, text|
636:			screen_text_extents.value(screen_instance, text);

r/test/test_script_1.scd
192:		~screen.move_((enc1_x + ~screen.text_extents_("LEVEL") + 3), enc1_y);

r/test/lib/common.scd
129:		pages[index][0].label_width = ~screen.text_extents_(page[0].label) - 2;
130:		pages[index][1].label_width = ~screen.text_extents_(page[1].label) - 2;
170:	var value_x = enc1_x + ~screen.text_extents_("LEVEL") + 3;
172:	var value_width = ~screen.text_extents_(the_value);
190:	var value_width = ~screen.text_extents_(the_value);
233:	var text_width = ~screen.text_extents_("PREV");
243:	var text_width = ~screen.text_extents_("NEXT");
275:			var value_x = enc1_x + ~screen.text_extents_("LEVEL") + 3;
277:			var value_width = ~screen.text_extents_(the_value);
358:			var value_width = ~screen.text_extents_(the_value);

r/test/screen_test_2.scd
20:var screen_text_extents = screen[\screen_text_extents];
99:	screen_move.(screen_instance, (enc1_x + screen_text_extents.(screen_instance, "LEVEL") + 3), enc1_y);

r/test/screen_test_3.scd
88:	screen.move_((enc1_x + screen.text_extents_("LEVEL") + 3), enc1_y);

arcologies/lib/graphics.lua
411:        self:rect(pie_chart_x + text_x, pie_chart_y + text_y, screen.text_extents(chart.values[i]) + 2, 7, 15)

yggdrasil/lib/graphics.lua
110:  if screen.text_extents(editor:get_title()) < 41 then
154:  if screen.text_extents(editor:get_title()) < 41 then

yggdrasil/lib/Slot.lua
126:  self:set_extents(screen.text_extents(self:to_string()) + view:get_slot_left_padding())

yggdrasil/lib/tracker.lua
787:  graphics:set_message_extents(screen.text_extents(s))

yggdrasil/lib/Field.lua
40:  local extents = screen.text_extents(s)

yggdrasil/lib/functions.lua
122:    local this_extents = screen.text_extents(e)

gridstep/gridstep.lua
3422:    -- local extents = screen.text_extents(track_string)
3440:    -- extents = screen.text_extents(bpm_string)

pitfalls/lib/display.lua
114:      x = 80 - s.text_extents(name)

pitfalls/lib/functions.lua
54:  local size = s.text_extents(text)
57:    size = s.text_extents(text)

flora/lib/flora_pages.lua
176:    cursor_location = cursor_location - screen.text_extents(instructions) - 1
234:          cursor_start_x = screen.text_extents(cursor_start)
236:          cursor_width = screen.text_extents(cursor_end_x)  - cursor_start_x
241:          cursor_start_x = screen.text_extents(cursor_start)
243:          cursor_width = screen.text_extents(cursor_end_x)  - cursor_start_x
248:          cursor_start_x = screen.text_extents(cursor_start)
250:          cursor_width = screen.text_extents(cursor_end_x)  - cursor_start_x

flora/lib/plant.lua
329:      letter_location.x, letter_location.y = screen.text_extents(letters_left_of_cursor)
333:      local clipped_sentence_len = screen.text_extents(clipped_sentence)
335:      local last_letter_len = screen.text_extents(last_letter)

euclidigons/euclidigons.lua
538:	local width = screen.text_extents(label .. value) + 3
548:	local w, h = screen.text_extents(value)

foundry/foundry.lua
338:    local text_width = screen.text_extents(new_text)

loom/loom.lua
1210:        local underline_length = _norns.screen_text_extents(scale_note_names[i]) + 1

qfwfq/qfwfq.lua
153:  local glyph_w = screen.text_extents(glyph)

**Comment by tehn** [comment]
actually this is not a huge number of scripts, i'm not terribly concerned

my assessment:

- a bunch of these are .scd  (irrelevant)
- some of these can simply use text_center or text_right
- most of these are UI flourish, ie underlines or sizing some other geometry to text size
- many are yggdrasil, which is a whole gnarly low-level UI toolkit down to the pixel level

of course we should do a lower-level implementation of text_center and text_right.

options:
- leave text_extends buggy and suggest workarounds (ie, don't use it, or don't switch fonts/sizes if you really want to use it)
- make it a callback
- other?

what are people's thoughts?

**Comment by catfact** [comment]
i'd propose just removing `screen.text_extents` (replace with dummy in lua) instead of leaving it buggy. 

(the buggy version requires the CR locks, which serve no other purpose if we also refactor `peek` to use a callback. the locks do impose a little performance overhead.)

**Comment by tehn** [comment]
sounds good to me.

**Comment by tyleretters** [comment]
i understand but.... 😱

**Comment by tehn** [comment]
@tyleretters can you outline some of the specific techniques you've scripted which require text_extends? just so that i might propose alternative approaches

**Comment by tyleretters** [comment]
all the "spreadsheet cells" are sized based on text_extents. so if you're viewing just midi notes the columns will be small, but if you want to see what samples are loaded it will stretch to show the whole file name.

i'll probably end up writing a string_to_pixels() function since i know what font i'm using and then i'll just have to measure each of the characters that can be used and store that in a lookup table.

**Comment by catfact** [comment]
adding an async version would be no problem if it helps

**Comment by tyleretters** [comment]
@catfact i think it would? what does async mean in this context? (fwiw i am not using multiple font sizes and faces in yggdrasil)

**Comment by tehn** [comment]
@tyleretters async means you'd write a callback ie

```
function len(x) print("the length is "..x) end

text_extends_async("yooooooooooo party people", len)
```

**Comment by catfact** [comment]
ok. will go in that direction and see what happens.

may be possible to make it a pretty transparent change using coroutines:

```
extens = {w=0, h=0}
redraw = function()
    cr = coroutine.create(function()
        -- draw stuff...
        screen.text_extents(function(w, h) 
            extents.w = w; extents.h = h
            coroutine.resume(cr)
        end)
        coroutine.yield()
        -- draw more stuff using `extents`
    end)
    coroutine.resume(cr)
end
```

**Comment by catfact** [comment]
in fact, after thinking about this, i'd go as far as to say that the only way to make these asynchronous routines really usable, would be to wrap them in coroutines within the API.

i'm thinking this particularly in the context of wanting to add more things like `get_current_point()`.

conceptually, i think this is not too difficut, something like this:

- screen module gets internal variable holding a coroutine for redraws. (`redraw_cr`)
- screen module gets internal variables holding async results. (`extents`, `point`, `buffer`?)
- async commands like `screen.text_extents` set the result variables and call `redraw_cr.yield()`
- screen module provides a wrapper for `redraw` functions that just looks more or less like
```
screen.run_redraw = function(redraw_fn)
  redraw_cr = coroutine.create(redraw_fn)
  redraw_cr.resume()
end
```
- script code would look pretty much identical except that screen functions don't return values directly:
```
...
screen.request_text_extents("blabla")
screen.request_current_point()
w, h = screen.text_extents()
x, y = screen.current_point()
...
```

the main obstacle/challenge i see here is that many bits of the system code assume `redraw` is a plain function and assign/call it directly.

to avoid restructuring all the stuff, we would maybe have to have a more complex global `redraw` object, make it callable with a `__.call` metamethod, and Do Evil with the global metatable to override access to it.

--------------

so, it's an open question for me: does doing all this makes sense, or seem worthwhile? i like the "feeling" of having a much more responsive system under heavy drawing commands. 

but adding all this ifrastructure comes with a risk of creating new bugs, edge cases and inefficiencies.

catfact:
^ 

if the verdict is that this all seems too insane, i'd propose this alternative to try:

- keep the screen events queue, but service it from the main thread.
- the main thread will only service a screen event if there are no pending events.
- since there are no longer multiple threads, cairo "getters" can just avoid the screen event queue and nothing has to change on the lua side.

this might at least help with what i see as the main purpose of this PR: making it less likely for heavy use of the screen API to affect the timing of musical events.

tehn:
my inclination is not to architect something complex just to preserve what i see to be edge-necessity functionality for a music computer. i understand the UI desire for extents/etc but treating them as normal callbacks and leaving the resulting script complexity to the script author is totally fine by me. this allows us to focus on the core mission of the ecosystem, and regretfully a small handful of scripts will need some rework (sorry tyler!)

basically, ezra i worry you're burning way too much of your smarts on this :)

tyleretters:
yeah my conversion thing should only be a few hours of work!

question - this will break all the text_center and text_right stuff, correct?

tehn:
text_center and text_right functionality will not change: the extents portion just gets moved to the c layer.

catfact:
ok have finished up these changes and will begin testing.

closing this PR in favor of a new one since the changes are pretty significant

PR #1296, opened 2021-01-11
api docs cleanup

tehn:
standardized `@module` instead of classmod. prepend `lib.` to lib modules. remove cruft.

this still needs some work, but it's an improvement.


tehn:
merging this since it's just docs

dndrks:
woop! all the `lib.`-prepended stuff isn't accessible -- do we need to do a docs reorg as well?

tehn:
huh, will fix. something happened between local and web

PR #1304, opened 2021-01-18
ldoc comment fixes

catfact:
fix a few ldoc warnings: arguments out of order in ldoc comments, extra colons, mixing `@param` with `@tparam`

there are still quite a few warnings thrown by `ldoc` invocation, mostly due to use of `@module`. 

it would be kind of nice to fix all the warnings so that we can actually pay attention to them. (the argument-mismatch warnings are really useful since those are actual documentation bugs.)


tehn:
thanks for the fixes and yes this is on my todo list

head up, i think you have an old version of `ldoc` as the generated html now has links to v5.1 (whereas it was previously v5.2)

but no worries, i'll regen it before the next release

tehn (approved):

PR #1308, opened 2021-01-24
implmements dedicated drawing thread

catfact:
screen drawing commands now have their own event loop, serviced by a dedicated thread.
other threads do not touch the cairo instance, so locks in `screen.c` are now disabled.

for the most part, there is no change to lua or weaver

however, some commands must now be asynchronous:
`peek`, `text_extents` and newly added `current_point`.

these lua functions now have no return value,
and callbacks for their respective results are registered by weaver in `_norns`,
with some defaults defined in `norns.lua`.
(but no further work has been done on the lua API for async results.)

other changes:
- `screen.text_center` and `screen_text_right` are now implemented as separate weaver commands,
  so they execute synchronously.
- `text_extents` gets the full results in addition to width and height

*testing status* :
have done only basic testing of `screen.extents`, with this script:
https://github.com/catfact/norns-tests/blob/master/screen_async_extents.lua

there may be bugs.

----

_PS: please just ignore the spurious softcut change_


catfact:
ugh sorry something went wrong here

PR #1309, opened 2021-01-24
Screen events, attempt 3.5

catfact:
screen drawing commands now have their own event loop, serviced by a dedicated thread.
other threads do not touch the cairo instance, so locks in `screen.c` are now disabled.

for the most part, there is no change to lua or weaver
(besides the trivial change from `screen_` to `screen_event` in lua->C glue functions)

however, some commands must now be asynchronous:
`peek`, `text_extents` and newly added `current_point`.

these lua functions now have no return value,
and callbacks for their respective results are registered by weaver in `_norns`,
with some defaults defined in `norns.lua`.
(but no further work has been done on the lua API for async results.)

other changes:
- `screen.text_center` and `screen_text_right` are now implemented as separate weaver commands,
  so they execute synchronously.
- `text_extents` gets the full results in addition to width and height

*testing status* :
have done only basic testing of `screen.extents`, with this script:
https://github.com/catfact/norns-tests/blob/master/screen_async_extents.lua

there may be bugs.

----

_PS: please just ignore the spurious softcut change_


catfact:
heads up that there is presently some bug in `text_right`

tehn:
i will review the code shortly, and tomorrow dan and i will test this with a large number of scripts. thanks for your work on this!

catfact:
ok. fixed center and right aligment fns.

if you are gonna modify existing scripts, i think you will quickly find that the coroutine approach outlined in op-linked test script is only reasonable way to handle it to avoid a callback hell. 

and it comes with caveats: breaking up the drawing routine via `coroutine.yield()` means that other event handlers could fire before it resumes. so any state that is used for drawing and might change in the meantime should be cached locally or something.

tehn (commented):
reviewed, looks great. testing tomorrow.

tehn:
proposition:

add arg to text functions to specify maximum width in pixels, after which text is truncated. this seems to be a common reason for using extents

dndrks:
quick heads up -- `util.trim_string_to_width` uses text extents: https://github.com/monome/norns/blob/main/lua/lib/util.lua#L79-L91

after a quick search, a few scripts use this util:
- arp index
- cheat codes
- drum room
- takt
- timber

catfact:
wow ok, that function is a performance disaster anyway, so indeed let's think of something else and implement it in C.

`screen.text_bounded(string, width)`?

catfact:
PS pretty trivial, but in case it's helpful here's what i used to clone all community scripts. (for subsequent `ack`/`find`)
https://gist.github.com/catfact/5521e15a66bc39474782f8751319460f

catfact:
added `screen.text_trim(text, width)`. this has a dedicated screen event type.

if desired, `width` could be added as an optional parameter to `screen.text()` instead. 

either way, seems to work. however. a flow like this will no longer work:
```
- take an arbitrary string 
- trim the string to fit in a given width using `trim`
- get the actual width of the trimmed string using `extents`
```

because the trimmed string is never returned to lua.

tehn:
awesome, thank you

i believe to effectively address the existing uses of the text extents util function we'd need to address center and right aligned text as well, which makes me think perhaps having it be an additional param to each function would make sense.

catfact:
ok, well in that case i guess i wil yank the dedicated event type b/c it seems silly to have 6 of them. (regardless of the lua API.) 
 
its hackish but might just go for a magical rule like, negative width == don't trim.

tehn:
i think that makes sense. also magic rule could just be zero?

catfact:
this is stale, and i believe superceded by the bigger screen module rewrite.

tlubke:
> my understanding was wrong - the other change in question doesn't move surface rendering off the main thread. (does it? correct me if i'm wrong @tlubke )

@catfact
You're correct, the surface rendering still happens on the main thread. I'm gonna look into merging this with #1650, do you think you'll get around to rebasing it? Or should I just rebase my branch and do a manual merge with this one?

catfact:
hey @tlubke, thanks! to be honest i don't know when i can get around to rebasing and re-testing this one. if you have time and energy to check it out i'd certainly appreciate it, but no pressure either.

tlubke:
> hey @tlubke, thanks! to be honest i don't know when i can get around to rebasing and re-testing this one. if you have time and energy to check it out i'd certainly appreciate it, but no pressure either.

I'll see if I can get anywhere with it in the next month.

tehn:
combined as #1720

PR #1311, opened 2021-01-25
Osc remote options

catfact:
add option for the single-argument OSC patterns `/remote/[enc|key]/n d`  in addition to the 2-argument patterns `/remote/[enc|key] n d`.

this allows the remote function to be used with limited/inflexible clients like TouchOSC.

supercollider client code looks like this:
```
~norns_addr = NetAddr("norns.local", 10111);

~enc = { arg n, d;
	~norns_addr.sendMsg('/remote/enc', n, d);
};

// other option:
~enc = { arg n, d;
	~norns_addr.sendMsg(("/remote/enc/"++n).asSymbol, d);
};

~enc.value(1, 8);
~enc.value(1, -8);

```

### DIFF
--- a/lua/core/osc.lua
+++ b/lua/core/osc.lua
@@ -78,24 +78,23 @@ local function param_handler(path, args)
   end
 end
 
+
 local function remote_handler(path, args)
   local cmd = string.sub(path, 1,11)
-  if cmd=="/remote/key" then    
-    if #args > 1 then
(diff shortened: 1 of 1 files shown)


catfact:
(oh to be clear, this is just a proposal, to take or leave.)

tehn (approved):

PR #1313, opened 2021-02-10
Crow update 2.1.2

trentgill:
* W/2.0 `ii` commands as `wtape`, `wdel` and `wsyn`
* crow `ii` commands as `crow`
* Disting EX `ii` commands as `disting`
* adds `crow.loadscript(file, is_persistent)` for uploading a local lua script to crow. optional `is_persistent` flag will save to crow's internal flash so it will run without norns attached. Searches in a `crow/` subfolder for local storage, and falls back to `dust/code/`
* fixes situation where clocksource = 'crow' could fail if crow was running a script with a custom input.stream event

### DIFF
--- a/lua/core/crow.lua
+++ b/lua/core/crow.lua
@@ -84,6 +84,30 @@ function crow.connected()
   return norns.crow.dev ~= nil
 end
 
+--- run / upload userscript
+function crow.loadscript(file, is_persistent)
+  local f = util.file_exists(file)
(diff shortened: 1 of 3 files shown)


trentgill:
I've built a small example script for loading the `bowery` scripts directly from norns here.

https://github.com/whimsicalraps/bowering

tehn (approved):

PR #1315, opened 2021-02-10
update mechanism to install beta

tehn:
added `releases.txt` which acts like a catalog

SYSTEM > UPDATE now queries this file (remotely, the live version on github, not the local copy)

normal update is exactly the same, checks against version.

if K1 is held when entering UPDATE, you are presented with two options: stable and beta, select-able with E2.

this allows forced re-update of stable, or installation of beta. theoretically this allows people to switch between both versions.

the backend is the same, ie update.sh. there will be no "reversion" of system-level changes (ie packages, systemctl units, etc), so we need to be aware about that going forward.

will do some additional testing when i get a non-dev unit, but initial tests show all is well.

ps. right now the `releases.txt` just has stable and beta pointing at the same files. we can update what these point to by commiting to `main` whenever, so this adds the benefit of having the system not live-query the github releases API, which had some people accidentally updating early when something wasn't finished.

### DIFF
--- a/lua/core/menu/update.lua
+++ b/lua/core/menu/update.lua
@@ -1,29 +1,28 @@
 local m = {
-  url = '',
+  alt = false,
+  install = 1,
   version = ''
 }
+
+local releases = {}
+local VER = { 1="stable", 2="beta" }
 
 local function checked() print("what??") end
 local function get_update_2() end
 
 local function check_newest()
   print("checking for update")
+  if m.alt then print("stable and beta") end
   norns.system_cmd( [[curl -s \
-      https://api.github.com/repos/monome/norns/releases/latest \
-      | grep "browser_download_url.*" \
-      | cut -d : -f 2,3 \
-      | tr -d \"]],
+      https://raw.githubusercontent.com/monome/norns/main/releases.txt \
+      ]],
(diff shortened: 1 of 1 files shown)


tehn:
reconfirmed this is working on a fresh non-dev unit.

happy to merge this as-is after someone gives it a quick glimpse, just to approve of my methodology.

catfact (approved):
at a glance: it looks ok, and i'm sure it works, but it strikes me as weird to have repeatedly hardcoded "stable" and "beta" as release names when this could easily be in the descriptor at `norns/main/releases.txt`. and on a related note, whey use `match` and `gsub` when the descriptor is already a valid lua chunk?

```
releases = { 
  stable = { version = ..., url = ..., sha = ...},
 beta = { version = ..., url = ..., sha = ...},
}
```

and in update.lua

```
checked - function(result)
   releases = load(result)
   VER = {}
   for k,v in pairs(releases) do table.insert(VER, k) end
   ...
end
```

but this is pretty minor of course and can be cleaned up post facto

tehn:
good suggestions and i like that `releases.txt` format a lot better. will do a cleanup/test prior to merge

PR #1316, opened 2021-02-15
added missing fonts to api documentation

hiqualityh2o:
While trying to insert some special characters into my project i checked this part of the API to find only 14 font faces on norns. Later, the "Foundry"-App was pointed out to me, which made me aware of all the different fonts available.

I am not sure if the API-list was truncated on purpose, if not, here is a proposal for a complete list of fonts.

### DIFF
--- a/doc/modules/screen.html
+++ b/doc/modules/screen.html
@@ -942,6 +942,59 @@
 <p> 12 Roboto Medium Italic
 <p> 13 Roboto Bold Italic
 <p> 14 Roboto Black Italic
+<p> 15 VeraBd
+<p> 16 VeraBI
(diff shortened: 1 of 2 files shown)


tehn (changes requested):
this is an auto-generated file, so what actually needs to be edited is here: https://github.com/monome/norns/blob/main/lua/core/screen.lua#L186

(i see no need for a line in between each font also)

thanks!

hiqualityh2o:
I see. I edited the screen.lua file accordingly, hope this is fine now.

tehn (approved):
perfect, thank you!

PR #1317, opened 2021-02-18
update softcut

catfact:
update softcut submodule


tehn (approved):

tehn:
augh was overenthusiastic, something is not right. will create a proper test scenario.

catfact:
alright I will look more closely

catfact:
i am not seeing any unexpected behavior....

catfact:
pushed another change to reset fade time as well. 

if you had long fade time, it might be confusing in something like the `recdub` study. because changing the rate alone would do this:
- you hear a sweep in buffer playback pitch over the duration of the xfade
- on repeated passes, the playback pitch sounds normal because write- and read-rates are identical (and pre==0)

so it sounds almost like the rate rises, and then decays back to unity.

tehn:
this makes sense, and sounds like what i was witnessing

PR #1323, opened 2021-03-04
update OSC handler signatures

catfact:
this avoids a warning/error on some newer `gcc`s (`-Werror=incompatible-pointer-types`) for functions with `void*` parameters.

### DIFF
--- a/lua/core/menu/home.lua
+++ b/lua/core/menu/home.lua
@@ -61,22 +61,24 @@ m.redraw = function()
   else
     screen.level(1)
     if norns.is_norns then
-      screen.move(0,10)
-      screen.text("BAT " .. norns.battery_percent)
-      screen.move(36,10)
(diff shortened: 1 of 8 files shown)


tehn (approved):

tehn:
@catfact this last merge broke compilation on-device (gcc6)

> ../matron/src/oracle.c:82:18: error: ‘struct lo_message_’ declared inside parameter list will not be visible outside of this definition or declaration [-Werror] struct lo_message_ *data, void *user_data);

causes a cascade of other errors.

i'm trying to figure it out but this technique is pretty opaque

tehn:
of course

s/struct lo_message_/void/g

fixes it, but that basically undoes your fix for later gcc's

catfact:
Ok I see, thanks, sorry. I think there is a fix for both

Issue #1325, opened 2021-03-12
callback for saving/loading parameters


schollz:
I'd like to be able to run arbitrary code after `ParamSet:read` or `ParamSet:write`, whether it is invoked via the `PSET` menu, or whether it is invoked manually. 

i have several use-cases that this would be useful:

- after reading a `PSET` I'd like to reset certain parameters
- after reading a `PSET` I'd like to load softcut buffers corresponding to that save
- after writing a `PSET` I'd like to dump softcut buffers corresponding to that save

i'm pretty sure it would be useful for others too. a common workaround pattern is to avoid `PSET` by making [custom `save/load` menus](https://github.com/schollz/barcode/blob/master/barcode.lua#L59-L72) (I do this a lot) in order to customize the saving/loading process. another pattern i've seen is to [rewrite the pset.write/read function](https://github.com/tlubke/shapes/blob/midi-cc/shapes.lua#L605-L627) to inject a callback.

what I'm thinking would be most useful is two new `ParamSet` class variables that contain callback functions that can be set using `ParamSet:set_callback_write` and `ParamSet:set_callback_read`. then at the end of `ParamSet:read/write` those functions would run if not nil.

another option would be to directly add callbacks to the functions, e.g. `ParamSet:read(filename, name, callback_fn)`...but this is not ideal for my use-case when someone saves/loads via the `PSET` menu (whereas the former option does).

is this appropriate for a PR? or is there is a way to already do this?

tehn:
i think your first suggestion makes sense (and as you say, the second suggestion accomplishes nothing new).

looking forward to a pr!

catfact:
nice

PR #1326, opened 2021-03-14
add callbacks for reading/writing parameters

schollz:
this PR addresses https://github.com/monome/norns/issues/1325

test code:

```lua
-- test
-- PSET read/write callbacks
-- press k2/k3 to write/read
-- try write/read with PSET menu

function init()
	params:set_callback_read(function(filename)
		print("finished reading '"..filename.."'")
	end)
	params:set_callback_write(function(filename,name)
		print("finished writing '"..filename.."'")
	end)
end

function key(k,z)
	if z==1 and k==2 then
		params:write("/tmp/parameters")
	elseif z==1 and k==3 then
		params:read("/tmp/parameters")
	end
end
```

i'm not attached to the api names, lmk if i can improve it.

### DIFF
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -453,6 +453,8 @@ function ParamSet:clear()
   self.name = ""
   self.params = {}
   self.count = 0
+  self.action_read = nil 
+  self.action_write = nil
 end
 
 return ParamSet
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -40,6 +40,8 @@ function ParamSet.new(id, name)
   ps.hidden = {}
   ps.lookup = {}
   ps.group = 0
+  ps.callback_write = nil
+  ps.callback_read = nil
   ParamSet.sets[ps.id] = ps
   return ps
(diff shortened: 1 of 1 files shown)


tehn:
we've avoided using the word "callback" in favor of "action" so it'd be best to continue that.

also do we need setter functions? maybe we just set `param.action_read` directly?

schollz:
yeah, that's way more consistent actually.

here's the new test code:

```lua
-- test
-- PSET read/write callbacks
-- press k2/k3 to write/read
-- try write/read with PSET menu

function init()
    params.action_read=function(filename)
        print("finished reading '"..filename.."'")
    end
    params.action_write=function(filename,name)
        print("finished writing '"..filename.."'")
    end
end

function key(k,z)
    if z==1 and k==2 then
        params:write("/tmp/parameters")
    elseif z==1 and k==3 then
        params:read("/tmp/parameters")
    end
end
```

also i forgot to clear the actions so the function was persisting between loads. e5d6af5 fixes that.

tehn (approved):

PR #1327, opened 2021-03-14
fix lua api links and update references to 'main' branch

jayzmiles:
I noticed a couple of broken links, and ran into a couple of issues on the contributing page in the process of trying to submit fixes.

One other bump I ran into:
Using `git clone` from the terminal seems to default to https (rather than ssh) so I needed to change the remote URLs - this step isn't accounted for in `CONTRIBUTING.md`. [GitHub docs here](https://docs.github.com/en/github/using-git/changing-a-remotes-url#switching-remote-urls-from-https-to-ssh)

Happy to add this step to the contributing doc if this is still the most straightforward path

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 contributing
 ==================================
 
-[![Build Status](https://travis-ci.org/monome/norns.svg?branch=master)](https://travis-ci.org/monome/norns)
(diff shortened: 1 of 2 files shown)


tehn (approved):
thank you, and i apologize these got so stale.

jayzmiles:
No worries and thank you

PR #1330, opened 2021-03-20
FIX parser for multiple lines across multiple packets

trentgill:
Fixes `_norns.crow.event` parser for the situation where crow sends a bunch of text that gets split over 2 separate usb packets. Previously any line over the packet limit would cause a norns event error, then the end of the string would be printed to console.

Now we maintain a buffer for any lines that are not terminated by line-endings.


tehn (approved):
brilliant!! much better.

PR #1331, opened 2021-03-28
fileselect: treat symlinks to directories like directories

csboling:
With `-L`, the `ls` command will treat symlinks to directories the same as "real" directories. Previously they did not get listed with the other directories and did not end in `/`, so the menu code would try to open them like files and error. This allows the user to e.g. `ln -s /media/usb /home/we/dust/audio/usb` and then play tape files / load samples from a USB disk.


tehn (approved):
who you totally read my mind! thank you for testing this and making it work.

PR #1332, opened 2021-03-28
add restart, work text for reset

tehn:
added SYSTEM > RESTART which simply restarts the software stack

SYSTEM > RESET now provides more text "clear settings and restart?" which is the same as before

likely a bunch of docs will need to be changed from RESET to RESTART (though, both will effectively fix whatever problem, just the latter more gracefully).

secondarily, re docs: SYSTEM will v3 be renamed to SETTINGS


ngwese:
yesterday got away from me. will check out tonight.

ngwese (approved):
lgtm - tested on device

PR #1335, opened 2021-03-31
Hard reset w modified divisions

kasperbauer:
ping @tyleretters 

When doing a `hard_restart` on lattice, the `phase` has to be recalculated instead of taken from the `phase_end` property.
This way, multiple patterns stay in sync, after divisions had been changed.

`phase_end` can be removed as it's no longer used.

Also, a dedicated reset method was added to provide a handy way to start from t = 0, after changing divisions.

(coming from: kasperbauer/metrix#20)


tyleretters:
this looks awesome. thank you so much for this improvement.

tested and works as expected. good to merge @tehn !

PR #1347, opened 2021-04-16
single threaded clock fixes

artfwo:


tehn:
this is feeling good! i'd be good to merge this and open an issue for the outstanding quirks.

artfwo:
there's still an issue with rescheduling that i hope to fix in the next few days and transport behavior is inconsistent across sources, i can work on those issues in this branch or in separate branches if you want, no problem.

tehn (approved):

PR #1348, opened 2021-04-19
softcut feature: query position

tehn:
softcut voice now stores position per processing block, and i've added a query function to get the (accurate to block size) position of a voice.

i wanted to add this a different method to async check positions without throwing away a phase poll--- ie, being able to save a playback position "now" via a keypress

```
-- create simple print report callback
cb = function(idx,pos) print(idx,pos) end

-- arg is callback function
softcut.event_position(cb)

-- arg is which voice
softcut.query_position(1)
```

plumbing, wow.

### DIFF
--- a/crone/src/OscInterface.cpp
+++ b/crone/src/OscInterface.cpp
@@ -782,11 +782,11 @@ void OscInterface::addServerMethods() {
                                      });
     });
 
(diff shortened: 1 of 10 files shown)


tehn:
@catfact tested on device, working great

catfact (approved):
looks ok to me.

(yeah it really is a ton of crud to add a new event, isn't it... sorrry)

PR #1350, opened 2021-04-20
Fix metro running state

catfact:
realized that `metro.is_running` variable is dead/useless for non-infinite metros when someone tried to use it in a script. (actually a crow script which uses the same lua code, fwiw)

this change causes that state to be updated after firing a metro's event callback, if the stage count exceeds the metro's count. 

i think this is the behavior we want, but haven't actually tried it so not totally sure.

of course we could also just remove that field.

### DIFF
--- a/lua/core/metro.lua
+++ b/lua/core/metro.lua
@@ -156,17 +156,17 @@ for i=1,Metro.num_script_metros do
   Metro.assigned[i] = false
 end
 
-
-
 -- callback on metro tick from C.
 _norns.metro = function(idx, stage)
   local m = Metro.metros[idx]
   if m then
-    if Metro.metros[idx].event then
-      Metro.metros[idx].event(stage)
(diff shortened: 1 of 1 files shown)


tehn (commented):
i'm wondering if this will fix the spurious "metro does not exist" errors, which dan and i have been trying to track down (mainly happen in the menu code, metro 31 and 35)

catfact:
i doubt it, this lua vairable is not used anywhere in our code, which is why we didn't know it was broken

i've written all i can say about that issue elsewhere. it needs a deep dive into metro.c with gdb. even my medium-shallow dives have not revealed the cause. the C code should not allow that condition to be reached but somehow it is happening.

tehn:
thanks for the refresher--- it's been awhile since i thought about this

tehn (approved):

PR #1352, opened 2021-04-20
clock sync chaining improvements

artfwo:
remove automatic sync events rescheduling logic in favor of explicitly rescheduling sync events to avoid the beat drifting away from hi-res sync-based counters.


tehn (approved):

tehn:
code looks good. we'll keep on testing!

PR #1357, opened 2021-04-28
TAPE records into nested folders + add util.file_size

dndrks:
fulfills a feature request from the norns study group for improved default filename and location.

### new: util.file_size
- added a `util` function which returns a file's size in megabytes:
```lua
> util.file_size(_path.code.."awake/awake.lua")
0.013152
> util.file_size(_path.audio.."tape/0000.wav")
71.710508
```
- if provided filepath isn't found:
```lua
> util.file_size(_path.audio.."tape/blahblah.wav")
no file found at /home/we/dust/audio/tape/blahblah.wav
<ok>
```
- if no path is provided:
```lua
> util.file_size()
util.file_size requires a path
<ok>
```

### TAPE changes
- when a user initiates `REC`, their recording is nested into a folder for the loaded script's `shortname`, eg. `tape/awake`
  - if no script is loaded, recording is nested into `tape/no_script`
- the default filename appends the current clock bpm to the `index`, as that is often forgotten when a user revisits their recordings, eg. `0109_120bpm`
  - if bpm is float, eg. 120.85: `0109_120-85bpm`
- if the user queues up a recording but changes their mind before pressing `START`, they can now hold K1 to call up a `CLEAR` action, which is selectable with K3
  - `CLEAR` will validate that the `0109_120bpm.wav` file has no data using `util.filesize` (rounding to nearest KB)
    - if it has no data, then the placeholder file will be deleted
    - if the placeholder file is deleted, the index (and `index.txt`) is reset to the prior state

by frontloading the index, we retain a sense of timeline across recordings, but by placing everything in named folders then a user can easily export all of their `compass` recordings in one move :)

### DIFF
--- a/lua/core/menu/tape.lua
+++ b/lua/core/menu/tape.lua
@@ -13,6 +13,10 @@ local TAPE_REC_ARM = 1
 local TAPE_REC_START = 2
 local TAPE_REC_STOP = 3
 
+local TAPE_KEY1_HELD = false
+local TAPE_QUEUED_FILE;
+local TAPE_QUEUED_INDEX;
+
 local m = {
   mode = TAPE_MODE_PLAY,
   play = {
@@ -45,7 +49,9 @@ local function tape_exists(index)
   if type(index) == "number" then
     index = string.format("%04d",index)
   end
-  local filename = _path.audio.."tape/"..index..".wav"
+  local current_tempo = string.format("%04f",clock.get_tempo()):gsub("%.?0+$", "")
+  local script_name = norns.state.shortname ~= "none" and norns.state.shortname or "no_script"
+  local filename = _path.audio.."tape/"..script_name.."/"..index.. "_" ..string.gsub(current_tempo,"%.","-") .. "bpm.wav"
   return util.file_exists(filename)
 end
 
@@ -62,6 +68,8 @@ local function update_tape_index()
     m.fileindex = m.fileindex+1
   end
 
+  TAPE_QUEUED_INDEX = m.fileindex
+
   local f = io.open(_path.tape..'index.txt','w')
   if f == nil then
     os.execute("mkdir -p ".._path.tape)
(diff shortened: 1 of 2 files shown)


tehn:
# util.file_size

perfect, looks great. thank you! (maybe this should've been a separate PR?)

# tape

super cool that you're digging into this: let's treat this as a springboard to visit the TAPE UI more broadly (which really wants an overhaul, been on the list for some time now.)

first, while i see the desire for more metadata, i think this implementation is a very specific personal preference: subfolders and bpm filenames. for myself i'm interested in neither (i'd rather work with a single folder chronologically, like a real physical "tape"--- and i don't really use bpm's, let alone consistently in the same recording, so it actually ends up being false/irrelevant metadata.) given i suspect there are others who may have the same inclination, let me counter-propose:

- we go forward with adding some system-wide "preferences" (this has come up in the past for other features, but sadly i cannot recall what they were) which i think could just get stored in `system.state` and added to the SYSTEM (should be renamed SETTINGS, as we planned for 3.0... perhaps we just do this soon and conquer the doc updates). so, SETTINGS would change to a scrolling list. the top items (WIFI, RESET, RESTART, UPDATE) could remain, and perhaps we move sub-DEVICES (MIDI, GRID, etc) up a level. TAPE could come after this, and include a few different options/actions:

SETTINGS >

```
WIFI
UPDATE
RESTART
RESET
---
MIDI
GRID
ARC
HID
TAPE
```

basically this allows us to add more "settings" down the line if needed. also it opens the devices submenus for additional options (ie, grid rotation, or keyboard layout, or midi passthru)

holy sidetrack, here's my TAPE config proposals:
- append script name (ie `-awake`) (on/off)
- append bpm (ie -166-66bpm) (on/off)
- generative name (on/off, see below)
- ERASE (fileselect single tape file, confirm)
- CLEAR ALL TAPE (erase all tape files, confirm)
- RESET COUNTER (not sure this is necessary, but i guess to be thorough)

i had a half-finished "random name generator" for tapes, which referenced a user-dictionary text file for word components. some people may prefer this sort of thing for mental cues of particular recordings.

with all of the naming, at ARM you should be able to change the default name completely, but the above just set up the default pattern

---

this doesn't actually get to the _actual_ TAPE UI (the tape screen itself) of course.

---

apologies for the somewhat misplaced brain dump! feedback welcome.

andr-ew:
could consider doing some kind of path name config file for ultimate flexibility, something like:
`return scriptname .. '/' .. index .. '_' .. bpm .. '.wav'`

tehn:
i think people will want to config via menu, not text.

also, config option for subfolder (like @dndrks original pr)

catfact:
+1 to both of the points made by @tehn:
- nested folders would hinder usability for me - its simple to batch-manage a single dir and use file creation timestamps. and the assumption that BPM is a meaningful descriptor is a little off-putting.
- suggesting a customization path involving lua coding is beside the point. if you are a programmer you can already use the `audio.tape_*` functions directly.

dndrks:
for sure -- apologies for not closing the loop here. allowing these options via customization, default being off, seems like a really nice way to ensure the current experience persists for those who prefer it.

fwiw, this wasn't based on personal preference, just an attempt to meet requirements from direct user feedback -- i think folks are up for anything that allows more meaningful identifiers to be added more quickly. random naming is a really nice way to do that. keyboard support for this menu could also work. or even a different on-screen layout for naming that allows quicker navigation of the alphabet.

all to say, i think what brian laid out is a really nice middleground :)

tehn:
closing this, let's open an FR issue and revisit

PR #1358, opened 2021-05-02
grid-st

catfact:
support for new grids based on stm32.

-----

this requires that `matron` be linked against `libmonome` built from the `grid-st` branch.

steps i had to take:

- clone `libmonome` and build the testing branch
```
git clone https://github.com/monome/libmonome
cd libmonome
git checkout grid-st
./waf configure && ./waf && sudo ./waf install
```

this will result in `libmonome` being installed at `usr/local/include`, `/usr/local/lib`.

however the norns image also has libmonome installed directly under `/usr`. (i'm not sure how.)

- so i had to change the runtime link behavior:

```
echo '/usr/local/lib' | sudo tee /etc/ld.so.conf.d/libmonome.conf
sudo ldconfig
```

(i also nuked the installation under `/usr` but probably not necessary.)

### DIFF
--- a/matron/src/device/device.c
+++ b/matron/src/device/device.c
@@ -71,8 +71,6 @@ err_init:
 
 void dev_delete(union dev *d) {
     int ret;
-    // fprintf(stderr, "dev_delete(): removing device %d\n", d->base.id);
-
     if (pthread_kill(d->base.tid, 0) == 0) {
         // device i/o thread still running
         ret = pthread_cancel(d->base.tid);
--- a/matron/src/device/device_common.h
+++ b/matron/src/device/device_common.h
@@ -2,6 +2,8 @@
 
 #include <stdint.h>
 
+
+
 typedef enum {
     // libmonome devices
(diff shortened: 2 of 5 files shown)


catfact:
as of hash `ff77fdb8af4eb7ff5f92f91cf197abf146adb1ec`, seems to be working. 
(was having PEBCAK trouble due to once again forgetting how vports work.)

tehn:
code looks great. will test on hardware when i return!

ngwese:
will give this a test tonight

ngwese:
confirmed working on norns, multiple generations of grids, switching back and forth, hot plug.

dmesg output:

ngwese (approved):
fwiw i couldn't find a `grid-st` branch in `monome/libmonome`  or a related pull request so i built `grid-st` from `catfact/libmonome`.

otherwise lgtm.

tehn:
merging this. works without updating libmonome.

PR #1364, opened 2021-05-11
add setters to control UI lib's 'active' state

dndrks:
while going through the `UI` lib for ref docs with @fardles, I noticed that directly modifying the `.active` variable was a pretty common action to change focus between dials or dim a specific component. so, rather than document adjustments to a variable, I thought it might be good to add these adjustments as supported functions to the API. I also added a `set_status` for `PlaybackIcon`, as modifying `.status` seemed like a common action.

tested on device, example test script:

```lua
UI = require("ui")

play_icon = UI.PlaybackIcon.new(10,10,30,1)

function redraw()
  screen.clear()
  play_icon:redraw()
  screen.update()
end

function key(n,z)
  if n == 3 and z == 1 then
    play_icon:set_active(not play_icon.active)
    if play_icon.status == 1 then
      play_icon:set_status(3)
    elseif play_icon.status == 3 then
      play_icon:set_status(1)
    end
    redraw()
  end
end
```

### DIFF
--- a/lua/lib/ui.lua
+++ b/lua/lib/ui.lua
@@ -169,6 +169,12 @@ function UI.List:set_index_delta(delta, wrap)
   self:set_index(index)
 end
 
+--- Set selected list's active state.
+-- @tparam boolean state Boolean, true for active.
+function UI.List:set_active(state)
+  self.active = state
+end
+
 --- Redraw List.
 -- Call when changed.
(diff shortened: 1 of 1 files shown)


markwheeler (approved):
Makes sense to me 👍

tehn (approved):

PR #1365, opened 2021-05-12
use jack system time for clock

artfwo:


dndrks:
daaaaaaaaaang, really nice work:

image

close enough for ambient-electro-jazz!

tehn:
this is looking great! happy to merge when you're ready.

artfwo:
i think this is ready for review now, but it would also be cool to give it a few smoke tests before merging. 

unfortunately, keeping a jack client in matron might make the system/menu unusable in case jackd freezes and requires restarting (happens every so often on my device). matron might need to reinitialize clock client in this case at best (and crash at worst), so i would like to eventually isolate the jack client and add some error handling to gracefully handle those cases.

artfwo:
and here are some quick crude performance metrics with a basic counter script scheduling a coroutine wakeup 100 times per beat. i've compared release, main and jack builds, the metric is CPU usage in percents.

the fastest implementation so far is current main branch, with JACK clock having a small performance impact as expected, and current stable release being the slowest, reaching as far as 8% CPU on my box.

norns_clock_cpu

dndrks:
closing the loop -- i stress tested this branch for a few hours on some heavy coroutine traffic, external syncing (to/from MIDI as well as crow), and transport start/stop and things seem to be really really clean. i didn't notice any performance issues.

unless there are any other hesitations, i feel like this is super work, artem!!!

tehn (approved):

PR #1369, opened 2021-05-26
add 'connected' state to midi vports

dndrks:
spun this up while working on a MIDI study, when I realized that it was a little wonky for authors to check whether a vport-stored MIDI device is _currently_ physically connected to norns. a niche gap, but one that's easy to fill :)

without this helper flag, connected state is only reflected by the presence of a `device` table in the MIDI vport's table or as a cross-reference to `midi.devices`. by piggybacking a `Midi.update_device_connected_state()` function to the end of `Midi.update_devices()`, the `midi.vports[x].device_connected` flag updates any time a MIDI device is connected or removed.

### DIFF
--- a/lua/core/midi.lua
+++ b/lua/core/midi.lua
@@ -13,7 +13,7 @@ for i=1,16 do
   Midi.vports[i] = {
     name = "none",
     device = nil,
-    device_connected = false,
+    connected = false,
     event = nil,
 
     send = function(self, ...) if self.device then self.device:send(...) end end,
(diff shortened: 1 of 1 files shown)


tehn:
i'd be up for simplifying this and just calling it `connected` if that makes sense to you?

dndrks:
@tehn , for sure! i just couldn't tell since it reflects the status of `midi.vports[x].device` if it needed to stick close to that naming, but `midi.vports[x].connected` should totally be legible

tehn (approved):
this is great, thank you!

PR #1370, opened 2021-05-27
fix start/stop sync in link clock source

### LINKED ISSUE #1368: clock beta: Ableton Link transport reset issue
the beta testing group on Discord has started diving in and found an issue with Ableton Link syncing -- happens in Live as well as Bitwig. this was not an issue with the previous public release.

steps to repro:
- get norns and Live/Bitwig on the same network and enable Link on both
- in Live/Bitwig, enable `Start Stop Sync` (issue only happens when this is enabled)
- load Awake
- toggle the transport on for your Live/Bitwig set
- Awake will ramp up its tempo for a few seconds (very very high), then settle back to the expected tempo

during this, the `tempo` under parameters doesn't reflect any weirdness -- it'll just keep displaying the same bpm -- but there's an audible rush of notes, which is also reflected in the awake UI's playhead.

artfwo:
fixes #1368

### DIFF
--- a/matron/src/clocks/clock_link.c
+++ b/matron/src/clocks/clock_link.c
@@ -36,7 +36,7 @@ static void *clock_link_run(void *p) {
 
     while (true) {
         if (pthread_mutex_trylock(&clock_link_shared_data.lock) == 0) {
-            state = ableton_link_capture_audio_session_state(link);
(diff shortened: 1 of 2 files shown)


tehn (approved):
code looks great, thank you! i'll merge this and build a new release for the testing crew.

PR #1371, opened 2021-05-28
return beat from clock.sync

artfwo:
this PR changes the clock API to return current beat from `clock.sync` calls as follows:

```
clock.run(function()
    local div = 1/4

    for i=1,4 do
        local beat = clock.sync(div)
        local step = math.floor(beat / div)
        print('beat', beat, 'get_beats', clock.get_beats(), 'step', step)
    end
end)
```

so the above code would print

```
beat	0.25134		get_beats	0.25143		step	1
beat	0.50106		get_beats	0.50106		step	2
beat	0.75134		get_beats	0.75137		step	3
beat	1.00103		get_beats	1.00109		step	4
```

this basically allows tracking time in beat-synced coroutines without external counters and additional transport restart handlers, as current beat will be always returned from `clock.sync` and it will be naturally synced to the source transport too.

with this change `clock.sleep` calls will also return elapsed system (currently jack) time.

`clock.sync` called with any value less than or equal to zero will result in an error.

`clock.sleep` called with any value less than or equal to zero will schedule the coroutine to resume immediately on next scheduler tick.

### DIFF
--- a/matron/src/clocks/clock_scheduler.c
+++ b/matron/src/clocks/clock_scheduler.c
@@ -32,9 +32,10 @@ static pthread_t clock_scheduler_tick_thread;
 static clock_scheduler_event_t clock_scheduler_events[NUM_CLOCK_SCHEDULER_EVENTS];
 static pthread_mutex_t clock_scheduler_events_lock;
 
-static void clock_scheduler_post_clock_resume_event(int thread_id) {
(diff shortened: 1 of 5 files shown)


tehn (approved):
this is great!

tehn:
@dndrks we need to get this improvement/feature into the docs properly

dndrks:
for sure -- the built in counter a really lovely feature + opportunity !

PR #1373, opened 2021-06-01
do not assume unmatched tty is crow

tehn:
basically just check the product string before sending on the tty device to the crow ^^identity check, to avoid sending unexpected bytes to a random tty device

also re-tabbed. sorry for the mess, this file was a crazy mix.

### DIFF
--- a/matron/src/device/device_monitor.c
+++ b/matron/src/device/device_monitor.c
@@ -31,25 +31,19 @@
 
 // enumerate unix files to watch
 enum {
-  DEV_FILE_TTY   = 0,
-  DEV_FILE_INPUT = 1,
-  DEV_FILE_SOUND = 2,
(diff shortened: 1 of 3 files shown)


matron/src/device/device_monitor.c
```diff
@@ -213,121 +206,123 @@ void *watch_loop(void *p) {
...
 #endif
-	 dev_list_add(DEV_TYPE_MONOME, node, get_device_name(dev));
-	 return;
-     }
-     
+        dev_list_add(DEV_TYPE_MONOME, node, get_device_name(dev));
+        return;
+    }
+
+    if (!strcmp(device_product_string, "crow:_telephone_line")) {
```
artfwo, line 303: :phone:

artfwo (approved):

PR #1377, opened 2021-06-05
arc quads fix

tehn:
sorry i didn't test this before earlier merge. works now.


tehn:
augh bad commits, redoing

PR #1380, opened 2021-06-08
show crow as an option for the set_source function

jaseknighter:


tehn:
thanks for the PR!

the html is autogenerated by `ldoc` based on the inline documentation, so the appropriate fix would be to edit `clock.lua` at the function's header doc

PR #1381, opened 2021-06-09
device monitor fixes

catfact:
- check for crow using serial string on removal as well as addition
- don't process non-physical `/dev/input` entries (suppresses some error messages at startup)
- clean up debug prints

(i thought there were memleaks too, but apparently not)

PR #1382, opened 2021-06-09
Device monitor fixfix

catfact:
fixes broken midi from pr #1381. (by moving check for null device nodes to the specific add routine for `/dev/input`.)

there are still spurious and apparently benign errors when adding (some) HID devices. not going to worry about that right now, but it would be good to hunt those down, at least insofar as making sure the failed attempts aren't leaking memory.


dndrks:
raaaad, tests well with all my devices -- appreciate the `device_monitor(): adding midi device UM-ONE` messages, as well! feels clean, thank you! :)

dndrks (approved):

PR #1383, opened 2021-06-18
add offset (swing) parameter to clock.sync

artfwo:
this change enables various swing styles in coroutines using clock.sync by adding an optional `offset` parameter for `sync`, which can be used as follows:

```lua
clock.sync(1) -- sync and resume as usual
clock.sync(1, 0.5) -- sync and resume with a half-beat delay
clock.sync(4, -2) -- sync and resume 2 beats earlier

-- create a swinging feel for the whole script
local beat = 1/4
local groove = 0

while true do
  local current_beat = clock.sync(beat, groove)
  local current_step = math.floor(current_beat / beat)
  if current_step % 2 == 0 then
    groove = beat / 2
  else
    groove = 0
  end
end

-- same as above, using XOR (~) for setting even/odd offset multiplier
local beat = 1/4
local offset = beat / 2
local swing = 0

while true do
  clock.sync(beat, offset * swing)
  swing = swing ~ 1
end

```

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -68,7 +68,7 @@ end
 clock.resume = function(coro_id, ...)
   local coro = clock.threads[coro_id]
 
-  local result, mode, time = coroutine.resume(coro, ...)
+  local result, mode, time, offset = coroutine.resume(coro, ...)
 
   if coroutine.status(coro) == "dead" then
     if result then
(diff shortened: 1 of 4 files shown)


tehn (approved):
this is a nice way to add swing--- hadn't thought of this approach before. thank you!

Issue #1384, opened 2021-06-18
proposal: modules


tehn:
aka mods (modifications?)

basically, libs that are user-selectable to be "always loaded" at script init

existing cases:
- `passthrough` midi util
- `n160`
- more grid options ie rotation and brightness

rather than extend the menu system more and more, i'd like to propose a simple system whereupon libraries can arbitrarily add functionality (with UI in the param menu) or redefine existing behavior (in a non-breaking way... such as adding extra midi functions).

this is pretty easy to do by modifying `norns/lua/core/script.lua`

```lua
Script.init = function()
  print("# script init")
  params.name = norns.state.shortname
  -- lines below added to always load this specific lib:
  if util.file_exists(_path.code.."passthrough") then
    local _passthrough = include 'passthrough/lib/passthrough'
    _passthrough.init()
  end
  -- continue
  init()
  _norns.screen_save()
end
```

rather than a single file, this section could call a second file which has a list of activated modules/libs. editing this file (say in `dust/data/system.modules` or something) could change what gets loaded

but, in the marketplace/gallery style of community collaboration, it'd probably be preferable to have a menu UI for scanning all available mods, with a list of enable/disable.

potential issues:

- namespace collisions (we need some conventions)
- init behavior
- pre-existing libs: ie, a script that already `includes` or `requires` a mod need not be re-loaded
- general complexity
- param menu overload. (this is in fact not really a issue to me, as this stuff should all be optional. what would be overload is a bigger and bigger menu/param system by default.)

all comments and collaborators welcome.

nattog:
I think its an elegant solution to handling a growing ecosystem of libraries. Some initial questions and thoughts:

- I think this functionality would benefit from some form of menu UI to be accessible and reduce complexity in use/user error. 
- In which case, how to distinguish between modules and scripts? Is some form of flag or tag needed? 
- Could this be controllable on individual scripts or applied for all scripts?
- Does it require some degree of conformity/shared standards across libs?
- Taking Passthrough as an example, it can interact with a script by defining callback functions at script level called from MIDI events handled within Passthrough. Will this proposal permit this behaviour - presumably there won't be any scoping issues?

jeffbr13:
One consideration for symmetry with initialisation: module **teardowns/cleanup**.

[...]

PR #1385, opened 2021-06-19
matron lifecycle hooks and system level mods.lua

ngwese:
this is a proof of concept which implements much of the functionality i proposed for #1384

this does not provide any ui or dynamic discovery of "mods". for demonstration purposes the `startup.lua` code was extended to look for `dust/data/mods.lua` and load that file at the end of matron's startup process. within `mods.lua` one can place code or call `require '<script_dir>/lib/<blah>'` to load files from the normal script/project area.

### DIFF
--- a/lua/core/config.lua
+++ b/lua/core/config.lua
@@ -27,4 +27,5 @@ _path.audio = _path.dust..'audio/'
 _path.tape = _path.audio..'tape/'
 _path.data = _path.dust..'data/'
 _path.favorites = _path.data..'system.favorites'
+_path.enabled_mods = _path.data..'system.mods'
 
--- a/lua/core/hook.lua
+++ b/lua/core/hook.lua
@@ -34,6 +34,7 @@ local hooks = {
   script_pre_init = Hook.new(),
   script_post_cleanup = Hook.new(),
   system_post_startup = Hook.new(),
+  system_pre_shutdown = Hook.new(),
 }
 
 return tab.readonly{table = hooks}
--- a/lua/core/hook.lua
+++ b/lua/core/hook.lua
@@ -1,0 +1,39 @@
+local tab = require 'tabutil'
+
+local Hook = {}
+Hook.__index = Hook
+
+function Hook.new()
+  local this = {}
+  this.callbacks = {}
+  return setmetatable(this, Hook)
+end
+
+function Hook:register(name, func)
+  self.callbacks[name] = func
+end
+
+function Hook:deregister(name)
+  self.callbacks[name] = nil
+end
+
+function Hook:__call(...)
+  for k, func in pairs(self.callbacks) do
+    if func ~= nil then
+      print('calling: ' .. k)
+      local ok, error = pcall(func, arg)
+      if not ok then
+        print('hook: ' .. k .. ' failed, error: ' .. error)
+      end
+    end
+  end
+end
+
+-- define the hook types
+local hooks = {
+  script_pre_init = Hook.new(),
+  script_post_cleanup = Hook.new(),
+  system_post_startup = Hook.new(),
+}
+
+return tab.readonly{table = hooks}
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -254,4 +254,5 @@ m["UPDATE"] = require 'core/menu/update'
 m["SLEEP"] = require 'core/menu/sleep'
 m["MIX"] = require 'core/menu/mix'
 m["TAPE"] = require 'core/menu/tape'
+m["MODS"] = require 'core/menu/mods'
 
--- a/lua/core/menu/mods.lua
+++ b/lua/core/menu/mods.lua
@@ -1,4 +1,5 @@
 local mods = require 'core/mods'
+local tabutil = require 'tabutil'
 
 local m = {
   pos = 0,
@@ -6,12 +7,15 @@ local m = {
   selected = ""
 }
 
+m.select_position = function(pos)
+  m.selected = string.upper(m.list[pos+1])
+end
+
 m.key = function(n,z)
   -- back
   if n==2 and z==1 then
     _menu.set_page("SYSTEM")
   elseif n==3 and z==1 and m.len > 0 then -- if there are mods
-    -- TODO: check if mod is enabled first!
     if _menu.m[m.selected] then -- does the mod have a menu page
       _menu.set_page(m.selected)
     end
@@ -21,15 +25,17 @@ end
 m.enc = function(n,delta)
   if n==2 then
     m.pos = util.clamp(m.pos + delta, 0, m.len - 1)
-    m.selected = string.upper(m.list[m.pos+1])
+    m.select_position(m.pos)
     _menu.redraw()
   elseif n==3 then
-    -- TODO
-    if d > 0 then
-      -- ENABLE MOD
+    if delta > 0 then
+      -- print("enable", m.selected)
+      m.set_enabled(true)
     else
-      -- DISABLE MOD
+      -- print("disable", m.selected)
+      m.set_enabled(false)
     end
+    _menu.redraw()
   end
 end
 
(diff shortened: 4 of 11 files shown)


tehn:
looking forward to checking this out tomorrow!

tehn:
for ease of UI management, i'm wondering if `dust/data/mods.lua` should simply be a table of enabled mod paths ie

```
{
  "dust/code/passthru/lib/passthru.lua",
  "dust/code/example/lib/example.lua",
}
```
and then the mods loader in `startup.lua` could just process this table sequentially with `require`

basically i'm considering how to structure the UI "selector" which will be just tables, so exporting these tables directly might be more sane than generating proper lua syntax with `require` and quotes.

tehn:
also--- do you have an example mod which shows an standard way of implementing hooks?

(note, previous comment edited)

ngwese:
sorry, should have included my test setup.

the `dust/data/mods.lua` file looks like this:
```lua
require 'MOD/lib/mod'
```

and then `dust/code/MOD/lib/mod.lua` looks like this at the moment:
```lua
local hook = require 'core/hook'

local state = {}
local api = {}

api.get_state = function()
  return state
end

-- dummy functionality to confirm things are getting called and test out access
-- to mod level state via mod supplied fuctions.

hook.script_pre_init:register('MOD-pre-init', function()
  state.script_pre_init = true
  print("<<script-pre-init>>")
end)

hook.script_post_cleanup:register('MOD-post-cleanup', function()
  state.script_post_cleanup = true
  print("<<script-pre-init>>")
end)

hook.system_post_startup:register('MOD-post-startup', function()
  state.system_post_startup = true
  print("<<system-post-startup>>")
end)

return api
```

the test mod obviously doesn't do anything at the moment other than print out something when the hook is called and set some module local state.  with this mod loading then in any script (or the repl) one can do the following:
```lua
local tab = require 'tabutil'
local mod = require 'MOD/lib/mod'
tab.print(mod.get_state())
```
...if everything is working (and at least one script has been loaded then the printed table should have one `true` entry per hook as well. the caveat with this setup is that if the mod has it's own api and a script uses `include` instead of `require` then the mod code will be loaded again, replacing the previously installed hook callbacks with new functions which access a new copy of the `state` variable.

regarding the nature of `dust/data/mods.lua`, i'm not 100% positive but i don't think `dust/data` is on the search path for `require` right now, as a result the current logic uses lua's `dofile` function to load/execute the contents of the file. `dofile` gets the job done but iirc it was not returning me any value (need to confirm). if `dust/data/mods.lua` was just a table of paths could be made to work but it doesn't seem any more compact or difficult the listing:
```lua
require 'passthrough/lib/mod'
require 'midigrid/lib/mod'
```
one advantage i see in letting `dust/data/mods.lua` be executable code instead of lua data which needs to be processed is that someone wanting to do some quick tweak like extending the package search path or defining some additional utility functions could do that right in `dust/data/mods.lua`. well i guess technically one still could as long as the value that resulted from loading that file was the table of stuff to include.

an alternate approach which i was experimenting with is to do away with the need for maintaining `dust/data/mods.lua` entirely and simply require any `dust/code/<script>/lib/mod.lua` which exists. this logic is implemented in this commit: https://github.com/ngwese/norns/commit/723d63ed6659ab6f9e573fc8eb18a4912f21f2fb#diff-fd8cfeecbaf7288d530c707188a6109f4898f184b17574b9de3141496c3ce5ad

the intension (not implemented in full) was to pair the pattern based loading logic with a system menu which required one to explicitly enable a "mod" after it is installed. the motivation for this was to accommodate folks that would be uncomfortable creating/editing/maintaining `dust/data/mods.lua`  - the ux for installing a "mod" shared via maiden (such as midigrid) would be to:

1. install `midigrid` via maiden
2. on device SYSTEM > MODS > MIDIGRID, and do some gesture to "enable"
3. SYSTEM > RESTART

the motivation for this being opt-in logic was that a script/project may offer libraries and scripts along side the mod and someone may want to try out the scripts or leverage the library functionality w/o having to also commit to the mod making system level changes.

tehn:
i like your final proposal! i'll implement the selector and get the dynamic ui working. thank you for the extensive writeup.

tehn:
@ngwese 

i'm likely having bad git etiquette, i apologize. working PoC for mod-defined submenus: https://github.com/monome/norns/tree/mod-submenus (it's a pretty tiny add)

here's an example mod which creates its own custom sub-menu:

```
local hook = require 'core/hook'

local state = {}
local api = {}

api.get_state = function()
  return state
end

-- register menu item
local function setup_menu(m)
  print("MENU: MODTEST")
  tab.print(m)
  _menu.m["MODTEST"] = m
end

-- MENU DATA
local m = {
  x = 0,
  y = 0
}

m.key = function(n,z)
  if n==2 and z==1 then
    _menu.set_page("MODS")
  end
end

m.enc = function(n,d)
  if n==2 then m.x = m.x + d
  elseif n==3 then m.y = m.y + d end
  _menu.redraw()
end

m.redraw = function()
  screen.clear()
  screen.move(64,40)
  screen.text_center(m.x .. "/" .. m.y)
  screen.update()
end

m.init = function() end -- on menu entry, ie, if you wanted to start timers
m.deinit = function() end -- on menu exit


-- dummy functionality to confirm things are getting called and test out access
-- to mod level state via mod supplied fuctions.

hook.script_pre_init:register('modtest-pre-init', function()
  state.script_pre_init = true
  print("<<script-pre-init>>")
end)

hook.script_post_cleanup:register('modtest-post-cleanup', function()
  state.script_post_cleanup = true
  print("<<script-pre-init>>")
end)

hook.system_post_startup:register('modtest-post-startup', function()
  state.system_post_startup = true
  print("<<system-post-startup>>")
  setup_menu(m)
end)

return api
```

we could clean up the syntax however desired, but this works drop-in as is.

i left TODO spots in `menu/mods.lua` for ENC 2/3 actions to enable/disable each mod. hadn't quite decided how to indicate them being on/off, however. highlighting might not be sufficient... but maybe a prepended `+` sign to the front of the mod name?

i'm thinking the enabled mods could be stored in `dust/data/system.mods` (given we have `system.favorites` and `system.state`) as a table, as you mentioned.

do you want to have a pass at the enable logic? i started, and then realized you had some intentional design with the structure and didn't want to clutter it up.

also each mod has a `return api` but the way the mods are loaded (via `require`) we seem to not capture this reference in a table anywhere? how is the best way for a script to perhaps "see" a mod? (since scripts are going to want to query this stuff).

perhaps `norns.mods` is an option?

ngwese:
cool, will check it out, give it a test!

ngwese:
@tehn - this is in a decent state now, ready for review and testing. i've done some further refactoring, exposing some menu and hook related functionality directly from the `mods` module.  in particular:
- provided `mods.menu` table which has a few functions which minimally wrap functionality which was previously accessed via `_menu`
- implemented the explicit enable/disable logic for mods and updated the menu to provide some visual feedback as to the enabled/loaded state
- fixed some case/naming sensitivity bugs
- provided `mods.hook.register` function so that menu registration and hook registration both use function invocation syntax (instead hooks using method invocation syntax).

i've also creating some minimal documentation/example here: https://github.com/ngwese/norns-example-mod

tehn (approved):
tested and works wonderfully.

i really like the selection/enable/disable UI and example mod you created. the tutorial is clear and i think you have plenty of built-in safeguards and best practices in mind.

i'll have a go at creating a "useful" mod (grid rotation/brightness?)

happy to merge this so others can have a "beta" go at it, unless there are any design concerns or suggestions first?

ngwese:
i think we are good to merge and beta. the devs behind `passthrough` and `midigrid` would probably be good testers. 

i recall @catfact was thinking of a tuning mod and i wanted to try it out as a way to extend the interpreter which some additional c-level functions.

ngwese:
i can also transfer the example mod to the monome org. i was unsure how much of this (if any) we wanted to incorporate in the main docs.

tehn:
thanks! and yes we will get a section in the main docs.

tehn:
@ngwese i've created a grid-settings mod: https://github.com/monome/grid-settings

the process was very straightforward, thank you!

ngwese:
cool! minor suggestion to use `paths` instead of `_path` 😀

tehn:
thanks!! i was pulling from `core/state.lua` as an example, so i can patch that up as well

PR #1386, opened 2021-06-23
focus/defocus callbacks

### LINKED ISSUE #1272: focus/defocus user callbacks
some scripts might want to know when you enter/leave the menu

proposed user callback names:
- `focus`
- `defocus`

question: should these be called when a user script calls the `fileselect` or `textentry`?

alternatively:

- `focus(state)` could just have the to-be-entered state as an arg ("menu", "script", "fileselect", etc)

tehn:
implemented user-definable callbacks, for example:

```
focus = function() print("FOCUS") end;
defocus = function() print("DEFOCUS") end
```

which are called when menu is toggled or a UI lib (fileselect, textentry, listselect)

also:

```
norns.focus()  -- returns true (script UI is focus) or false (menu/lib-ui)
```

fixes #1272

### DIFF
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -147,6 +147,7 @@ _menu.set_mode = function(mode)
   if mode == false then -- ACTIVATE PLAY MODE
     if _menu.mode == true then _norns.screen_restore() end
     _menu.mode = false
+    norns.focus_lib = false
     m[_menu.page].deinit()
     screen.clear()
     screen.update()
@@ -155,7 +156,7 @@ _menu.set_mode = function(mode)
     norns.encoders.callback = enc
     norns.enc.resume()
     redraw()
-  elseif mode == true then -- ACTIVATE MENu MODE
+  elseif mode == true then -- ACTIVATE MENU MODE
     if _menu.mode == false then _norns.screen_save() end
     _menu.mode = true
     _menu.alt = false
@@ -172,6 +173,7 @@ _menu.set_mode = function(mode)
     norns.encoders.set_sens(3,2)
     _menu.set_page(_menu.page)
   end
+  norns.focus_change()
 end
 
 -- set page
@@ -254,4 +256,5 @@ m["UPDATE"] = require 'core/menu/update'
 m["SLEEP"] = require 'core/menu/sleep'
 m["MIX"] = require 'core/menu/mix'
 m["TAPE"] = require 'core/menu/tape'
+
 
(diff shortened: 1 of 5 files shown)


tehn:
@tyleretters @catfact @dndrks 

can we review actual use cases where this would be necessary? i think i just assumed this made sense, but i'm suddenly having a hard time coming up with legitimate problems that this solves which couldn't be solved a different way without adding to the API.

that said, i think it does make sense to add more callback options for potentially-script-affecting system stuff (some of these should be done by refactoring certain functionality into classes, such as pmaps)

any insights welcome

ngwese:
iirc it came up that someone was getting flicking due to calling screen update outside of the redraw callback (since the menu system simply swaps the redraw callback out when it has focus).

i mentioned that wrote code to handle this in sky/tambla because it too didn’t use the redraw callback. i think at the time this came up i suggested a focus/defocus callback to formalize the pattern of detecting when your script has control of the screen and hardware controls and when it doesn’t instead my work around which was comparing the address of the redraw function.

dndrks:
[here's the original query which led to the ticket](https://llllllll.co/t/is-there-a-way-for-a-norns-script-to-know-that-the-user-has-tapped-k1-and-has-either-left-to-or-returned-from-the-norns-menu/39412)

tehn:
thanks for the insight.

if redraw handling is the only thing this potentially addresses, i would opt _not_ to merge this.

the menu system (i think) manages the entry/exit when the `redraw` function is used correctly (ie, it's fine to have it call other functions, which i do all of the time, but the redraw should be the _only_ origin of called screen functions so they can be properly blocked).

are there any other reasons to have focus/defocus?

catfact:
> are there any other reasons to have focus/defocus?

i suppose it could help to stop processes if you are really pushing the performance limits and running the menu on top of the script is crossing over into the system seizing up / falling behind. 

but no, it doesn't seem generally useful to me.

i do find the present behavior a little wonky. its weird to say, "here are these API functions, but don't use them except under a particular arbitrary condition. no, you won't get an error message otherwise - it will just basically work but with weird glitches."

what about: instead of just swapping the `redraw` function, swapping the whole `screen` module with a dummy screen module full of no-ops?

```

local screen = require 'screen'

local screen_dummy = {}

for k,v in pairs(screen) do
    if type(v) == 'function' then
        screen_dummy[k] = function() end
    end 
end 
```

tehn:
that's actually exactly what i did back in v1 (dummy-op the screen)

and i have no recollection of a justification for why i changed it, so yeah could go back there.

PR #1387, opened 2021-06-29
add util.file_size

dndrks:
### new: util.file_size

adds a `util` function which returns a file's size in megabytes:

```lua
> util.file_size(_path.code.."awake/awake.lua")
0.013152
> util.file_size(_path.audio.."tape/0000.wav")
71.710508
```
- if provided filepath isn't found:
```lua
> util.file_size(_path.audio.."tape/blahblah.wav")
no file found at /home/we/dust/audio/tape/blahblah.wav
<ok>
```
- if no path is provided:
```lua
> util.file_size()
util.file_size requires a path
<ok>
```

### DIFF
--- a/lua/lib/util.lua
+++ b/lua/lib/util.lua
@@ -46,6 +46,26 @@ util.file_exists = function(name)
   end
 end
 
+--- query file size.
+-- @tparam string name filepath
+-- @treturn number filesize in mb
+util.file_size = function(path)
+  if path ~= nil then
+    local f = io.open(path,"r")
+    if f~=nil then
+      local c = f:seek()      -- get current position
+      local s = f:seek("end") -- get file size
+      f:seek("set", c)        -- restore position
+      io.close(f)
+      return s/1000000
+    else
+      print("no file found at "..path)
+    end
+  else
+    print("util.file_size requires a path")
+  end
+end
+
 --- make directory (with parents as needed).
 -- @tparam string path
 util.make_dir = function(path)
--- a/lua/lib/util.lua
+++ b/lua/lib/util.lua
@@ -59,10 +59,10 @@ util.file_size = function(path)
       io.close(f)
       return s/1000000
     else
-      print("no file found at "..path)
+      error("no file found at "..path)
     end
   else
-    print("util.file_size requires a path")
+    error("util.file_size requires a path")
   end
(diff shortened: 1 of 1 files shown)


lua/lib/util.lua
```diff
@@ -46,6 +46,26 @@ util.file_exists = function(name)
...
+  if path ~= nil then
+    local f = io.open(path,"r")
+    if f~=nil then
+      local c = f:seek()      -- get current position
+      local s = f:seek("end") -- get file size
+      f:seek("set", c)        -- restore position
+      io.close(f)
+      return s/1000000
+    else
+      print("no file found at "..path)
```
artfwo, line 62:
```suggestion
      error("no file found at "..path)
```

lua/lib/util.lua
```diff
@@ -46,6 +46,26 @@ util.file_exists = function(name)
...
+    end
+  else
+    print("util.file_size requires a path")
```
artfwo, line 65:
```suggestion
    error("util.file_size requires a path")
```

lua/lib/util.lua
```diff
@@ -46,6 +46,26 @@ util.file_exists = function(name)
...
+--- query file size.
+-- @tparam string name filepath
+-- @treturn number filesize in mb
+util.file_size = function(path)
+  if path ~= nil then
+    local f = io.open(path,"r")
+    if f~=nil then
+      local c = f:seek()      -- get current position
+      local s = f:seek("end") -- get file size
+      f:seek("set", c)        -- restore position
```
artfwo, line 58: it's not necessary to seek to the start, if you're closing the file, the first `seek()` call can be also safely omitted.

artfwo (approved):

lua/lib/util.lua
```diff
@@ -46,6 +46,26 @@ util.file_exists = function(name)
...
+      io.close(f)
+      return s/1000000
```
artfwo, line 60: it might make sense to return bytes and let the script author choose the units to use, sometimes MB is 1048576 bytes too :)

dndrks:
@artfwo , thank you for the code review!! folded in your suggestions, updated PR

artfwo (approved):
good to go!

PR #1393, opened 2021-07-02
script loads & calls norns.crow.init() instead of crow.init()

trentgill:
We broke the crow initialization with the namespace update -- when a script is loaded it now calls `norns.crow.init()`.


tehn (approved):
omg

PR #1396, opened 2021-07-07
util.wrap and util.wrap_max: avoid min > max freezes

dndrks:
currently, if a script passes a smaller max than min to `util.wrap` or `util.wrap_max`, then norns will freeze. this PR adds an error print to maiden when the max is less than the min and avoids hangs

### DIFF
--- a/lua/lib/util.lua
+++ b/lua/lib/util.lua
@@ -252,15 +252,19 @@ end
 -- @tparam number max
 -- @treturn number cycled value
 function util.wrap(n, min, max)
-  local y = n
-  local d = max - min + 1
-  while y > max do
-    y = y - d
+  if max >= min then
+    local y = n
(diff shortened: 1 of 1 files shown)


dndrks:
@tehn , implemented your suggested behavior where min + max values are swapped if max < min

lua/lib/util.lua
```diff
@@ -252,6 +252,11 @@ end
 -- @tparam number max
 -- @treturn number cycled value
 function util.wrap(n, min, max)
+  local given_vals = {min,max}
+  if max < min then
```
tehn, line 256: whattabout
```
if max < min then
  local temp = min
  min = max
  max = temp
end
```

dndrks: good call!

tehn (approved):

PR #1397, opened 2021-07-17
port sequins to norns

tyleretters:
this is just a straight port from https://github.com/monome/crow/blob/34ce1e455f01fdef65a0d37aa97163b4cd14a115/lua/sequins.lua

@trentgill said it would be just a copy paste job and was right!


tehn (approved):
yes!! who thought of this ;)

PR #1398, opened 2021-08-04
hold K1 + scroll with E2 to jump character indices in SELECT

dndrks:
i found that i've been prepending my WIP scripts with A's so that I don't have to scroll through a bunch of scripts to even favorite them, so i thought borrowing the quick-scroll method from params made sense for the SELECT menu!

tested with every script installed via maiden, no hiccups:
- nav to SELECT
- hold K1 + turn E2 to jump by leading character
- if current position in the favorites, CW turn will jump to first non-favorite script

please lmk if it needs any additional cleaning!

### DIFF
--- a/lua/core/menu/select.lua
+++ b/lua/core/menu/select.lua
@@ -4,7 +4,9 @@ local m = {
   pos = 0,
   list = {},
   favorites = {},
-  len = "scan"
+  nav_indices= {},
+  len = "scan",
+  alt = false
 }
 
 local function menu_table_entry(file)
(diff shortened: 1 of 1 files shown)


tehn:
cool!

i agree there needs to be a quicker scroll method, but in my mind the character index is a very uneven guide for jumping given a script collection could be quite heavy with a certain character?

counter-proposal: just have K1+E2 jump by screen-fulls (ie, delta = 6 rather than 1)

dndrks:
@tehn , rad, thank you for the wayyy simpler suggestion! i've implemented it in the latest commit.

PR #1401, opened 2021-08-15
Post startup

catfact:
this change moves the execution of the mods post-startup hook until after initially-connected external devices have been scanned by `dev_monitor` and handled by appropriate lua modules.

its sort of a quick hack so please take a look and see if anything looks crazy!

also could easily be convinced that this is not necessary since there are plenty of workarounds available. (e.g. mod could monkeypatch `_norns.midi.add`)


catfact:
(wow github is using my fake email field to determine the contributor for these commits. how stupid)

lua/core/norns.lua
```diff
@@ -237,3 +237,8 @@ end
 _startup = function()
   require('core/startup')
 end
+
+_post_startup = function()
+   print('_norns._post_startup')
+   hook.system_post_startup()
```
ngwese, line 243: i get slightly weirded out by references to a module in a file which doesn’t import it. i realize that pattern is all over the core code. iirc the `mods` does import `hooks` but i double it is importing it globally and calling it `hook`.

matron/src/main.c
```diff
@@ -83,23 +83,44 @@ int main(int argc, char **argv) {
...
     // i/o subsystems are ready; run user startup routine
     w_startup();
+
     // scan for connected input devices
+    fprintf(stderr, "scanning devices...\n");
     dev_monitor_scan();
 
+    // handle all resulting events, then run "post-startup"
+    fprintf(stderr, "handling pending events...\n");
+    //event_handle_pending();
```
ngwese, line 118: is event_handle_pending intended to be commented out?

ngwese (commented):
took a quick look (on my phone so couldn’t test)

matron/src/main.c
catfact: oops no

lua/core/norns.lua
catfact: yea i agree. but we constantly assume the global names assigned in `startup`. hppy to do my part to change that style

catfact: ACTUally nvm i am locally importing `hook` b/c i am a good boy

catfact: actually nvm hook is a local alias, i'm a good boy

ngwese: sure enough. sorry for the false alarm (hard to review on the phone)

ngwese (approved):
gave a quick test on device. lgtm

Issue #1402, opened 2021-08-17
Broken crow APIs


discohead:
Another report of the same issue here: https://llllllll.co/t/crow-help-norns/25863/174

When setting the input modes from norns, it seems the most recent call to set a mode is applied to both inputs. Also seems like getters for `volts`, `slew`, etc... are broken and return tables instead of the values (details at the bottom).

Just got my crow today running v3.0.1 with my fates running norns for fates 210706. It seems to be a norns issue as I don't have the same problem when scripting from druid.

This norns code...

```
  crow.input[1].mode("stream")
  crow.input[1].stream = function(v) print("stream: "..string.format("%.3f", v)) end
  crow.input[2].mode("change", 2, 0.1, "both")
  crow.input[2].change = function(s) print("change: "..tostring(s)) end
```

produces this output...

```
change: true
change: false
change: -0.986633
change: true
change: false
change: 3.76117
change: true
change: false
change: 7.14529
```

Note that `stream` is not called, instead `change` is being called by both inputs, with `volts` from `input[1]` and `state` from `input[2]`.

The same code in druid...

```
input[1].mode("stream")
input[1].stream = function(v) print("stream: "..string.format("%.3f", v)) end
input[2].mode("change", 2, 0.1, "both")
input[2].change = function(s) print("change: "..tostring(s)) end
```

produces the expected output...

```
change: true
change: false
stream: -0.986633
change: true
change: false
stream: 3.76117
change: true
change: false
stream: 7.14529
```

I didn't manage to track down the bug, but my best guess is that this broke when @Galapagoose added support for the crow namespace: https://github.com/monome/norns/commit/4e9690297740d8b26c10c8cc57160ef40796820b

It appears there are other things broken as well, most of the getters are returning tables instead of the value and it doesn't seem to matter if the variable even exists or not...
```
crow.input[1].volts
table: 0x4f4990
crow.output[1].volts
table: 0x480420
crow.output[1].volts[1][1][1][1][1][1][1][1]
table: 0x480af0
crow.foo['bar']
table: 0x4ea400
``` 

in druid...
```
> print(input[1].volts)
0.001770258

> print(output[1].volts)
1.0

> print(foo['bar'])
repl:1: attempt to index a nil value (global 'foo')
stack traceback:
  repl:1: in main chunk
runtime error.
```

tehn:
@trentgill for extra eyes

trentgill:
[...]

PR #1403, opened 2021-08-18
Use tab.count to get count of table w/ string keys

discohead:
Fix for https://github.com/monome/norns/issues/1402

`norns.crow.events` is a table of string keys, the `#` operator only works on tables of integer keys.


trentgill:
Great thank you.

trentgill (approved):

PR #1409, opened 2021-08-28
Softcut NRT Enhancement: add preserve and mix to `read_buffer`

ryleelyman:
Adds `mix_buffer_mono(file, start_src, start_dst, dur, preserve, mix, chan_src, chan_dst)` and `mix_buffer_stereo(file, start_src, start_dst, dur, preserve, mix)` methods to softcut. These methods act like `read_buffer`, except they mix the existing buffer content with the new soundfile according to the `preserve` and `mix` params. See #1151.

Questions:
- should there be a fade_time parameter?
- is `clamp(frSrc, bufFrames - 1);` towards the beginning of `BufDiskWorker::readBufferMono` a bug? It seems like this says that we can't read a position in a file past the length of the softcut buffer?
- I mildly refactored `BufDiskWorker::Job` to add a `mix` parameter before the callback. I hope this is okay.
- when I run `crone.sh` or `start.sh` on my Norns after building, I get `liblo error: 9913; Invalid message path;`. Is this a problem?

### DIFF
--- a/crone/src/BufDiskWorker.cpp
+++ b/crone/src/BufDiskWorker.cpp
@@ -502,7 +502,7 @@ noexcept {
             goto cleanup;
         }
         file.readf(ioBuf, ioBufFrames);
-        for (int fr = 0; fr = ioBufFrames; ++fr) {
+        for (int fr = 0; fr < ioBufFrames; ++fr) {
             buf.data[frDst] = buf.data[frDst] * preserve + ioBuf[fr * numSrcChan + chanSrc] * mix;
             frDst++;
         }
--- a/crone/src/BufDiskWorker.cpp
+++ b/crone/src/BufDiskWorker.cpp
@@ -67,30 +67,17 @@ void BufDiskWorker::requestCopy(size_t srcIdx, size_t dstIdx,
 }
 
 void
-BufDiskWorker::requestReadMono(size_t idx, std::string path, float startSrc, float startDst, float dur, int chanSrc) {
+BufDiskWorker::requestReadMono(size_t idx, std::string path, float startSrc, float startDst, float dur, int chanSrc,
+                               float preserve, float mix) {
     BufDiskWorker::Job job{BufDiskWorker::JobType::ReadMono, {idx, 0}, std::move(path), startSrc, startDst, dur,
-                           chanSrc};
+                           chanSrc, 0.f, preserve, mix};
     requestJob(job);
 }
 
 void BufDiskWorker::requestReadStereo(size_t idx0, size_t idx1, std::string path,
-                                      float startSrc, float startDst, float dur) {
+                                      float startSrc, float startDst, float dur, float preserve, float mix) {
     BufDiskWorker::Job job{BufDiskWorker::JobType::ReadStereo, {idx0, idx1}, std::move(path), startSrc, startDst, dur,
-                           0};
-    requestJob(job);
-}
-
-void BufDiskWorker::requestMixMono(size_t idx, std::string path, float srcStart, float dstStart,
-                                   float dur, float preserve, float mix, int chanSrc) {
-    BufDiskWorker::Job job{BufDiskWorker::JobType::MixMono, {idx, 0}, std::move(path), srcStart, dstStart, 
-                           dur, chanSrc, 0.f, preserve, mix};
-    requestJob(job);
-}
-
-void BufDiskWorker::requestMixStereo(size_t idx0, size_t idx1, std::string path, float srcStart, float dstStart,
-                                     float dur, float preserve, float mix) {
-    BufDiskWorker::Job job{BufDiskWorker::JobType::MixStereo, {idx0, idx1}, std::move(path), srcStart, dstStart, 
-                           dur, 0, 0.f, preserve, mix};
+                           0, 0.f, preserve, mix};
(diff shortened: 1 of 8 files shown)


ryleelyman:
an idle thought: there's no particular reason to have this as a separate method from `read_buffer`, since with `preserve = 0`, `mix = 1`, they behave the same way.

catfact:
Awesome! I will be off my phone in a bit and should be able to give it a spin

Yea, I'm usually in favor of refactoring for fewer methods if it makes sense

The liblo error is not familiar to me...

crone/src/BufDiskWorker.cpp
```diff
@@ -437,6 +459,140 @@ noexcept {
     delete[] ioBuf;
 }
 
+void BufDiskWorker::mixBufferMono(const std::string &path, BufDesc &buf,
```
catfact, line 462: i think these are close enough to `bufRead` counterparts to warrant combining them. id replace `bufReadMono` etc with these funcs and default params. (maybe still worth having 2 lua funcs, or make the extra params optional from lua too, by checking args count in `weaver`.)

catfact (changes requested):
looks good! 

one typo that needs be fixed in lua function registration.

one refactor that could be done if you feel like it (otherwise i'll probably do it later, no worries eithher way)

catfact:
> an idle thought: there's no particular reason to have this as a separate method from `read_buffer`, since with `preserve = 0`, `mix = 1`, they behave the same way.

ah  -yea indeed. i'd like them to be combined in the cpp methods at least; probably in the OSC methods too. i think i would leave both methods in lua, but combine them and add default args in the `weaver.c` glue.

crone/src/BufDiskWorker.cpp
```diff
@@ -340,7 +319,7 @@ noexcept {
     size_t bufFrames = buf.frames;
 
     size_t frSrc = secToFrame(startSrc);
-    clamp(frSrc, bufFrames - 1);
+    clamp(frSrc, file.frames() - 1);
```
ryleelyman, line 322: I think this is what is actually meant here: the starting point in the source file should be at least one frame before the end of the source file.

crone/src/BufDiskWorker.cpp
ryleelyman: I ended up leaving only one function in lua, but made the extra parameters optional—which only needs work in `softcut.lua`, not in `weaver`.

matron/src/weaver.c
```diff
@@ -333,6 +335,8 @@ void w_init(void) {
     lua_register_norns("cut_buffer_copy_stereo", &_cut_buffer_copy_stereo);
     lua_register_norns("cut_buffer_read_mono", &_cut_buffer_read_mono);
     lua_register_norns("cut_buffer_read_stereo", &_cut_buffer_read_stereo);
+    lua_register_norns("cut_buffer_mix_mono", &_cut_buffer_mix_mono);
+    lua_register_norns("_cut_buffer_mix_stereo", &_cut_buffer_mix_stereo);
```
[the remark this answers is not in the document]
ryleelyman: got it!

ryleelyman:
changes done! now `read_buffer` has the functionality of `mix_buffer` but defaults to its old behavior, so no changes to old script code should be needed.

lua/core/softcut.lua
```diff
@@ -333,17 +333,21 @@ end
 -- @tparam number dur : duration in seconds. if -1, read as much as possible.
 -- @tparam int ch_src : soundfie channel to read
 -- @tparam int ch_dst : buffer channel to write
-SC.buffer_read_mono = function(file, start_src, start_dst, dur, ch_src, ch_dst)
-  _norns.cut_buffer_read_mono(file, start_src, start_dst, dur, ch_src, ch_dst)
+-- @tparam number preserve : level of existing material
+-- @tparam number mix : level of new material
+SC.buffer_read_mono = function(file, start_src, start_dst, dur, ch_src, ch_dst, preserve, mix)
+  _norns.cut_buffer_read_mono(file, start_src, start_dst, dur, ch_src, ch_dst, preserve or 0, mix or 1)
```
catfact, line 339: ooh i like putting the fallback here, very nice

catfact: we should definitely add more defaults, don't know why we haven't before

catfact (approved):
looks good to me. i haven't run the most recent on device but i'm ok to just merge and be sure we hit all the NRT stuff in beta testing.

lua/core/softcut.lua
ryleelyman: we could even go so far as to say 

```lua
cut_buffer_read_mono(file, start_src or 0, start_dst or 0, 
dur or -1, ch_src or 1, ch_dst or 1, preserve or 0, mix or 1)
```

catfact: Yeah exactly

ryleelyman: I'm happy to add those to `softcut.lua` if that's not outside the scope of this PR!

ryleelyman: just added! couldn't think of a sensible default for `samples` in `buffer_render`, so I left that one as-is.

catfact:
:+1:

PR #1410, opened 2021-08-29
Graceful handling for vcgencmd not installed.

winder:
When running in a docker container `vcgencmd` isn't going to be available and we fail with the following message:
```
we@e821707a6a8c:~/norns$ ./matron.sh 
attempting to bind socket at url ws://*:5555
*** WARNING *** The program 'matron' uses the Apple Bonjour compatibility layer of Avahi.
*** WARNING *** Please fix your application to use the native API of Avahi!
*** WARNING *** For more information see <http://0pointer.de/avahi-compat?s=libdns_sd&e=matron>
sh: 1: vcgencmd: not found
child killed (signal 11)
```

What happens is the command returns 127 and writes `vcgencmd: not found` to stderr. There is no other output so `fgets` doesn't read anything so the temp is not updated.

When making this PR I thought this was related to the `child killed (signal 11)` crash, but that is unrelated.

### DIFF
--- a/matron/src/hardware/stat.c
+++ b/matron/src/hardware/stat.c
@@ -21,12 +21,17 @@
 // static int fd[3];
 // static char buf[8];
 static pthread_t p;
+static bool have_vcgencmd;
 
 void *stat_check(void *);
 
 // extern def
 
 void stat_init() {
+    if (!(have_vcgencmd = system("which vcgencmd > /dev/null 2>&1") == 0)) {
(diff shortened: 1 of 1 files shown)


tehn (changes requested):
rather than check for the command's existence every single time the stat is run--- how about using a flag and checking for the command only once in stat_init() ?

winder:
Good idea, updated.

tehn (approved):

PR #1411, opened 2021-08-29
Add -h option to matron.

winder:
Two changes here:
1) New `-h` option to print out the available options.
2) If an invalid option is provided matron will terminate with the default error message:
```
~$ ./matron -a
./matron: invalid option -- 'a'
```


tehn (approved):
thanks!

Issue #1414, opened 2021-08-31
Question: help debugging norns docker environment


winder:
I'm trying to get the norns environment running in a docker container. The configuration is based on norns-dev, which works (including audio passthrough from the container), but it's configured to use software from 2018. I've updated the software to recent versions and everything is building and the stack runs, but is not working properly.

If someone could review the output I have below and suggest where to go from here I would really appreciate it.

jackd started with no error (once things are working I'll try :
```
we@e719483be391:~$ jackd --no-realtime -d dummy
jackdmp 1.9.12
Copyright 2001-2005 Paul Davis and others.
Copyright 2004-2016 Grame.
Copyright 2016-2017 Filipe Coelho.
jackdmp comes with ABSOLUTELY NO WARRANTY
This is free software, and you are welcome to redistribute it
under certain conditions; see the file COPYING for details
JACK server starting in non-realtime mode
self-connect-mode is "Don't restrict self connect requests"
```

crone started with no error:
```
we@2a57a8b183f5:~/norns$ ./crone.sh 
we@2a57a8b183f5:~/norns$ attempting to bind socket at url ws://*:5556
constructed Client: crone
constructed Client: softcut
initializing buffer management worker..
setting up jack clients..
engine sample rate: 48000
engine sample rate: 48000
starting jack clients..
compiling class library...
connecting ports... 
starting OSC interface...
osc methods: 
/hello []
/goodbye []
/quit []
/poll/start/vu []
/poll/stop/vu []
/set/level/adc [f]
/set/level/dac [f]
/set/level/ext [f]
/set/level/cut_master [f]
/set/level/ext_rev [f]
/set/level/rev_dac [f]
/set/level/monitor [f]
/set/level/monitor_mix [if]
/set/level/monitor_rev [f]
/set/level/compressor_mix [f]
/set/enabled/compressor [f]
/set/enabled/reverb [f]
/set/param/compressor/ratio [f]
/set/param/compressor/threshold [f]
/set/param/compressor/attack [f]
/set/param/compressor/release [f]
/set/param/compressor/gain_pre [f]
/set/param/compressor/gain_post [f]
/set/param/reverb/pre_del [f]
/set/param/reverb/lf_fc [f]
/set/param/reverb/low_rt60 [f]
/set/param/reverb/mid_rt60 [f]
/set/param/reverb/hf_damp [f]
/set/enabled/cut [if]
/set/level/cut [if]
/set/pan/cut [if]
/set/level/adc_cut [f]
[...]

PR #1415, opened 2021-08-31
Handle no mods found.

winder:
One error I asked about in #1414 doesn't seem to be a real error:
```
running startup...
start_audio(): 
---> I added the next line. the arguments to scan are null so it uses these defaults, which don't seem to be correct.
r = /home/we/dust/code/, p = */lib/mod.lua
lua: /home/we/norns/lua/core/mods.lua:60: bad argument #1 to 'pairs' (table expected, got nil)
stack traceback:
	[C]: in function 'pairs'
	/home/we/norns/lua/core/mods.lua:60: in function 'core/mods.load'
	/home/we/norns/lua/core/startup.lua:111: in main chunk
	[C]: in function 'require'
	/home/we/norns/lua/core/norns.lua:238: in function '_startup'
scanning devices...
```

This happened because my dust repo doesn't have a `code` directory. I updated the code slightly so that there is an error message instead of a stack trace:
```
start_audio(): 
glob failed for pattern (/home/we/dust/code/*/lib/mod.lua): No such file or directory
scanning devices...
```

### DIFF
--- a/lua/core/mods.lua
+++ b/lua/core/mods.lua
@@ -16,7 +16,7 @@ function Mods.scan(root, pattern)
   local p = pattern or Mods.search_pattern
 
   local matches, error = norns.system_glob(r .. p)
-  if not matches then return nil end
+  if not matches then return {} end
 
   local mods = {}
   local name_pattern = "^" .. r .. "([%w_-]+)/"
@@ -55,11 +55,6 @@ function Mods.load(scan, only_enabled)
     require(package_path)
     Mods.this_name = nil
     loaded_mods[name] = true
-  end
-
-  if scan == nil then
-    print('No mods found!')
-    return
   end
 
   for name, details in pairs(scan) do
--- a/lua/core/mods.lua
+++ b/lua/core/mods.lua
@@ -57,6 +57,11 @@ function Mods.load(scan, only_enabled)
     loaded_mods[name] = true
   end
 
+  if scan == nil then
+    print('No mods found!')
+    return
+  end
+
   for name, details in pairs(scan) do
     if only_enabled then
       if Mods.is_enabled(name) then
--- a/matron/src/weaver.c
+++ b/matron/src/weaver.c
@@ -2586,7 +2586,6 @@ int _system_glob(lua_State *l) {
     glob_status = glob(pattern, glob_flags, NULL, &g);
 
     if (glob_status != 0) {
-        fprintf(stderr, "glob failed for pattern (%s): %s\n", pattern, strerror(errno));
         lua_pushnil(l);
         lua_pushinteger(l, glob_status);
         return 2;
--- a/matron/src/weaver.c
+++ b/matron/src/weaver.c
@@ -2586,6 +2586,7 @@ int _system_glob(lua_State *l) {
     glob_status = glob(pattern, glob_flags, NULL, &g);
 
     if (glob_status != 0) {
+        fprintf(stderr, "glob failed for pattern (%s): %s\n", pattern, strerror(errno));
         lua_pushnil(l);
         lua_pushinteger(l, glob_status);
         return 2;


ngwese (changes requested):
I think it would be more appropriate to handle this at the point of origin for the scan value then treating a nil value as a no-op in load. What is the root case for the scan returning nil?

tehn:
this is confusing as "no mods" is a common state, and seems to resolve itself fine. is this a file path issue?

ngwese:
Looking at the code I believe this error is the result of the underlying `glob` call failing since the docker environment being used possibly lacks the dust directory structure setup.

winder:
Thanks for taking a look. I'll do some more experimenting, you make a good point about the scan function returning an empty list instead of handling nil so I'll figure out exactly what is failing in Scan instead of handling it in load.

The dust repo is being cloned to /home/we/dust, but something clearly isn't working properly in this environment.

If you could take a look, here is the container I'm working on. This Dockerfile is only the services, the libraries are installed by a parent container. There is a tmux config that launches the different processes as a kind of systemd alternative:
[Dockerfile](https://github.com/winder/norns-dev/blob/master/norns-test-dummy/Dockerfile)
[Tmux services](https://github.com/winder/norns-dev/blob/master/norns-test-dummy/norns.yaml)

ngwese:
I'll caveat this with don't have experience with Tmux services and only minimally use Tmux itself...

Looking at the Tmux service file gives me the impression that it is running using the `crone.sh` and `matron.sh` shell scripts - which could be bit rotten - from the source directory as opposed to `/home/we`. No idea if that is the problem but for reference the systemd units don't use those scripts, for example: https://github.com/monome/norns-image/blob/main/config/norns-matron.service

I don't believe any of the active norns developers are running components in docker. The components like crone and maiden which are largely agnostic to the host do see some development on desktop machines but matron; being the heart that glues the hardware and user experience together really wants to be run on an actual device.

winder:
@ngwese I added an error message in the glob function and made a simple modification to the lua script to gracefully handle this failure. The problem I saw was because my `dust` branch is out of date and missing the `code` directory. Is there a development branch I should be using?

The new error looks like this:
```
start_audio(): 
glob failed for pattern (/home/we/dust/code/*/lib/mod.lua): No such file or directory
scanning devices...
```

It sounds like this is a normal-ish scenario, so maybe this still looks like too much of an error?

**soapbox below**
Regarding hardware, of course, I agree that there's no substitute. Especially for release testing (and using!). A convenient development environment is also nice, my last audio project was Teensy based and I did not l like having to drag a device around during development. The docker environment also removes the $300 barrier that faces community developers who made it to the "youtube" stage of interest. [Personally, my interest grew exponentially after running a command at the top of this thread, it started Norns v1.0.3, and with a little more effort I had 'awake' running](https://llllllll.co/t/getting-norns-running-inside-docker-working/14938).

Thanks for the pointer to norns-image, I had briefly looked there earlier but missed some things. The shell scripts look very similar to the services, but I did find some differences.

Thanks for bearing with me, I realize this is not a well-trodden path... yet!

matron/src/weaver.c
```diff
@@ -2586,6 +2586,7 @@ int _system_glob(lua_State *l) {
     glob_status = glob(pattern, glob_flags, NULL, &g);
 
     if (glob_status != 0) {
+        fprintf(stderr, "glob failed for pattern (%s): %s\n", pattern, strerror(errno));
```
ngwese, line 2589: Sorry. I should have been more precise in my statement about addressing the error at the source. That comment was made w.r.t. the `Mod.scan` function. Printing an error here presumes that all cases where the glob pattern fails are error cases which may not be the case. Further the weaver functions in general use `luaL_error(...)` to signal an error so that the caller can determine the outcome. If anything `_system_glob` should probably be using `luaL_error(...)` instead of simply returning an error code as the second value.

If you are feeling motivated I'd recommend using `luaL_error` else leave the code as is.

lua/core/mods.lua
```diff
@@ -16,7 +16,7 @@ function Mods.scan(root, pattern)
   local p = pattern or Mods.search_pattern
 
   local matches, error = norns.system_glob(r .. p)
```
ngwese, line 18: If the underlying call is changed to throw an error then I think it is reasonable for this function to can catch the error, print out the error, and return `{}` as a fallback value.

ngwese (changes requested):

tehn:
any further takes on how to manage missing `dust` folder?

ngwese:
i’d rather leave the existing code as is or change the glob function to use lua error reporting as mentioned above. i can take on that change if need be.

PR #1419, opened 2021-09-07
make keyboard work w/ textentry

p3r7:
this allows to use the keyboard in `textentry` menu (`TAPE`, Wi-Fi...).

please note that i had to create a new module `lib/textentry_kbd.lua` to hold the (temporary) callbacks. otherwise i would have had a circular dependency between `core/keyboard.lua` and `lib/textentry.lua`.
there might be a more elegant way to do this.


tehn (approved):
yess!

PR #1420, opened 2021-09-07
don't load tape on cancel

p3r7:
when in textentry for tape name, and canceling (by hitting `K2` / `ESC` on keyboard):
- past behavior: a tape gets loaded anyway w/ a newly incremented `fileindex`, `K3` triggers record
- fixed behavior: tape doesn't get loaded, `K3` allows reentering the textentry for the tape name

### DIFF
--- a/lua/core/menu/tape.lua
+++ b/lua/core/menu/tape.lua
@@ -49,7 +49,7 @@ local function tape_exists(index)
   return util.file_exists(filename)
 end
 
-local function update_tape_index()
+local function read_tape_index()
   local f = io.open(_path.tape..'index.txt','r')
   if f ~= nil then
     local a = tonumber(f:read("*line"))
@@ -61,7 +61,9 @@ local function update_tape_index()
   while tape_exists(m.fileindex) do
     m.fileindex = m.fileindex+1
   end
+end
 
+local function write_tape_index_v(v)
   local f = io.open(_path.tape..'index.txt','w')
   if f == nil then
     os.execute("mkdir -p ".._path.tape)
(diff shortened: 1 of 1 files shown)


tehn (approved):

lua/core/menu/tape.lua
```diff
@@ -70,16 +72,25 @@ local function update_tape_index()
...
 local function edit_filename(txt)
   if txt then
     m.rec.file = txt .. ".wav"
   else
-    m.rec.file = string.format("%04d",m.fileindex) .. ".wav"
+    _menu.redraw()
+    return
+  end
+  if txt == string.format("%04d",m.fileindex+1) then
+    update_tape_index(v)
```
p3r7, line 93: NB: we use `update_tape_index()` instead of `write_tape_index_v(txt)`.
so it re-reads index value on disk before updating, which can be nice to handle a tad better **concurrent writes**.

still, **can concurrent writes actually happen (triggered by scripts)?**

even if this way of doing provides very basic protection against some LLW cases (last write wins), it's not that robust.
i could implement a strong serializable guarantee w/ a loop (given that `fileindex` can basically work as a [Lamport timestamp](https://en.wikipedia.org/wiki/Lamport_timestamp) but do we want to go there?

tehn: it'd be good to have sane protections. yes, all tape functions can be controlled via scripts via https://github.com/monome/norns/blob/main/lua/core/audio.lua#L147

p3r7: i was gonna PR something like

```lua
local function update_tape_index()
  local want_to_write = m.fileindex + 1
  read_tape_index() -- updates m.fileindex for next step
  while m.fileindex < want_to_write do
    write_tape_index_v(want_to_write)
    read_tape_index() -- read our own write case concurrent write happened
  end
end
```

but then realized that (even w/ coroutines) norns Lua env is single-threaded, so the above protection should be sufficient.

catfact:
No, concurrent writes are not possible

PR #1423, opened 2021-09-08
Add override to tabutil

ryanlaws:
Returns a new table, overriding default table's values with those of override value table, using the keys of the default.

Neither the default table nor the override table are mutated.

Examples of potential usage:
https://github.com/monome/norns/blob/main/lua/lib/lattice.lua#L17
https://github.com/monome/norns/blob/main/lua/lib/lattice.lua#L119
These uses the `nil` check  + default pattern for each arg to override. This utility could sweeten the syntax a bit by providing all the default values in one object, then reassigning the result to args (whose original value is used for the overrides).

### DIFF
--- a/lua/lib/tabutil.lua
+++ b/lua/lib/tabutil.lua
@@ -235,5 +235,16 @@ function tab.readonly(params)
   return proxy
 end
 
+--- return new table with defaults overridden by values
+-- @tparam table defaults base values (keys from this)
+-- @tparam table values override values
+-- @treturn table sorted table
+function tab.override(defaults, values)
+	local result = {}
+  for k,v in pairs(defaults) do 
+		result[k] = (values[k] ~= nil) and values[k] or v
+	end
+	return result
+end
 
 return tab


ryanlaws:
Example from maiden REPL:
```lua
x = ({a=1,b=2})
--  <ok>
y = ({a=5,c=6})
--  <ok>
z = tabutil.override(x,y)
--  <ok>
tabutil.print(x)
--  a	1
--  b	2
--  <ok>
tabutil.print(y)
--  a	5
--  c	6
--  <ok>
tabutil.print(z)
--  a	5
--  b	2
--  <ok>
```

p3r7:
i like the idea but not the name.

to me `override` sounds like mutation. maybe `overriden` would be more explicit about making a copy of the left input?

also, it's not consistent with what i've encountered in other programming languages.
- PHP calls it [replace](https://www.php.net/manual/en/function.array-replace.php)
- SQL and python's pandas have this called a `left_join`

likewise, it would be trivial to implement an `inner_join` which could prove itself valuable as well.

p3r7:
also, it's not consistent with what i've encountered in other programming languages.
- PHP calls it [`replace`](https://www.php.net/manual/en/function.array-replace.php) which isn't much better. it also has a `merge` that works more like a `concat`. 
- python's pandas calls it `merge` w/ different join options (`left`, `inner`, `outer`) and has the `join` shortcut. it works on a more complex data structure but the idea is similar.

there is indeed some synergy w/ SQL joins.
we could something like:
- `left_merge` (your implem)
- `inner_merge` (keep only keys present in both tables)
- `full_merge` or more simply `merge` (keep all keys of both tables)

image

ryanlaws:
The merge idea is good! Borrowing name may lead to some undesired expectations though.

How about `gather`?

tehn:
i agree that given the types of merges, this is a pretty specific case to have such a general name.

another option would be to include a flag param for type of merge, but that's probably more confusing than just having separate merge functions that are named as such

tyleretters:
i like `gather`. i'd use this.

ryanlaws:
@tehn and @p3r7 what do you think of this?

p3r7:
to recap our last chat between @ryanlaws & me + some additional toughts:

**usefulness**

i still pretty like the idea, it's a pretty common use-case to have to:

```lua
local overriden_conf = tab.gather(<default_config>, <k-v_overrides>)
```

**implem**

indeed, it's close to a `left_merge`, *but* this version is **immutable** (makes a copy of `<default_config>` to not alter it globally).

if ever deemed useful, `merge` functions could be added at a later time (in which case the implem would get simplified into something like `tab.left_merge(tab.copy(<default_config>), <overrides>)`).


**naming**

lots of programming languages don't have this fn (which makes no sense when they do [opy-on-write](https://en.wikipedia.org/wiki/Copy-on-write)).

so we don't have much examples of a "standard" name for this feature.

i'm ok w/ `tab.gather`, as well as `tab.update` (js & Clojure style) or `tab.replace` (PHP style).

as we discussed previously, would be nice to have feedback from people w/ greater experience w/ Lua and its naming convention (@tehn, @catfact...).

ryanlaws:
Thanks for getting back to me @p3r7 - I like the `tab.update` idea. Added to branch.

Issue #799, opened 2019-04-04
audio engine fail after update, sometimes


catfact:
quoting @tehn:

note: i’ve sometimes seen a fresh startup with “audio engine fail” after update. either reboot or RESET through the menu and everything will be happy. trying to locate the reason for the first-up error (that then goes away)

catfact:
i have a vague suspicion that this is somehow related to the way we add SC include paths (with [`norns_config.sc`](https://github.com/monome/norns/blob/dev/sc/norns-config.sc).) if you wipe out the .yaml or remove and recreate the config class (??) then sclang it has to go through a cycle before the .yaml is created again. 

that is:

- update wipes out SC config (does it?)
- next boot, SC runs the `initClass` method in norns_config.sc and writes the config .yaml
- but _in the meantime_, it can't find classes in `~/dust` or whatever and it fails to compile.
- on the _next_ run, the new .yaml is loaded and everything is happy.

catfact:
if thus is indeed what's happening, we could a) run `sclang` once as part of the update script, b) retire `norns_config.sc` and go back to symlinks

antonhornquist:
how about removing `norns_config.sc` but adding the generated ´yaml` file in the installation process?

simonvanderveldt:
I guess alternatively we could just have a fixed `sclang_conf.yaml` and not do the changes to the `includePaths` during init?

catfact:
that seems dangerous, since the .yaml can be blasted by sclang any time

okyeron:
My experience over the last day or two - reboot I get the screen showing `error: audio engine` - sometimes it will play the startup warble, sometimes not (which is very odd to me). 

A start/stop or RESET and I'm usually back in business.

FWIW - I had killed  the .yaml file the other day, but all the paths seem valid right now

antonhornquist:
@catfact only if `LanguageConfig.store` is run, no? and then, i was under the impression the yaml file actually is actually read to the LanguageConfig on boot, so storing would, uh, restore things.

catfact:
i have not dug deep into whats up w/ the .yaml lifecycle, but yea - paths should be valid after running sclang a couple times, no matter what you do.

is there any actual objection to symlinks?

[...]

PR #1424, opened 2021-09-11
Bootstrap supercollider after installing trampoline.

winder:
When setting up matron / supercollider for the first time you need to bootstrap things by running `sclang` twice. This codifies that process.

This idea was shamelessly borrows from this comment: https://github.com/monome/norns/issues/799#issuecomment-483065692

This approach works great with my docker environment, but I'm not sure if it fits with the way `install.sh` is normally used. Figured I'd make a PR and see what people thought.

### DIFF
--- a/sc/install.sh
+++ b/sc/install.sh
@@ -2,3 +2,4 @@
 
 sc_ext_dir=~/.local/share/SuperCollider/Extensions
 cp norns-config.sc $sc_ext_dir/
+echo | sclang


catfact:
it's a little problematic to me because we have `sclang` set to boot `scsynth` as well. this starts up a JACK daemon, kicking off stuff that outlives the script and is unexpected when you are just trying to copy a couple files.

if we really need to solve this behavior, which i don't think is a major issue, i'd rather do it with a lighter touch, by manually generating/copying the sclang config .yaml, or at least by invoking `sclang` here such that it skips our normal startup routine (somehow, if possible.)

catfact:
ok, so after recalling more about original issue - in that thread i think we concluded that it would be fine to retire `norns_config.sc` and simply install symlinks from SC support directory to norns directories. 

we never did this simply because it doesn't really  matter to us - this script is never actually run by users or during update cycles.

catfact:
anyways - while i appreciate the intention here i would rather not put this in the main repo. we just want install script to copy files or make links, not kick off heavy processes that outlive the life of the script.

PR #1425, opened 2021-09-14
custom events

ngwese:
this change opens up matron's event loop for access by c-based lua modules loaded at runtime.

the motivation for this change is to allow hardware devices such at the soundplane to have have the same direct, low overhead, channel for pushing events into the running script that built in matron devices enjoy. ...while avoiding the need to integrate/maintain the soundplane support directly in matron.

the approach is straightforward and modeled a bit after how the linux kernel achieves modularity. the new `event_custom` struct contains two pointers, one is an `ops` structure with functions pointers which the event loop core uses to handle the custom event behavior, the other is an arbitrary `value` pointer which gets passed back into the custom functions.

as a proof of concept (and to validate the changes) i have a mostly complete external lua module for demo purposes:
https://github.com/ngwese/norns-event-demo

the demo still needs some polish and documentation.

### DIFF
--- a/matron/src/event_types.h
+++ b/matron/src/event_types.h
@@ -3,6 +3,7 @@
 #include "oracle.h"
 #include "osc.h"
 #include <stdint.h>
+#include <lua.h>
 
 typedef enum {
     // unused (do not remove)
@@ -315,7 +316,7 @@ struct event_softcut_position {
     float pos;
 };
 
-typedef void (*event_custom_weave_op_t)(void *value);
+typedef void (*event_custom_weave_op_t)(void *value, lua_State *lvm);
 typedef void (*event_custom_free_op_t)(void *value);
 
 struct event_custom_ops {
--- a/matron/src/event_types.h
+++ b/matron/src/event_types.h
@@ -85,6 +85,8 @@ typedef enum {
     EVENT_SOFTCUT_RENDER,
     // softcut position callback
     EVENT_SOFTCUT_POSITION,
+    // custom events defined in lua extensions
+    EVENT_CUSTOM,
 } event_t;
 
 // a packed data structure for four volume levels
@@ -313,6 +315,21 @@ struct event_softcut_position {
     float pos;
 };
 
+typedef void (*event_custom_weave_op_t)(void *value);
+typedef void (*event_custom_free_op_t)(void *value);
+
+struct event_custom_ops {
+    const char *type_name;
+    event_custom_weave_op_t weave;
+    event_custom_free_op_t free;
+};
+
+struct event_custom {
+    struct event_common common;
+    struct event_custom_ops *ops;
+    void *value;
+}; // +8
+
 union event_data {
     uint32_t type;
     struct event_exec_code_line exec_code_line;
@@ -348,4 +365,5 @@ union event_data {
     struct event_system_cmd system_cmd;
     struct event_softcut_render softcut_render;
     struct event_softcut_position softcut_position;
+    struct event_custom custom;
 };
--- a/matron/src/event_types.h
+++ b/matron/src/event_types.h
@@ -3,7 +3,6 @@
 #include "oracle.h"
 #include "osc.h"
 #include <stdint.h>
-#include <lua.h>
 
 typedef enum {
     // unused (do not remove)
@@ -316,14 +315,8 @@ struct event_softcut_position {
     float pos;
 };
 
-typedef void (*event_custom_weave_op_t)(lua_State *lvm, void *value, void *context);
-typedef void (*event_custom_free_op_t)(void *value, void *context);
-
-struct event_custom_ops {
-    const char *type_name;
-    event_custom_weave_op_t weave;
-    event_custom_free_op_t free;
-};
+// forward declaration to hide scripting layer dependencies
+struct event_custom_ops;
 
 struct event_custom {
     struct event_common common;
--- a/matron/src/event_types.h
+++ b/matron/src/event_types.h
@@ -316,8 +316,8 @@ struct event_softcut_position {
     float pos;
 };
 
-typedef void (*event_custom_weave_op_t)(void *value, lua_State *lvm);
-typedef void (*event_custom_free_op_t)(void *value);
+typedef void (*event_custom_weave_op_t)(lua_State *lvm, void *value, void *context);
+typedef void (*event_custom_free_op_t)(void *value, void *context);
 
 struct event_custom_ops {
     const char *type_name;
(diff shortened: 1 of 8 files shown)


artfwo:
given that you still need to implement an external "driver" for soundplane for pushing events, implementing this in the matron device i/o layer instead could provide a slightly better scripts integration (handling plug/unplug events, device-specific APIs, etc.).

as for supporting generic controllers that require custom events - can matron already do this with a relatively low overhead using OSC?

ngwese:
fwiw while the soundplane is the motivator for this change the custom event mechanism is useful for many other purposes.

maintaining integrations like the soundplane outside of the core code base (i.e. not making it a built in device) has several advantages
- compile time for the driver on device is about the same as the entire norns stack
- there are 90 soundplanes in existence total, the cross section of that and norns owners is small
- avoids the need to play C++ compile and multiple build system games
- the soundplane doesn’t fit into any existing device category (midi, osc, hid, arc, or grid)

doing the integration as a separate process and using OSC is undesirable for few reasons
- the soundplane driver has 20-30 tunable parameters and a whole touch zones system which would would be clunky and time consuming to expose via osc from matron to process (we already know osc isn’t great for rpc type interaction)
- managing the lifetime of an external process from code running in matron (menu or script) would be clunky, using systemd to do it means making system level changes that would create maintenance problems for us if other users started to leverage the pattern
-  the soundplane driver can put out a lot of data, 1Hz scanning rate. taking all of that, encoding as osc, sending through the kernel networking stack, decoding osc inside matron, and routing to lua callbacks is a lot of overhead relative to passing a pointer to structure in a ring buffer from one thread to another via the event queue. the touch tracker alone consumes 15% of a core on an rpi3 before sending data. adding osc/networking overhead on top seems further strains the system

for reference the “external” driver code (in progress) which is looking to take advantage of this is https://github.com/ngwese/soundplaneclient

ngwese:
one additional change to this PR which is needed is to have `event_handle` pass the address of the lua vm to the custom event’s `weave` op

tehn:
wow--- i hadn't really seen this sort of modularity implementation before and i'm very impressed.

given we're on a fixed platform, theoretically a built object (.so) could be included in the `norns-event-demo` repo so it wouldn't need to be compiled by the enduser?

ngwese:
adding a .so directly into a git repo is generally not recommended since the history will contain every version of the binary committed bloating the size of the repo. this is actually a case where using the existing support for installing via url to a .zip archive would be better than url to git repo. the end user wouldn’t need to do any compilation.

regarding single vs multi-arch, the search path for the c based module is defined in the mod glue so if ever there was a need one could put architecture specific binaries into subdirectories and extend the path accordingly on the device.

ngwese:
okay @catfact and @tehn - i think this is ready for a proper review. i'll follow up with improvements to the demo module in a bit.

ngwese:
i've put a bit of polish on the `norns-event-demo` mod and added some documentation. to see this whole thing in action:

- build matron with the change from this PR
- restart matron and run the following in maiden REPL `;install https://github.com/ngwese/norns-event-demo/releases/download/v0.1/norns-event-demo.zip`
- go to SYSTEM > MODS, select NORNS-EVENT-DEMO, and turn E3 to enable
- restart matron via maiden with: `;restart`
- in the REPL run the lua code listed at the end of the [README](https://github.com/ngwese/norns-event-demo/blob/main/README.md)

tehn:
excited to dig into this! (parents in town until the weekend, sorry for the delay)

catfact:
sorry i haven't had a chance to do hands-on test. 

this code and the demo module look good to me.

one little thought it raises is that modules break the pattern of keeping all lua FFI glue in one place, instead locating FFI with the C functionality. this is the same pattern adopted by @tehn in his rewrite project `isms`. so it raises the question in my mind of whether this new pattern is perhaps preferred and should be adopted throughout. (personally i like that modules like `metro` and `device_*` are agnostic of the scripting layer.)

ngwese:
i very much like the separation of the lua functions that exists today. the experiments i’ve done in the past of turning the drawing, event handling, input handling, and screen portions of matron into a library and loading that into another language runtime (common lisp) would have been much more difficult if not impossible if the lua glue had been mixed in. keeping things separate also helps with make the sum total of the lua interface clear.

i wouldn’t advocate changing that. 

truth be told i don’t like that i had to add an include to lua.h at the top of event_types.h. the only other option i could think of was to just use void* everywhere but that felt even more dirty. as things are now it is absolutely the tiniest pin prick of an encapsulation leak that i could come up with. the co-location lua glue and functionality in the loadable module is out of necessity more than anything. i not sure we want to embark on building out a full internal api to support modules which directly call anything other than the event loop bits.

ngwese:
after reflection based on usage i’m going to add a “context” pointer to the custom event struct and expose that to the callbacks as well. the change will make it easier to handle cases where the custom event value needs to be paired with another abstraction which represents an instance of a device (for example)

ngwese:
i've updated the `norns-event-demo` code to reflect the addition of the `context` value and altered the demo/testing instructions above.

tehn:
ah we should get `cmake` into the default image

(testing now!)

tehn:
this works great and i learned a lot reading the code.

i acknowledge the existing separation of lua glue for the rest of the codebase and i don't think we should go overhauling, but this one case seems like it makes sense as is.

this feels like a exceedingly flexible way to add extensibility to the system, thank you for putting in the time

tehn (approved):

matron/src/weaver.c
```diff
@@ -53,7 +53,7 @@
 
 //------
 //---- global lua state!
-lua_State *lvm;
+static lua_State *lvm;
```
catfact, line 56: i did in fact notice this! thanks.

yes i would fully endorse making all things explicitly `static` that can be.

matron/src/event_types.h
```diff
@@ -3,6 +3,7 @@
 #include "oracle.h"
 #include "osc.h"
 #include <stdint.h>
+#include <lua.h>
```
catfact, line 6: ouch! it hurts! but ok.

ngwese: the only other option that i can think of is to explore keeping the ops structure opaque and move it to a separate header. that might be enough to keep the exposure to lua limited to both weaver.c and the external code.

ngwese: awareness of lua has been moved to `event_custom.h` which is only needs to be included by external code which implements  custom events and internally to `event.c` and `weaver.c`.

Issue #10, opened 2017-03-06
folder structure


tehn:
currently all of the lua scripts live in the same place. eventually we'll need to split the app-level files and user-files into different locations.

first draft proposals:

`~/norns/config.lua` config file (last used script, state, etc?)
`~/norns/script` user scripts (including default scripts)
`~/norns/state` corresponding states for scripts
`~/norns/audio` audio files

some sort of user-install script can set this up, copy default scripts etc.

this way the `norns.lua`, `helper.lua`, etc will be in the application folder, away from user manipulation. (though users manipulating this stuff is fine, it should just be separated).

of course, this makes development a little annoying, so it can wait probably

PR #1429, opened 2021-09-17
Support PSET entries above 99

p3r7:
This PR makes PSET files list parsing more generic.

Before this PR, PSET list show only up to entry `#99`.
PSET entry `#100` could be saved but not loaded. Worse, it in fact overrode entry for PSET `#10`.


tehn (approved):
yikes and thank you!!

PR #1431, opened 2021-09-17
Midi clock broadcast

tekknovator:
Allow sending midi broadcasts to all devices as discussed on lines here:
[https://llllllll.co/t/norns-midi-clock-to-all-connected-devices/48673](url)
The change also reduces the logic in the clock.run loop by accessing a function in an array. If I understood correctly, this costs more memory but saves decisions by moving the conditions into the parameter setup.
Todo: Offer broadcast to connected devices only by filtering the array. I suppose a smaller array might bring even better timing. All performance gains a hypothetical. I did not measure anything because I do not know how. Suggestions welcome.

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -215,21 +215,41 @@ function clock.add_params()
       norns.state.clock.link_quantum = x
     end)
   params:set_save("link_quantum", false)
-  local clock_midi_out_table = {"off"}
+	
+  local clock_midi_out_parameter_table = {"off"}
 	local max_midi_clock_out_devices = 16
+	local clock_midi_sync_function_table = {(function() print("midi clock send is off") end)} -- !DAS ERGIBT NIL VALUE!!!! SCHEISSE!
+	local clock_midi_connected_device_numbers = {}
+
   for i = 1,max_midi_clock_out_devices do
     local short_name = string.len(midi.vports[i].name) < 12 and midi.vports[i].name or util.acronym(midi.vports[i].name)
-    clock_midi_out_table[i+1] = "port "..(i)..""..(midi.vports[i].name ~= "none" and (": "..short_name) or "")
+		print("adding "..short_name.." to clock out table at i: "..i)
+    clock_midi_out_parameter_table[i+1] = "port "..(i)..""..(midi.vports[i].name ~= "none" and (": "..short_name) or "")
+		if midi.vports[i].name ~= "none" then
+			clock_midi_sync_function_table[i+1] = (function() print("midi clock send is set to CONNECTED vport: "..i) midi.vports[i]:clock() end)
+			clock_midi_connected_device_numbers[i] = i
+		else
+			clock_midi_sync_function_table[i+1] = (function() print("midi clock send is set to disconnected vport"..i ) end)
+		end
 		if i == max_midi_clock_out_devices then -- TODO: if clock in is set to midi, we do not allow broadcast. 
 																						-- If I find a cost efficient way, we will filter the incoming device to avoid feedback.
-			clock_midi_out_table[i+2] = "Broadcast"
+			clock_midi_out_parameter_table[i+2] = "Broadcast"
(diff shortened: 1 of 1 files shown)


tehn:
i've been considering another "virtual" midi device which is "all devices" --- so that it could be selected normally through the vport system and also used by normal scripts.

the question is, perhaps: do people ever want to send notes/etc to all devices besides clock?

(i believe there's a feature like this in max/msp?)

the one bonus of such a feature is that for random-single-connections, using the "all midi" channel would mean everything always works... ie, if you hotswap two synths that want to receive notes from a norns sequencer.

sorry to derail this PR with a slightly more complex feature addition.

@dndrks @ngwese @catfact

catfact:
broadcasting midi clock to all devices seems like a normal feature (even a reasonable default.)

mmirroring *all* midi output to all devices seems like a weird niche feature for which i am having a hard time thinking of a real use case.

> (i believe there's a feature like this in max/msp?)

i am not familiar with such a feature and am not able to discover it at present in Max 8.

-------------


regarding this PR, the idea is sound but the refactoring sort of misses the point, which would be to avoid lots of conditionals &c in the pulse update function. 

i can take a shot at cleaning it up i suppose, and add protection against clock feedback. (did realize that i have a couple of usb-midi converters and a couple of drum mcahines, so i can in fact test locally.)

tehn:
indeed it's probably too weird (and also i don't have a computer that runs max/msp but that's not really a strong precedent)

avoiding clock feedback is a good point.

i'm wondering if we need more granular control over the midi system if we're going to be pursuing these daw-like builtin structures--- for example, ableton has on/off settings per midi in/out for track/sync/remote

(sorry, i'm just trying to anticipate a future FR that will present a breaking pattern for whatever the next implementation is)

ngwese:
sorry. meant to respond earlier. i couldn’t think of a need for echoing all midi traffic other than pushing norns to act like like a midi thru / distributor. that use case really seems to want for a generalized routing layer along the lines of what motu midi interfaces offer. the only other case of sending to all midi outputs that came to mind is implementing a panic/all notes off button but that doesn’t need any special support.

distribution of clock and other midi real-time messages like start, stop, continue does seem like a reasonable base feature.

tekknovator:
Having virtual device "all devices" would be very nice indeed.
There is also synth stacking which could benefit of @thens proposal to allow sending all data to all ports, if you use norns as a sequencer. I would not add any global menus beyond such an entry though, too much clutter. If it is really required, it can be implemented at script level (ie transport midi commands etc.).

> the refactoring sort of misses the point, which would be to avoid lots of conditionals &c in the pulse update function.
Been very naive approaching this, sorry. @catfact please look again at the diff. I tried to move the conditionals into the parameter setup, leaving only the broadcast with a nasty for loop. Is what I did futile? Or asked different: Is a LUA function call as expensive as a conditional? Sorry again for beeing so dense, but performance is of no concern for the server stuff I do usually.
My thinking was that accessing a function in any table can't be much more expensive then accessing a function in the G table.
What I am definitely not happy with, is the iteration over all 16 devices and the dummy function calls if the device is not connected. This would be improved if I had bothered to filter the table down to only connected devices, adding iterations for each connected device only. Any other ideas?

tehn:
ok--- let's just keep this to clock. agreed that the implementation needs some optimization

catfact:
@tekknovator really sorry for the delay on this! it sort of slipped my mind in a busy couple weeks.

if you don't mind, could you edit your PR to "allow edits from mainainers?" its a little checkbox in the bottom right.

image

then i can go ahead and just make a little optimization pass.

if you don't want to do that i can just re-do the change on a separate PR, that's fine too.

tekknovator:
Hi @tehn any progress on the proposed "all midi devices" virtual midi device? If that is no priority, I would like to take another try and fix this PR up.

tekknovator:
I just discovered passthrough mod. Gonna play with that. It seems that i can use it to distribute an external clock instead of norns own. If that works for me, i will abandon this PR and close it.

PR #1435, opened 2021-09-18
Use SDL surface when configured with --desktop.

winder:
Minimal set of changes to create an SDL backed cairo surface when building with `./waf configure --desktop`.

Using this code I was able to run norns in a docker container and view the screen remotely.


tehn (commented):
this is simpler than expected--- looks good. and thank you for adding the error messages.

i need to chat with the core devs about feature collision... will revisit soon

PR #1439, opened 2021-09-20
gamepad support

p3r7:
i came up w/ this during the WE.

`core/gamepad` provides an API of callback similar to `core/keyboard`'s.

the 2 script-registerable callbacks are `gamepad.dpad(axis,val)` (directional pad presses) and `gamepad.button(b,val)` (other button presses).

the tricky bit is that many gamepads act crazy:
- they don't respect the HID key codes described in `core/hid_events` (e.g. `BTN_SOUTH`, `BTN_TL`...)
- they sometimes treat their X/Y axis in reverse (sometimes both, sometimes only one....)

that's why many video games will prompt you w/ a wizard to configure your gamepad if it hasn't detected it in its whitelist.

so what i did was to make a whitelist of supported models `core/gamepad_model/index` w/ the idea of having a wizard (maybe a as a script?) to allow user to register new devices.

we could also make a fallback entry that assumes a well-behaved gamepad.

everything works but code could be optimized a bit and is in dire need of testing more models (i only have 1 at my place), esp. those w/ joysticks (mine doesn't have any).


zunaito:
i have a xbox style gamepad somewhere in storage (2 joysticks iirc)

what's the best way to help/test?

p3r7:
@zunaito i could make a small script that'd log stick / dpad / button presses.

the forked code [does it already](https://github.com/monome/norns/blob/397233d96b48c10b406fcdd9742ecbd45bd54662/lua/core/gamepad.lua#L52) when [this debug flag](https://github.com/monome/norns/blob/397233d96b48c10b406fcdd9742ecbd45bd54662/lua/core/gamepad.lua#L15) is set to a verbosity of `2`.

but first and foremost, i guess we'd have to decide if this PR is a good idea in the first place.

catfact:
i think it's a great idea and definitely was hoping someone would hit those TODOs and add a gamepad HID class and module.

the part i didn't anticipate is using it for menu navigation. i defer to @tehn on that but can't think of any objections myself.

i can add profiles for an old xbox controller and a switch pro.

tehn:
will take a look re: menu nav

p3r7:
so i did a small script that helps debugging a HID device behaviour.

[p3r7/hid-monitor](https://github.com/p3r7/hid-monitor)

i could easily extend it w/ a wizard to generate a gamepad config file

p3r7:
welp, so i've found a USB adapter for PS1/PS2 controllers in a box.

i'm gonna be able to continue this.

gatedfuzz:
i have a stadia and switch usb controllers that seem to already work great with other RPi based projects, but theyre rather less common i think. Would they be worth me testing and documenting?

p3r7:
alright, managed to handle analog sticks.

still need to:
- support `C` & `Y` keys (megadrive style controllers)
- support joystick button clicks
- get rid of costly calls to `gamepad.code_2_keycode` which are now only used for axis events
- do a device menu integration / configuration wizard
 
right now i only use the hid name to identify makes & models. this is rather limitative. typically, my PSX 2 USB adapter registers as `DWC OTG Controller` which is like the most generic name possible...

so having a dedicated selection menu for controllers (+ ideally a wizard to generate a [config file](https://github.com/monome/norns/blob/b508d3700db7d1e71c6e35322d507d718ab6260e/lua/core/gamepad_model/psx_usb_adapted.lua) is certainly more viable.

@gatedfuzz appreciated. this is still a tad early but it'll be interesting testing more controllers when we'll have something a tad more user friendly. i'll `@` you when that comes.

p3r7:
alright, it has been a while.

i've found the motivation to hack together a working wizard (for now as a script but the idea would be to integrate it in the device menu): https://github.com/p3r7/gamepad_wiz

for now it's pretty barebones graphically (it says "press <BUTTON>") and behaves like [RetroPie's](https://retropie.org.uk/docs/Controller-Configuration/).

it'd be easy to make it pretty w/ graphical indication on which button to press.

very quick mockup of what could be done:

image

catfact:
what would you think about first merging just the "low-level" support into norns/core, relying on device descriptor/whitelist? to me it seems useful by itself, especially if we add a few more common descriptors. 

adding a wizard (whether its a script/mod/core library) seems like a separate thought.

p3r7:
i'm ok with it!

what got me digging in the wizard thing is that i fear that doing the device matching based solely on the HID device name might be overly simplistic. maybe is there a smarter way? (md5 of `device.name` + `device.codes` + `device.types`...)

the other approach (taken by RetroPie) is to not embed any profile but instead have users create configuration maps for their own controllers using a wizard. profiles could then be "enabled" in the device menu.

catfact:
yo! are we happy to make this non-(WIP) and merge it? it's old but i think nothing has changed around here

p3r7:
alright!

there will certainly be more stuff to figure out (best way to configure controller profiles) but i'm pretty happy about the `gamepad.lua` extension).

the branch of this PR is from an old commit, i will certainly need to fast-forward it.
please note that i suck at git and may mess this up :)

p3r7:
well, no conflict when syncing!

so the forked branch should be mergeable.

catfact (commented):
finally got a chance to try things out and everything working as expected! all makes sense and looks proper to m

catfact (approved):

p3r7:
thanks for taking the time to review it.

petercolombo:
Hello!

I'm attempting to use a Sony-manufactured USB gamepad from the PS1 Mini set. Successfully downloaded this onto both a stock norns (silver enclosure, running 230614) and a fates (questionable authenticity re: PCB but running the most recent firmware as well). Upon restarting and starting the script, when establishing the analog d-pad inputs, I would get random double-triggers, even after going back one page and recalibrating, and in the absence of analog joysticks, after hitting skip and believing I've successfully created and saved a profile (getting the FINISHED screen), both units were bricked after restart. Maybe I attempted to back-page by hitting K2 and double-saving the profile, maybe the double-triggers at the beginning of the calibration are contributing factors, but since the norns would both not boot (warranting fresh installs on both units just to work), I was unable to retrieve any text being barfed out in real time. 

Should I try again? Am I missing something, perhaps running too-unstable a firmware (does this work on norns image 230526, for example)?

Please advise! Would love to get this working, am even open to picking up an 8BitDo wired controller with joysticks and trying it out, as long as I know it won't brick my norns. Thanks!

p3r7:
@petercolombo there was a bug in reading user-generated profile.

it got fixed (https://github.com/monome/norns/pull/1697) but isn’t available yet a norns firmware release.

removing the generated profile under /home/we/dust/data/gamepads should unbrick your norns.

also the profile builder script currently has a small bug where it doesn’t save one info in the generated profile (guid of the controller). i discovered the issue last week while debugging w/ another user but didn’t get the time to fix it yet.

it’d be cool to have this specific model of controller working as there should be quite a few of them in circulation.

petercolombo:
@p3r7 The make and model are Sony SCPH-1000R - there are definitely more than a few out there, I'm sure.

PR #1441, opened 2021-09-24
fix crash when trying to detect crow

nihilazo:
In some cases, `udev_device_get_property_value` can return a NULL. This causes a segmentation fault when trying to detect crow on some setups. Adding a check that the value isn't null fixes this.


catfact (approved):

catfact:
thanks!

PR #1442, opened 2021-09-28
Configurable OSC remote port.

winder:
Parameterize a magic number, and put it in the boot log for extra visibility.


catfact (approved):
looks OK to me!

PR #1443, opened 2021-09-28
Add select_values method to tabutil lib

nattog:
This filter creates a new table based on the passing the existing table values through a condition - similar to JS Array.filter method

### DIFF
--- a/lua/lib/tabutil.lua
+++ b/lua/lib/tabutil.lua
@@ -237,13 +237,13 @@ end
 
 --- Create a new table with all values that pass the test implemented by the provided function.
 -- @tparam table t table to check
--- @param condition callback function that tests all values of provided table, takes value and key as argument
+-- @param condition callback function that tests all values of provided table, passes value and key as arguments
 -- @treturn table table with values that pass the test
 tab.select_values = function(tbl, condition)
   local t = {}
 
   for k,v in pairs(tbl) do
-    if condition(v, k) then t[k] = v end
+    if condition(v,k) then t[k] = v end
   end
 
   return t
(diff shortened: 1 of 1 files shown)


catfact:
thanks! leaving some inline feedback

lua/lib/tabutil.lua
```diff
@@ -235,5 +235,18 @@ function tab.readonly(params)
...
 end
 
+--- Create a new table with all values that pass the test implemented by the provided function.
+-- @tparam table t table to check
+-- @param condition callback function that tests all values of provided table
+-- @treturn table table with values that pass the test
+function tab.filter(tbl, condition)
+  local t = {}
+
+  for i=1, #tbl do
```
ngwese, line 245: given that tables in lua are a generalized associative structure which is more flexible than the Array type in Javascript it might be more appropriate to use `pairs(...)` to iterate over the table since it will address all values in the table not simply those addressed via positive number or be more explicit in the comment about the expected shape of the table. the `#` operator can produce unexpected results depending on the table:
- http://lua-users.org/wiki/LengthsInLua
- https://www.lua.org/manual/5.1/manual.html#2.5.5

consider the following:
```lua
> t = {1, 2, nil, nil, 5}
> #t
5
> for i,v in ipairs(t) do print(i, v) end
1	1
2	2
> for i,v in pairs(t) do print(i, v) end
1	1
2	2
5	5
> for i=1,#t do print(i, t[i]) end
1	1
2	2
3	nil
4	nil
5	5
```
...its not immediately clear which type of iteration one would association with a function called `filter`. looking at the implementation `select_values` might be a more descriptive name.

ngwese: not sure if @catfact had other thoughts?

catfact:
no, that's all I had.

nattog:
Appreciate the feedback! I had initially overlooked some of different use cases with table keys. I've implemented the suggested changes, updating function name to select_values and using `pairs` to iterate.

tehn (commented):
newest version looks good to me

catfact (approved):

PR #1444, opened 2021-09-29
Decouple fb merge

catfact:
this is a reconciliation of PR #1221 on top of current `main`. seems to run on desktop but haven't done any real testing of scripts there. i have used it a bunch on norns.

if we want this functionality (draw to SDL and remap GPIO) then i'd vote for doing it this way (as opposed to the other, much simpler PR exposing an SDL surface.)


matron/src/hardware/input/sdl.c
```diff
@@ -0,0 +1,121 @@
...
+        fprintf(stderr, "got SDL input %d\n", event.type);
+        switch (event.type) {
+            case SDL_KEYDOWN:
+            case SDL_KEYUP: {
+                fprintf(stderr, "key event\n");
+                int z = event.type == SDL_KEYDOWN ? 1 : 0;
+                switch (event.key.keysym.sym) {
+                    case SDL_SCANCODE_Q:
+                        ev = event_data_new(EVENT_KEY);
+                        ev->key.n = 0;
```
tehn, line 58: got this to work by shifting to index 1, not 0

matron/src/hardware/input/sdl.c
```diff
@@ -0,0 +1,121 @@
...
+    while (true) {
+        SDL_WaitEvent(&event);
+        fprintf(stderr, "got SDL input %d\n", event.type);
+        switch (event.type) {
+            case SDL_KEYDOWN:
+            case SDL_KEYUP: {
+                fprintf(stderr, "key event\n");
+                int z = event.type == SDL_KEYDOWN ? 1 : 0;
+                switch (event.key.keysym.sym) {
+                    case SDL_SCANCODE_Q:
```
tehn, line 56: got this working by changing to `SDLK_q` etc for all matches

tehn (commented):
nice work--- super weird seeing the small screen! (yes it'd be sane to stretch the pixels... my initial attempt failed as the surface arrangement is different than my other SDL project)

having SDL polls (input when SDL window is focused) is a bit fiddly. some inline comments re: what i had to do to make it work. keyboard interfacing seems problematic given there are scripts that want a full keyboard (ie orca), so stealing the keyboard for just enc/keys seems like not the right move. perhaps a separate OSC controller app would suffice.

or a better approach might be to just forward the key and mouse data to virtual HID entries in the vport system: for example, when i plug in an external HID keyboard (to my laptop) i can nav the menu with the HID-driven arrow keys (like i can with norns).

another observation--- E1 nav has sensitivity, so you need to hit the W key 5 times to move the screen

i'd be up for improving the menu nav via HID keyboard as a solution

---

some random notes for future testers:

- norns has to be run from `~/norns`
- `~/dust/` must exist (though i'm still getting some errors)
- `/norns/sc/install.sh` must be run

matron/src/hardware/input/sdl.c
catfact: yeah seems like basically we should not use the SDL event loop..

okyeron:
> super weird seeing the small screen! (yes it'd be sane to stretch the pixels... my initial attempt failed as the surface arrangement is different than my other SDL project)

FWIW -  I was able to make a larger SDL window by scaling up the widow size in `SDL_CreateWindow`... and  then using the following instead of `SDL_BlitSurface(priv->draw_surface, NULL, priv->window_surface, NULL);`
```
SDL_Rect stretchRect;
stretchRect.x = 0;
stretchRect.y = 0;
stretchRect.w = 512;
stretchRect.h = 256;
SDL_BlitScaled( priv->draw_surface, NULL, priv->window_surface, &stretchRect );
```

I've not tested on this branch yet, but it seemed to work well on that other PR.

tehn (approved):

tehn:
i'm up for merging this, then following up with removing the SDL events and increasing the screen size.

on-device testing seems good and i can post a new pre-release for wider testing.

catfact:
well heck I'll just put in those changes, not a big deal

tehn:
i'm happy to do it, within my capabilities and you've got enough more fun stuff to think about

catfact:
woop sorry

tehn:
dang that was quick

PR #1445, opened 2021-09-29
make lamonte's intonation monotonic

catfact:
very small change. the ratio set in the `lamonte` tuning was non-monotonic. i'm not actually sure if that was a typo or just correctly insane. anyways this changes that so it is less crazy.


ngwese (approved):
confirmed in REPL, that would have been interesting indeed.

PR #1451, opened 2021-10-04
watcher reset, menu passwd entry

tehn:
- adds `SYSTEM > PASSWORD` which allows entry of new password for user `we` (does not require existing password, no confirmation, no display of current password)
- `watcher` now checks for `/boot.norns.txt`, if present resets pw to `sleep` and deletes `norns.txt` (this file can be created by mounting a norns in DISK mode via `rpiboot` or by inserting shield sdcard into computer) --- this is a failsafe in case the menu system won't load due to a broken update (or something)... the theory being that `watcher` is less fragile.

bonus (i guess): `SYSTEM` menu is now scrolling

### DIFF
--- a/lua/core/menu/system.lua
+++ b/lua/core/menu/system.lua
@@ -47,8 +47,8 @@ m.deinit = norns.none
 
 m.passdone = function(txt)
   if txt ~= nil then
-    print("password change to " .. txt)
-    os.execute("echo 'we:"..txt.."' | sudo chpasswd")
+    local status = os.execute("echo 'we:"..txt.."' | sudo chpasswd")
+    if status then print("password changed") end
   end
   _menu.set_page("SYSTEM")
 end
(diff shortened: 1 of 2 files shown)


lua/core/menu/system.lua
```diff
@@ -21,21 +27,31 @@ end
...
   screen.update()
 end
 
 m.init = norns.none
 m.deinit = norns.none
 
+m.passdone = function(txt)
+  if txt ~= nil then
+    print("password change to " .. txt)
+    os.execute("echo 'we:"..txt.."' | sudo chpasswd")
```
ngwese, line 51: suggest checking that return value of `os.execute(...)` and confirming that the command did succeed then simply printing out whether the password change succeeded or failed. print the actual password in cleartext would end up writing it to the system logs and over the network (unencrypted) if maiden is open.

ngwese (changes requested):

lua/core/menu/system.lua
tehn: lol i guess forgotten prints are a common security hole

ngwese (approved):

Issue #1212, opened 2020-10-22
favorites: removed/renamed script remains starred at top


dndrks:
some small weirdness with 'favorites' -- if a starred script is removed or renamed, its entry in the favorites table persists, which creates a ghost entry in the favorites section.

unsure what would be the best way to address, but perhaps a quick read on the favorites table to check whether the script path is still valid (and if invalid, remove the entry) could be enough? what do you think @zzsnzmn ?

tyleretters:
i could hit this

tehn:
this sounds good--- but i also think calling `build_favorites()` on every
SELECT menu might be overkill. i think calling it once at startup would be
sufficient, in case someone deleted some stuff in maiden then shut down. it
seems improbable that someone would intentionally delete something in
maiden and then be confused by its presence in favorites.

my main point here is to not add even more processing time to SELECT, which
already has a substantial load time if you have a lot of scripts.

tyleretters:
i don't think a `validate_favorites()` at startup would solve the bug. my understanding is the code already works that way. it builds the table on startup and saves it to memory. there's no cache or persistence.

i believe the scenario is specifically for when, "someone would intentionally delete something in
maiden and then be confused by its presence in favorites." @dndrks can you confirm?

tehn:
lol, i wrote "major" and meant "minor"

PR #1460, opened 2021-10-14
check for script before launching

tyleretters:
for #1212

following @tehn's guidance. tried to use as little code as possible.


tehn (approved):
thank you for this.

it breaks the pattern of having screen code outside the redraw function, but i've also implemented some really dirty state behavior (ie `m.len = 'scan'` wtf) so adding more state behavior is going to make it even messier without a bigger refactor (which isn't hard, so i can do this after merge sometime in the next weeks...)

tyleretters:
yeah!

i'd enjoy to screenshare with you while you refactor so i can learn the ways

PR #1461, opened 2021-10-14
verbose 'missing' message

dndrks:
'missing' errors don't give the user much in the way of next steps if they're new to the ecosystem -- even though troubleshooting is all well-documented and maiden has a heads-up built in, folks still ask how to resolve the error (spun this up after seeing a message posted to the study discord this morning).

this commit would add two bits of helper text, one on-screen and the other printed to the REPL.

**on-screen**, user would see:

```
error: missing ____
try 'SYSTEM > RESTART'
SELECT >
SYSTEM >
SLEEP >
```

**in REPL**, user would see:

```
loading engine: ____
### SCRIPT ERROR: missing ____
### try 'SYSTEM > RESTART'
```


tehn (commented):
does this work when home screen has extra info mode enabled?

PR #1464, opened 2021-10-16
watcher pw reset msg

tehn:
fixes error (warning) and gives some minor feedback


ngwese (approved):

ngwese:
irony is that reviewing this revealed that my dev device has never been running watcher since i've never run `update.sh` apparently.  the added `sprintf(...)` would show in the logs for the unit so this will help in debugging.

PR #1465, opened 2021-10-19
add rerun()

tyleretters:
started doing this in `lua/lib/devtools.lua` but that `require()`d too much effort on the dev side. scanning all the files already included this felt like the most logical place. open to feedback!


tyleretters:
updated!

tehn (approved):

PR #1468, opened 2021-10-24
lib/lattice: add swing option

schollz:
this is a PR to add "swing" to the lattice library, at the pattern level. these changes have already been implemented in thirtythree awhile ago and it has be working well, but I thought it would be useful to have this feature in the core library.

I don't know the idiomatic way to do swing so I followed the swing attributes like in the OP-1, where swing takes a percentage. a swing percentage of 50% means no swing. above 50% is the swing where the first beat has the long rest:

 ```
1   2 3   4
```

below 50% the swing causes the first beat to have the shorter rest:

```
1 2     3 4
```

the max/min distinguishable "swing" really depends on the ppqn and on the current division. on the OP-1 it limits you to between 20% and 80%. I didn't do any limiting here (though its clamped to be between 0 and 100%), but above ~70% the swing is indistinguishable and it reduces to a division half of what is intended (at least at 96 ppqn and 1/4 division).

swing is added at the pattern level with a new function `Pattern:set_swing` where the swing can be set on the fly, staying in sync with all the patterns. so you can set the swing to something and set it back to 50% and it will still be in tempo with a non-swinging pattern.

the default swing is 50% so it shouldn't affect the default behavior of lattice.

I would appreciate any feedback on the implementation. 

here's a little test script I've used to test it:

```lua
-- a

engine.name="PolyPerc"

function init()

  lattice = require("lattice")

  -- default lattice usage, showing default arguments
  my_lattice = lattice:new{}

  my_patterns={}
  my_patterns[1]=my_lattice:new_pattern{
    action = function(t)
      engine.amp(0.5)
      engine.hz(110)
    end,
    division = 1/4,
  }  
  my_patterns[2]=my_lattice:new_pattern{
    action = function(t)
      engine.amp(0.2)
      engine.hz(440)
    end,
    division = 1/4,
  }

  -- start the lattice
  my_lattice:start()

  -- swing controller
  for i=1,2 do 
    params:add{type="control",id=i.."swing",name="swing "..i,
      controlspec=controlspec.new(0,100,'lin',0.1,50,'%',0.1/100),action=function(v)
        my_patterns[i]:set_swing(v)
    end}
  end

  clock.run(redraw_clock)
end

function redraw_clock() -- our grid redraw clock
  while true do -- while it's running...
    clock.sleep(1/10) -- refresh
    redraw()
  end
end


function key(k,z)
  if z==0 then 
    do return end 
  end
  if k==2 then 
    my_lattice:reset()
  elseif k==3 then 
    my_lattice:start()
  end
end

function enc(k,d)
  if k>1 then 
    params:delta((k-1).."swing",d)
  else
    for k=2,3 do 
      params:delta((k-1).."swing",d)
    end
  end
end

function redraw()
  screen.clear()
  screen.level(15)
  screen.move(10,10)
  screen.text("pattern 1 swing: "..params:get("1swing").."%")
  screen.move(10,20)
  screen.text("pattern 2 swing: "..params:get("2swing").."%")
  screen.move(10,30)
  screen.text("K2 stops, K3 starts")
  screen.move(10,40)
  screen.text("E1 adjusts both swings")
  screen.move(10,50)
  screen.text("E2/E3 adjust pattern swing")
  screen.update()
end
```

(ping @tyleretters)

### DIFF
--- a/lua/lib/lattice.lua
+++ b/lua/lib/lattice.lua
@@ -1,7 +1,7 @@
 --- module for creating a lattice of patterns based on a single fast "superclock"
 --
 -- @module Lattice
--- @release v1.1.0
+-- @release v1.2.0
 -- @author tyleretters & ezra & zack
 
 local Lattice, Pattern = {}, {}
@@ -44,6 +44,7 @@ function Lattice:reset()
   end
   for i, pattern in pairs(self.patterns) do
     pattern.phase = pattern.division * self.ppqn * self.meter
+    pattern.downbeat = false
   end
   self.transport = 0
   params:set("clock_reset",1)
@@ -96,9 +97,14 @@ function Lattice:pulse()
     for id, pattern in pairs(self.patterns) do
       if pattern.enabled then
         pattern.phase = pattern.phase + 1
-        if pattern.phase > (pattern.division * ppm) then
+        local swing_val = (2*pattern.swing/100)
+        if not pattern.downbeat then 
+          swing_val = (2*(100-pattern.swing)/100)
+        end
+        if pattern.phase > (pattern.division * ppm)*swing_val then
           pattern.phase = pattern.phase - (pattern.division * ppm)
           pattern.action(self.transport)
+          pattern.downbeat = not pattern.downbeat
         end
       elseif pattern.flag then
         self.patterns[pattern.id] = nil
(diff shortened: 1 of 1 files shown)


tyleretters (approved):
absolutely love it. sounds awesome. thank you for the test script. really really nice work. :)

Issue #1376, opened 2021-06-03
clock: Link workflow desires


dndrks:
an attempt to collect disparate chunks of conversation about Link from Discord, Slack and Github.

1. ideally, norns should not announce itself as a Link peer until Link is selected in `CLOCK > source`
2. if norns is joining an existing Link session as a new peer, it should not push its tempo to other peers -- rather, it should inherit the Link session's tempo
3. if norns is the sole member of a Link session, it should push its tempo to new peers
4. if there is no Link session established and norns clock is switched _to_ Link, the last source's tempo should carry over
5. norns should be able to send Link start/stop commands to other peers

most of this is also covered in the [Link test outline](https://github.com/Ableton/link/blob/master/TEST-PLAN.md#tempo-2-opening-an-app-with-link-enabled-should-not-change-the-tempo-of-an-existing-link-session), but figured it'd be nice to have a central ticket to chat!

dndrks:
additional data point: since norns is established as a Link peer regardless of clock source, norns automatically pushes its tempo to all other Link peers if the `clock > source` selector passes over the `link` entry -- notably, on the way to `crow`.

we _could_ swap the order of the clock sources, so that `link` is last in the table

tehn:
alt solution: for option type, K3 creates a LISTSELECT which allows for a scroll/confirm without activating intermediary values

Issue #1374, opened 2021-06-02
clock beta: tempo jumps to 120 when changing sources


dndrks:
current behavior:
- set norns to internal clock at 90bpm
- become inspired to switch to Link to sync a DAW or app
- since the option layout for `CLOCK > source` is `internal`,`midi`,`link`,`crow`, I'll often hit `midi` before `link`
- switching from `internal` to `midi` causes the tempo to jump up to 120bpm
- now, when i switch from `midi` to `link`, I carry over the 120bpm

if I'm fast + precise, I _can_ sneak past `midi` to `link` without my tempo changing, but I'm often neither fast nor precise :)

i guess some super-low hanging fruit would be to just switch `midi` and `link`'s positions in the menu, so that the shift from `internal` to `link` can carry over the current `internal` bpm? but i can't tell how much infrastructure might rely on indexes from `params:get("clock_source")`...

artfwo:
(changed the title as it's not a beta-specific problem)

midi clock basically has a fallback tempo of 120 bpm - something we require for the sync mechanism to work and every source technically keeps track of its own tempo.

but we also have a [source tempo follower](https://github.com/monome/norns/blob/main/lua/core/clock.lua#L291) that updates clock_tempo param every second to keep the internal source tempo in sync with midi/crow (see #1097 and #1100 for reference).

one solution would be to remove it completely to let each source live with its own tempo and only allow updates to internal and link tempo when changing the param, wdyt?

dndrks:
ah super good thoughts!! thank you artem <3

> but we also have a source tempo follower that updates clock_tempo param every second to keep the internal source tempo in sync with midi/crow (see #1097 and #1100 for reference).

I can't recall, but this was to avoid a jarring tempo change when (for example) the artist switches from `MIDI` to `internal`, yeah?

----

_nb. I apologize if any of the following is annoying / off-base, but figured it'd be helpful to share_

> one solution would be to remove it completely to let each source live with its own tempo and only allow updates to internal and link tempo when changing the param, wdyt?

I think the ideal situation might require approaching the `source` more flexibly than we might want to. for example, if I set `source` to `MIDI`, i can't change the tempo from norns -- the new tempo _has_ to come from an external device.

this is a departure from what folks might expect working with DAWs or MIDI devices, which typically follows:
- I have an "internal" tempo of 96bpm
- I enable external syncing, but the tempo remains 96bpm until the DAW/device receives an external clock signal (let's say at 130bpm)
[...]

PR #1475, opened 2021-11-08
Joining an existing Link session should adopt the session tempo, not force the device tempo

Dewb:
### background/motivation

The [Link test plan](https://github.com/Ableton/link/blob/master/TEST-PLAN.md) has two clauses about the expectations for how joining a session should affect the app/device tempo:

[TEMPO-2](https://github.com/Ableton/link/blob/master/TEST-PLAN.md#tempo-2-opening-an-app-with-link-enabled-should-not-change-the-tempo-of-an-existing-link-session): Opening an app with Link enabled should not change the tempo of an existing Link session.
[TEMPO-5](https://github.com/Ableton/link/blob/master/TEST-PLAN.md#tempo-5-enabling-link-does-not-change-apps-tempo-if-there-is-no-link-session-to-join): Enabling Link does not change app's tempo if there is no Link session to join.

So the behavior needs to be different based on whether the session has other participants or not. Luckily, this behavior is already implemented in the Link library; in order to take advantage of it, we just need to ~~construct a new Link object when enabling Link~~ manipulate the enable state, and make sure tempo requests are processed before we change the enable state.

### primary change: enable/disable Link when it is selected/deselected as a source

The existing behavior is to create and join a Link session at norns startup, and then switch it to be the clock source (and set the tempo) when Link is enabled from the params menu. The primary change in this PR is to only enable the Link client when Link becomes the clock source, and to send the tempo change request before Link becomes the clock source.

This has a nice consequence in that other Link clients that report number of connected devices will now only count norns when it's using Link for sync, rather than any time it's on the network regardless of clock source.

~~This also has a not-nice consequence: the Link "start/stop sync" feature, which norns previously always enabled, now exhibits behavior that doesn't seem correct or desirable. With start-stop sync always enabled, it's easy to get to a state where you have no way of re-enabling transport from norns, you have to kick the session from another client (or run `_norns.clock.start()` from maiden.)~~

### ~~secondary change: allow user to toggle Link start/stop sync~~

~~To mitigate this, this PR also exposes the Link 3.0 start/stop sync enable/disable state in the menu, and defaults it to disabled. This way Link behaves more like the other clock sources, and it's harder to inadvertently stop the transport by switching sources.~~

~~Start-stop sync is currently tricky to implement correctly because the system does not currently have a clock-source-independent, script-independent notion of whether the global transport is running or not. Having such a state isn't an explicit requirement for Link clients, but reading between the lines of the test plan, that assumption does seem to be in there. A follow-up PR could add a global transport state to cover the edge cases when switching Link on and off with start/stop sync enabled.~~

### DIFF
--- a/matron/src/clocks/clock_link.c
+++ b/matron/src/clocks/clock_link.c
@@ -16,6 +16,7 @@ static struct clock_link_shared_data_t {
     double quantum;
     double requested_tempo;
     bool playing;
+    bool enabled;
     pthread_mutex_t lock;
 } clock_link_shared_data;
 
@@ -31,7 +32,6 @@ static void *clock_link_run(void *p) {
     link = ableton_link_new(120);
     clock = ableton_link_clock(link);
 
-    ableton_link_enable(link, true);
     ableton_link_enable_start_stop_sync(link, true);
 
     while (true) {
@@ -63,6 +63,8 @@ static void *clock_link_run(void *p) {
                 clock_link_shared_data.requested_tempo = 0;
             }
 
+            ableton_link_enable(link, clock_link_shared_data.enabled);
+
             ableton_link_session_state_destroy(state);
             pthread_mutex_unlock(&clock_link_shared_data.lock);
         }
@@ -83,8 +85,21 @@ void clock_link_start() {
 
     clock_link_shared_data.quantum = 4;
     clock_link_shared_data.requested_tempo = 0;
+    clock_link_shared_data.enabled = false;
 
     pthread_create(&clock_link_thread, &attr, &clock_link_run, NULL);
+}
+
+void clock_link_join_session() {
+    pthread_mutex_lock(&clock_link_shared_data.lock);
+    clock_link_shared_data.enabled = true;
+    pthread_mutex_unlock(&clock_link_shared_data.lock);
+}
+
+void clock_link_leave_session() {
+    pthread_mutex_lock(&clock_link_shared_data.lock);
+    clock_link_shared_data.enabled = false;
+    pthread_mutex_unlock(&clock_link_shared_data.lock);
 }
 
 void clock_link_set_quantum(double quantum) {
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -183,13 +183,13 @@ function clock.add_params()
     norns.state.clock.source)
   params:set_action("clock_source",
     function(x)
+      if x==3 then clock.link.set_tempo(params:get("clock_tempo")) end -- for link, apply tempo before setting source
       clock.set_source(x)
       if x==4 then
         norns.crow.clock_enable()
       end
       norns.state.clock.source = x
-      if x==1 then clock.internal.set_tempo(params:get("clock_tempo"))
-      elseif x==3 then clock.link.set_tempo(params:get("clock_tempo")) end
+      if x==1 then clock.internal.set_tempo(params:get("clock_tempo")) end
     end)
   params:set_save("clock_source", false)
   params:add_number("clock_tempo", "tempo", 1, 300, norns.state.clock.tempo)
--- a/matron/src/clock.c
+++ b/matron/src/clock.c
@@ -149,6 +149,12 @@ void clock_reschedule_sync_events_from_source(clock_source_t source) {
 }
 
 void clock_set_source(clock_source_t source) {
+    if (clock_source != source && source == CLOCK_SOURCE_LINK) {
+        clock_link_join_session();
+    } else if (clock_source != source && clock_source == CLOCK_SOURCE_LINK) {
+        clock_link_leave_session();
+    }
+
     clock_source = source;
     clock_scheduler_reschedule_sync_events();
 }
--- a/matron/src/clocks/clock_link.h
+++ b/matron/src/clocks/clock_link.h
@@ -2,6 +2,8 @@
 
 void clock_link_init();
 void clock_link_start();
+void clock_link_join_session();
+void clock_link_leave_session();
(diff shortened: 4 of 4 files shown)


Dewb:
I think this covers 1-4 in #1376. Item 5 would be a good fit with that hypothetical followup that adds a global transport state.

Dewb:
it's also possible that I'm missing something and start/stop sync is totally possible to implement ergonomically without global transport state!

artfwo (changes requested):
it should not be necessary to expose join_session/leave_session in the API, as well as initializing clock in clock_link_run to solve these issues:

> 
> TEMPO-2: Opening an app with Link enabled should not change the tempo of an existing Link session.
> TEMPO-5: Enabling Link does not change app's tempo if there is no Link session to join.

link already does this internally, but it doesn't quite work in norns because of the [global tempo follower](https://github.com/monome/norns/blob/main/lua/core/clock.lua#L281) that basically synchronizes internal tempo with midi/link/crow.

a less complicated solution would be removing the tempo follower altogether (as it's broken anyway) and replacing it with fallback tempo values for midi and crow, so setting the clock tempo parameter will only change fallback values for those sources. with this, we won't need to track the tempo and will be able to piggyback on link's default behaviors without additional API changes (also see https://github.com/monome/norns/issues/1374 for discussion and some context).

happy to review the start/stop sync change separately, could you please bring it up as a different PR?

Dewb:
> it should not be necessary to expose join_session/leave_session in the API, as well as initializing clock in clock_link_run to solve these issues:

Ah yes, I see the join behavior is actually triggerable through Controller::enable, and there was no need to expose join/leave to the lua layer. I can definitely simplify this quite a bit.

Thanks, I hadn't seen #1374 and I wasn't aware the tempo follower behavior was implicated in this issue. I don't quite follow how it's preventing us from letting Link decide how to handle the join tempo though? It looks like the fundamental issue is that changing the clock source in params calls `clock.link.set_tempo()`, which gets passed to the Link client as if it were a user-initiated change, which it is not.

artfwo:
> Thanks, I hadn't seen #1374 and I wasn't aware the tempo follower behavior was implicated in this issue. I don't quite follow how it's preventing us from letting Link decide how to handle the join tempo though? It looks like the fundamental issue is that changing the clock source in params calls `clock.link.set_tempo()`, which eventually gets passed to the Link client as if it were a user-initiated change, which it is not.

you're right, in this case it doesn't (sorry, I haven't had coffee yet :)) but I think this code should be removed too.

generally, we should let link handle all the joining internally, as it implements the expectations correctly by default and maybe decouple the clock sources a bit to let that happen.

Dewb:
Drastically simplified the changes. Yes, most of that stuff was unnecessary! Manipulating the enable state and making sure we set the local tempo before enabling Link is enough to satisfy TEMPO-2 and TEMPO-5, and have the correct number of peers show up in other clients.

I'm happy to keep iterating on this, or if you ultimately don't think this direction is promising, I can trim it down to just the start/stop state sync exposure.

artfwo:
> I'm happy to keep iterating on this, or if you ultimately don't think this direction is promising, I can trim it down to just the start/stop state sync exposure.

Yes, please go ahead. I'm going to have a closer look at the changes later this week, but a quick question for now - is clock start/stop sync preserved correctly after norns is reset?

Dewb:
> is clock start/stop sync preserved correctly after norns is reset?

A sleep/reboot cycle preserves the state, yes. `SYSTEM > RESET` resets it to default (disabled.)

Dewb:
Split this into 2 PRs, this one for the join behavior, the other for adding the parameter for start/stop sync.

matron/src/clocks/clock_link.c
```diff
@@ -63,6 +63,8 @@ static void *clock_link_run(void *p) {
                 clock_link_shared_data.requested_tempo = 0;
             }
 
+            ableton_link_enable(link, clock_link_shared_data.enabled);
```
artfwo, line 66: this is a [thread-safe method](https://github.com/Ableton/link/blob/master/include/ableton/Link.hpp#L98), so technically we don't need to pass the flag through clock_link_shared_data and just call ableton_link_enable with the appropriate parameter from clock_set_source.

Dewb: The nice thing about having the `ableton_link_enable()` call inside the mutex-guarded link thread is that we can guarantee that any pending tempo change requests are processed before we call enable.

artfwo: how does mutex ensures that? the only thread where these requests could come from is the single matron thread, so there can't be any race conditions around tempo.

Dewb: this is totally theoretical, just thinking out loud, but:
* change clock source from midi to link, param handler for clock_source runs
* handler calls `clock.link.set_tempo()`. link thread is sleeping so this grabs mutex, updates shared data structure, and releases mutex.
* link thread wakes up and starts running
* handler next calls `clock.set_source` and directly sets enable state
* link thread reaches the block `if (clock_link_shared_data.requested_tempo > 0)` and processes tempo change. since link has already been enabled, this is treated as a user-initiated tempo change and if there are other clients in the session, this violates the TEMPO-2 testcase.

Dewb: in other words, we need the disabled -> enabled transition to happen here on line 66, after the tempo change request has been committed.

artfwo (approved):
thanks, i think the approach described in https://github.com/monome/norns/issues/1374#issuecomment-854067186 could be implemented as a follow-up, so let's merge this first.

PR #1476, opened 2021-11-10
add clock param for link "start/stop sync" option

Dewb:
Separating this addition out from #1475.

An issue that this PR does not yet fix: the link clock source state in `clock_link_shared_data.playing` is initialized to false and is not kept in sync with any script-level notions of whether the transport is running. So if you load a script that starts playing by default (like awake) and join a running link session, any stop messages will be ignored until link sends a start message. (If we instead initialized `.playing` to true, we'd have the same problem with scripts that didn't start playing on init.) 

To fix that issue, I think there needs to be a script-independent, clock-source-independent flag somewhere indicating whether transport is running (e.g. the script, user, or events have triggered `clock.start()` more recently than `clock.stop()`.)


artfwo (approved):
looks good, but it's marked as draft. @Dewb would you like to change something before this is merged?

Dewb:
I think this is ready as an individual feature, but it exacerbates the issue of script playing state getting decoupled from `clock_link_shared_data.playing`. So it could be held until we have a solution to that. On the other hand, this is relatively simple and is still useful despite the bigger, complex issue, so maybe they shouldn't be tied together.

artfwo:
> I think this is ready as an individual feature, but it exacerbates the issue of script playing state getting decoupled from `clock_link_shared_data.playing`. So it could be held until we have a solution to that. On the other hand, this is relatively simple and is still useful despite the bigger, complex issue, so maybe they shouldn't be tied together.

Since we don't have well-defined transport states for scripts (do we even need those?), I'd suggest just proceeding if this has a practical use-cases, such as disabling link restart in case a script transport callback is being triggered.

PR #1477, opened 2021-12-05
Add Tilt support for legacy grids

okyeron:
Adds support for tilt.

Tested on norns shield and a grayscale 64 (2010-2012) - serial m64-0993  
grid presents as `grid monome 64 m64-0993 8x8`

For testing I added tilt to my grid-test script in this branch: https://github.com/okyeron/grid-test/tree/tilt
(tilt can be enabled/disabled in the params menu)

Results:	 
* X/Y data comes through, no Z (although no Z data was observed on Max test-grid patch either).
* Values displayed are between ~90 and ~170, with ~130 in the center (grid flat on table)

Questions:  
* Is there no Z on that grid model? Is there Z on others?
* What are expected values? Seems like 0 to <200 when testing with the Max test-grid patch

### DIFF
--- a/matron/src/device/device_monome.c
+++ b/matron/src/device/device_monome.c
@@ -183,7 +183,7 @@ void dev_monome_handle_tilt(const monome_event_t *e, void *p) {
     ev->grid_tilt.sensor = e->tilt.sensor;
     ev->grid_tilt.x = e->tilt.x;
     ev->grid_tilt.y = e->tilt.y;
-    ev->grid_tilt.z = e->tilt.y;
+    ev->grid_tilt.z = e->tilt.z;
     event_post(ev);
 }
 
--- a/matron/src/device/device_monome.c
+++ b/matron/src/device/device_monome.c
@@ -21,6 +21,7 @@ static void dev_monome_handle_lift(const monome_event_t *e, void *p);
 static void dev_monome_handle_encoder_delta(const monome_event_t *e, void *p);
 static void dev_monome_handle_encoder_press(const monome_event_t *e, void *p);
 static void dev_monome_handle_encoder_lift(const monome_event_t *e, void *p);
+static void dev_monome_handle_tilt(const monome_event_t *e, void *p);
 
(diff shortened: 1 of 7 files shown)


ngwese:
One quick note, this PR appears to include what appear to be unrelated updates to the `softcut` and `link` submodules. It is preferable to have the submodules remain the as is (admittedly we haven't done a great job of that in the past.)

okyeron:
Apologies, I noticed the softcut/link items after the fact.

I'm not entirely sure what I did that made those include here - aside from doing `git submodule update --init --recursive` on the source from the current norns head prior to making these changes.

Is there a good way to remove those references from this PR?

ngwese:
I think you can checkout the previous commits for the `softcut` and `link` modules then `git add` those modules and `git commit --amend`. You likely need to `git push --force` to this PR branch as Github will likely complain because your local and this remote branch will have diverged as a result of using `--amend`.

The result should be a single commit which no longer shows the submodule changes.

matron/src/device/device_monome.c
```diff
@@ -166,6 +176,17 @@ void dev_monome_handle_lift(const monome_event_t *e, void *p) {
...
 }
 
+void dev_monome_handle_tilt(const monome_event_t *e, void *p) {
+    struct dev_monome *md = (struct dev_monome *)p;
+    union event_data *ev = event_data_new(EVENT_GRID_TILT);
+    ev->grid_tilt.id = md->dev.id;
+    ev->grid_tilt.sensor = e->tilt.sensor;
+    ev->grid_tilt.x = e->tilt.x;
+    ev->grid_tilt.y = e->tilt.y;
+    ev->grid_tilt.z = e->tilt.y;
```
tehn, line 186: typo here? tilt.y repeated

tehn (changes requested):
this is great, thank you!

i totally don't remember designing the tilt protocol for multiple sensors... had to dig back

tehn (approved):

PR #1479, opened 2021-12-09
V3 wip/compile with cpp

catfact:
### motivation

we would like to consolidate the audio backend and the interpreter into the same process, so that we eliminate the use of OSC as a local IPC layer. (and enable fun things like accessing softcut buffers directly from lua.)

### solution

- changed all `matron` sources to use `.cc` extension. this encourages `waf` to build them with the loaded c++ toolchain, but also distinguishes them from "actual" c++ code which continues to be suffixed with `.cpp`.

- fixed incompatible code (implicit conversions and field initializations.)

### issues

it would be a little less weird looking to keep matron sources as .c and just compile them with g++. there was not an obvious way to do this with `waf` without editing environment variables. (telling it not to load a C toolchain didn't work, which seems like a bug.)


catfact:
for these v3-wip changes (which build on each other) i'm going to go ahead and merge to the v3 branch without approval. (PR is  just to keep a relatively organized view of the history.)

tehn (commented):
thanks for making this trackable, looks good. must've been a lot of type warnings

PR #1480, opened 2021-12-09
consolidate crone/matron processes

catfact:
### motivation

combine processes, as described in PR #1479 

### solution

- add new target `norns/wscript`, combining sources from `matron/src` and `crone/src`, producing executable at `build/norns/norns`
- small necessary edits to component `main` sources / routines; otherwise sources are unmoved/unchanged

### status / issues

seems to mostly work, with a big caveat: connection to sclang is pretty well broken since the order of process dependencies no longer applies and it is not possible to launch `sclang` between `crone` and `matron`.

fixing this will require some changes to sclang classes (which will be subject of a future PR to v3) and of course to systemd service definitions (which are not in this repo.)

### DIFF
--- a/crone/src/crone_main.h
+++ b/crone/src/crone_main.h
@@ -1,0 +1,8 @@
+#ifndef _CRONE_MAIN_H_
+#define _CRONE_MAIN_H_
+
+
+int crone_main();
+void crone_cleanup();
+
+#endif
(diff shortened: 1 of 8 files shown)


tehn:
shall i make a `v3` branch of norns-image? the systemd changes should be trivial (famously said)

catfact:
i guess that would be a good idea.

PR #1481, opened 2021-12-09
V3 wip/rename sisters

catfact:
### motivation

avoid use of the word "crone," which has derogatory associations.

### solution

rename every occurence of `matron` and `crone` to `lachesis` and `atropos` respectively. 

### status / issues

i may have messed up one of the font resource files...

also have not yet touched the sclang classes since those are scheduled for a rewrite anyways.


catfact:
closing for now pending further decisions

PR #1483, opened 2021-12-15
fix alt variable

dndrks:
`alt` variable was not enclosed in the `m` table, which causes namespace collision with scripts which have `alt` in the script's scope.


tehn (approved):
good find, thank you!

PR #1484, opened 2021-12-18
No popen

tehn:


matron/src/system_cmd.cc
```diff
@@ -40,60 +31,19 @@ void system_cmd(char *cmd) {
...
-
     // this will get freed when the event is handled
-    ev->system_cmd.capture = cap;
+    ev->system_cmd.capture = buff;
     event_post(ev);
 
-    free(line);
-    free(capture);
-    pclose(f);
-
```
catfact, line 97: needs `free(buff)` ?

catfact: oh nvm, event handler frees eventually?

catfact (approved):

matron/src/system_cmd.cc
tehn: that was my rationale, yes.

thanks!

PR #1489, opened 2021-12-21
V3 wip/replace osc hack

catfact:
mostly complete replacement of OSC->crone with direct calls in `oracle.cc`

one tricky point concerns the use of parameter strings from lua. these are used in generic sofcut and FX param setters simply to avoid LOC in glue modules.  this made more sense with OSC as endpoint (matches on strings anyway). now it feels a little sillier, but for now instead of adding lots of LOC, added a dispatch module mapping `std::string` to setter funcs. have not completed this for FX params, only softcut params.

compiles but still quite untested. opening PR early for visibility.

### DIFF
--- a/crone/src/Commands.h
+++ b/crone/src/Commands.h
@@ -9,126 +9,124 @@
 
 namespace crone {
 
-class MixerClient;
-class SoftcutClient;
+    class MixerClient;
+    class SoftcutClient;
 
-class Commands {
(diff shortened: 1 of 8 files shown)


catfact:
added FX glue and fixed conflicts with base. seems finished, though i may be missing something... going ahead with merge on this one.

Issue #1491, opened 2021-12-24
screen_png bug


tyleretters:
hypothesis: calling screen_png clobbers screen_level values. https://llllllll.co/t/norns-display-graphics/22075/17

PR #1492, opened 2021-12-24
save/restore for png-display

tehn:
compiles, but untested. fixes https://github.com/monome/norns/issues/1491


tehn:
tested. fine/simple

PR #1493, opened 2021-12-27
add link transport activation from norns

schollz:
this is a PR to address number 5 on this list: https://github.com/monome/norns/issues/1376:

> norns should be able to send Link start/stop commands to other peers

I followed @Dewb's recent additions for implementation. the implementation  was simple enough - just bound @artfwo's wrapper for settings the transport and added it to the lua api. 

I tested it with norns<->norns and norns<->Ableton and both worked well with getting/setting transports.

happy to fix this for a better implementations / better api names if needed!

example demo: 

https://cdn.discordapp.com/attachments/825387870873190410/924811264335675412/linklink.mp4

example code:

```lua
-- link est

engine.name="PolyPerc"

lattice_=require("lattice")
beat=0
function init()

  lattice=lattice_:new()
  pattern=lattice:new_pattern{
    action=function(t)
      beat=math.floor(t/96)+1
      redraw()
      engine.hz(220)
      engine.pan(-1)
    end,
    division=1/4,
  }
  lattice:start()
  lattice:stop()
  beat=0

  redraw()
end

function key(k,z)
  if z==0 then 
    do return end 
  end
  if k==2 then 
    print("clock.link.stop()")
    clock.link.stop()
  elseif k==3 then 
    print("clock.link.start()")
    clock.link.start()
  end
end

function clock.transport.start()
  print("we begin")
  lattice:hard_restart()
end

function clock.transport.stop()
  print("we end")
  lattice:stop()
  beat=0
  redraw()
end

function redraw()
  screen.clear()
  screen.font_size(24)
  screen.move(64,32)
  screen.text_center(beat)
  screen.update()
end
```


artfwo (changes requested):
thanks, the low-level implementation is mostly correct, but the lua API is a bit counter-intuitive.

for the scripts, it would be better to have `clock.link.start()` and `clock.link.stop()` (or `clock.link.transport_start()` and `clock.link.transport_stop()` in the API instead and change the condition in `clock_link_run` accordingly, e.g. to something like:

```
if (clock_link_shared_data.transport_start) {
...
}

if (clock_link_shared_data.transport_stop) {
...
}
```

eventually, it would be nice to move commit_app_session_state call outside of the conditionals, but that could be done in a separate PR.

schollz:
that sounds better for sure! I updated the code for that and the example above.

matron/src/clocks/clock_link.c
```diff
@@ -41,6 +43,19 @@ static void *clock_link_run(void *p) {
...
+            if (clock_link_shared_data.transport_start) {
+                ableton_link_session_state_set_is_playing(state, true, 0);
+                ableton_link_commit_app_session_state(link, state);
+                clock_link_shared_data.transport_start=false;
+            }
+
+            if (clock_link_shared_data.transport_stop) {
+                ableton_link_session_state_set_is_playing(state, false, 0);
+                ableton_link_commit_app_session_state(link, state);
+                clock_link_shared_data.transport_stop=false;
```
artfwo, line 56:
```suggestion
                clock_link_shared_data.transport_stop = false;
```

matron/src/clocks/clock_link.c
```diff
@@ -41,6 +43,19 @@ static void *clock_link_run(void *p) {
             double link_tempo = ableton_link_session_state_tempo(state);
             bool link_playing = ableton_link_session_state_is_playing(state);
 
+
+            if (clock_link_shared_data.transport_start) {
+                ableton_link_session_state_set_is_playing(state, true, 0);
+                ableton_link_commit_app_session_state(link, state);
+                clock_link_shared_data.transport_start=false;
```
artfwo, line 50:
```suggestion
                clock_link_shared_data.transport_start = false;
```

artfwo (approved):

tehn (approved):

tehn:
@dndrks doc alert

PR #1495, opened 2021-12-27
lattice: patterns pulse in order based on id

schollz:
this gives lattice a more desired behavior to run its patterns in order of creation (i.e. in order of its pattern id).

this is really useful if you want to separate musical things into different patterns, i.e. one pattern for chord progression changes and one pattern for playing notes of the chord - in such case you would always want the chord progression change to occur before notes play. this can be easily accomplished by pulsing through patterns in order of their id (by default patterns are added with incrementing ids).

to emulate this, consider this lua test example:

```lua
-- lattice test

local lattice=require("lattice")

function init()
    local my_lattice = lattice:new()
    local shared_value=0
    first_pattern=my_lattice:new_pattern{
        action = function(t) shared_value=shared_value+100 end,
        division = 1
    }
    second_pattern=my_lattice:new_pattern{
        action = function(t) 
          shared_value=shared_value+1
          print(shared_value) 
          if shared_value==207 then 
            first_pattern:destroy()
          end
        end,
        division = 1/4
    }
    my_lattice:start()
end
```

in the current `lattice` the above will sometimes not run the first pattern `first_pattern` first, so the first number printed will be `1`. it usually prints something like this:

```
1
102
103
104
105
206
207
208
209
```

after this change to lattice, the patterns are evaluated in order so that it will always evaluate the `first_pattern`'s `+100` first:

```
101
102
103
104
205
206
207
208
209
210
211
```

destruction still works normally too.

### DIFF
--- a/lua/lib/lattice.lua
+++ b/lua/lib/lattice.lua
@@ -1,7 +1,7 @@
 --- module for creating a lattice of patterns based on a single fast "superclock"
 --
 -- @module Lattice
--- @release v1.2.0
+-- @release v1.2.1
 -- @author tyleretters & ezra & zack
 
 local Lattice, Pattern = {}, {}
@@ -23,6 +23,7 @@ function Lattice:new(args)
   l.superclock_id = nil
   l.pattern_id_counter = 100
   l.patterns = {}
+  l.pattern_ordering={}
   return l
 end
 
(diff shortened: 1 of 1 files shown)


schollz:
forgot to tag for review! whenever you have time @tyleretters could please you take a look?

tyleretters:
love it. thank you. really smart addition

tyleretters:
@tehn ok to merge

tehn (approved):

PR #1497, opened 2021-12-31
Make crow clock out always rise on the given division.

sixolet:
Previously, it would either rise on the division and fall halfway
between the divisions or vice versa, and which one would be based on
the vagueries of timing.

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -273,9 +273,11 @@ function clock.add_params()
     while true do
       clock.sync(1/params:get("clock_crow_out_div"))
       local crow_out = params:get("clock_crow_out")-1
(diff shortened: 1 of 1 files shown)


tehn:
thanks for spotting this! have you tested this at various tempos and divisions? i'm curious how well the inter-mixing of sync and sleep like this. (the race condition would be lessened by doing something like a 49% duty cycle instead of 50%) --- but also i'm theorizing so if it works, great!

sixolet:
My testing has so far been incidental to working on a new script, and has
consisted of turning the division up and down a few times to see if the bug
was gone at 90bpm.

I can set up a patch soon to test at faster bpms and finer divisions.

tehn (approved):

PR #1498, opened 2022-01-03
fix #1490

### LINKED ISSUE #1490: norns UI allows one MIDI device to occupy multiple device slots
Hi, I'm formulating this as an issue, but maybe I fail to see a productive or fun use for it. Let me know if llllllllines would be better place for this.

Norns allows setting the same MIDI device to occupy more than device slots, ie. one-to-many mapping between physical devices and the device listing.

This gets confusing, because some of the MIDI events get lost, e.g. just now I had a synth both at slot 1 and slot 5 (=also out of sight on 4-line listing in SYSTEM > DEVICES > MIDI). I got pretty confused; some scripts would respond to MIDI events, e.g. *Lost Future* would play sounds but others e.g. *MIDI Monitor* would not get MIDI data from the device. I now think the former listened to notes from all devices on all channels. Also `m = midi.connect(1); m.event = function(data) print("bah") end` callback would not print anything when pressing keys.

Everything works as expected when removing the duplicate device allocation, in this case I set device 5 to none.

I suggest a usability improvement UI redesign so that the MIDI device list only allows each device to occupy one slot. Maybe simply when assigning a device to a slot, remove it from other slots?

Cheers thanks for all the good stuff, take care, stay safe :)

dndrks:
i added a local function to `lua/core/menu/devices.lua` to check if a newly-assigned vport shares its device name with any other previously-assigned vports of the same device type. if there's a match, then the other vports are unassigned by changing their name to 'none' ahead of running that device type's `update_device` function. this solves the midi device issue described in #1490 , but it also avoids this behavior across all other device types (eg. one grid being assigned to multiple ports)

### DIFF
--- a/lua/core/menu/devices.lua
+++ b/lua/core/menu/devices.lua
@@ -45,6 +45,14 @@ local function set_len_for_list()
   end
 end
 
+local function check_and_rename(dev_type,pos,name)
+  for i = 1,#dev_type.vports do
+    if i ~= pos and dev_type.vports[i].name == name then
+      dev_type.vports[i].name = "none"
+    end
+  end
+end
+
 m.key = function(n,z)
   if m.mode == "type" then
     if n==2 and z==1 then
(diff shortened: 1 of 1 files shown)


tehn:
fixes #1490

tehn (approved):

PR #1504, opened 2022-01-17
add utility script: capture multiple CPU load readings for a process

catfact:
this quick-and-dirty python utility just batches many CPU readings for a given process using `top` and prints the collected results. (for subsequent plotting or analysis.)

the motivation is to facilitate more statistically useful analyses of CPU impact due to code changes, primarily on audio side.

usage:

`python ./test_cpu_load <procname> <n> <d>`

where
`procname` is the name of a running process (default `crone`)
`n` is a count of samples (default `10`.) NB: should not be less than 2
`d` is a delay between samples (default `0.2`.) NB: have noticed that smaller values yield sporadic zero readings.


tehn (approved):

catfact:
Hold off on merge 1 moment, would like to add a mode for capturing `jack_cpu_load` also

tehn:
indeed, was planning to let you merge at your leisure, no rush

catfact:
totally nuked the python version (slow, and `top` output is kind of useless)
it is now a trivial c program using jack's load average estimate and xrun handler.
this very useful for comparing performance between commits, doesn't affect anything else, i'd like to merge it to access from other branches (should we have a `tools/`?)

this stuff can also be easily added to matron, etc

tehn:
tools/ is a good idea

catfact:
will address that separately

PR #1505, opened 2022-01-17
fix softcut param types and _iif base

catfact:
two small changes:

- erroneous cast to `float` and back in softcut `_ii` parameter setter. looks like a copy/paste or codegen error. i am pretty sure this was benign.

- in weaver `_set_cut_param_iif`, indices **were not being converted from 1-base to 0-base**. this would cause a pretty serious bug. fortunately it was as un-serious as could be expected, since it only actually affects the `voice_sync` command which i believe is pretty rarely used. (i wonder if there are any scripts using it actually.)

(note that several other commands that are handled by `SoftcutClient` also accept `iif` parameters; but these are all "levels" and are processed through a different `weaver` function.)

### DIFF
--- a/matron/src/weaver.c
+++ b/matron/src/weaver.c
@@ -2518,17 +2518,17 @@ int _set_cut_param(lua_State *l) {
 int _set_cut_param_ii(lua_State *l) {
     lua_check_num_args(3);
     const char *s = luaL_checkstring(l, 1);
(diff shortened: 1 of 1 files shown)


tehn (approved):
good find--- fwiw this tripped me up in my review a few days ago to spot an off-by-one error (since the var name `voice` made sense for index conversion, but `val` didn't).

PR #1506, opened 2022-01-17
Revert "stop after next loop+fade "

tehn:
Reverts monome/norns#1494

@schollz i apologize--- i misunderstood some of the technical issues with this PR (i misinterpreted the problem as pre-existing), and without a clean fix i'd prefer not to add technical debt to the project. i understand your soft-fix of pushing the play command later generally "works" but we'd need a robust fix prior to adding this feature to main.

furthermore there are some needed optimizations, as i learn more about the careful architecting done by @catfact in the dsp loop.

i'm reverting this PR as an overdue norns update is waiting, and i can't foresee a realistic timeline for fixing this issue.

### DIFF
--- a/crone/osc-methods.txt
+++ b/crone/osc-methods.txt
@@ -44,7 +44,6 @@ osc methods:
  /set/param/cut/rec_level [if]
  /set/param/cut/pre_level [if]
  /set/param/cut/rec_flag [if]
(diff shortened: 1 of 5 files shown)


schollz:
no need to apologize @tehn! I understand. too bad I couldn't find the clean fix!

PR #1507, opened 2022-01-26
update readme-setup with packages, component descriptions

catfact:
- some required packages were missing (`jack-dev`, `libboost-dev`)

(yes yes, we should remove boost dependency.)

- general description of system components was very old (particularly bits about supercollider)

- included some pointers on how to run individual components during development

### DIFF
--- a/readme-setup.md
+++ b/readme-setup.md
@@ -2,21 +2,17 @@
 
 ## build prerequisites
 
-### packages
-```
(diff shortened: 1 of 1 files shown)


catfact:
oh; should actually update this with some instructions for desktop. (needs `libsdl2-dev`, needs `waf` options, needs `~/.matronrc`)

tehn (approved):
thank you so much, this is so needed

catfact:
added some stuff on desktop builds. probably not perfect (this is no longer exactly just a "setup" readme... `readme-development`?) but def better than nothing.

PR #1509, opened 2022-01-29
k1 + k3 for random script

stvnrlly:
This takes advantage of k1 as the alt key to hop to a random script with k3. That should be it!


ngwese:
I’m concerned adding behavior like this would be interpreted by most people as a bug in the selection code rather than an intentional feature.

Providing this as a mod seems more appropriate than putting it into the core script selection code.

tehn:
i agree. this would be a straightforward mod.

PR #1510, opened 2022-02-01
Add function/field docs for norns table

### LINKED ISSUE #1466: docs: missing functions
[e.g.] the entire `norns` table is not documented and there are a few very user-relevant functions in there, such as the system call with callback `norns.system_cmd = function(cmd, callback)`

https://github.com/monome/norns/blob/main/lua/core/norns.lua#L206

it was a shortcut to put this function in the `norns` table, but i'm not sure if there are enough other functions meriting a `system` library or similar.

needless to say, this should be documented so people can find it.

midouest:
Screen Shot 2022-02-01 at 8 14 56 AM

Fixes #1466

The primary documentation changes are in the first commit: 56b8aef900d8bcc7c018fc0b4fe95162345c8201

The second commit is the result of running `ldoc .` in the root of the repo. It seems like some unrelated changes may have been picked up by this. Let me know if this isn't the right process for updating the docs!

I'm happy to tweak the comments for each function if folks have suggestions. Thanks! 😄


midouest:
It just occurred to me that I marked this as fixing the related issue, but perhaps it's useful to leave that open to capture other Lua modules that are missing documentation?

catfact (approved):
thanks for cleaning these up

PR #1512, opened 2022-02-03
make param ID collision a scary warning instead

catfact:
reverting this error to a warning. 

i believe it should ultimately be an error, but that's not ideal for the moment without an extended beta period. i'm not quite convinced its doing what we want as i'm seeing some seemingly spurious/inconsistent examples that should be examined more thoroughly.

one obvious difficulty is that script params may collide with mod params! oy.

in any case, this change should make things better than before the error, because it adds an early return to the function instead of clobbering an existing param in the lookup table

...actually now that i think about it i'm not sure this is "better." scripts will still be broken because they have non-functional parameters. but they will not "steal" parameters from the system, breaking stuff even after the script is cleaned up.

should i add a more explicit set of tests for clobbering system parameters only? hm... @tehn et al, would appreciate opinions.

### DIFF
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -119,7 +119,11 @@ function ParamSet:add(args)
   end
 
   if self.lookup[param.id] ~= nil then
-    error("paramset.add() error: id '"..param.id.."' is already used by another parameter")
+    print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
+    print("!!!!! ERROR: parameter ID collision: ".. param.id .. ' - parameter not added!')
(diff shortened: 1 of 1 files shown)


catfact:
pushed update. this represents a compromise:

- system parameters are never clobbered. instead the parameter is simply not added. so scripts which try to do this, those parameters will stop working. (but won't throw an error. now i'm not 
even sure this is better... but in any case i honestly don't think this situation applies often, if ever.)

- script or mod parameters *are* clobbered. these cases are hard for us to predict, and really should be dealt with, but shouldn't ever cause an unrecoverable broken state as with system params. so by allowing this, we don't change any present functionality (or lack thereof) for any script/mod combo.

tehn (approved):
this compromise is fine, thank you for putting in the extra attention.

PR #1513, opened 2022-02-08
Add step for confirming PSET deletion

nattog:


lua/core/menu/params.lua
```diff
@@ -599,6 +607,10 @@ m.redraw = function()
         screen.text(line)
       end
     end
+  elseif m.mode == mPSETDELETE then
+    screen.move(63,40)
+    screen.level(15)
+    screen.text_center("CONFIRM DELETE")
```
tehn, line 613: i think adding a question mark clarifies that this is an action request. how about

DELETE PSET?

tehn (changes requested):
works great! thank you.

lua/core/menu/params.lua
nattog: Sure that makes sense. Are you happy with the text being uppercase? It felt this was most consistent with the existing pages in the params menu, but I had noticed there's a few inconsistencies across other menus - update text is lowercase, for example.

tehn (approved):

PR #1514, opened 2022-02-09
get tape index from file system, not index.txt

stvnrlly:
This looks directly to the contents of `tape/` to figure out what the index is, which _should_ be more reliable than looking at `index.txt`. I was having an issue where appending to the filename of stuff I'd recorded (e.g. `0012 - mlr - c minor`) prevented `index.txt` from incrementing.

It didn't seem possible to make this a mod, but if I'm missing something I'm happy to go that route instead.

### DIFF
--- a/lua/core/menu/tape.lua
+++ b/lua/core/menu/tape.lua
@@ -50,6 +50,7 @@ local function tape_exists(index)
 end
 
 local function read_tape_index()
+  os.execute("mkdir -p ".._path.audio.."tape/")
   tape = util.os_capture("ls ".._path.audio.."tape/", true)
   local t = {}
   for f in tape:gmatch("([^\n]+)") do
--- a/lua/core/menu/tape.lua
+++ b/lua/core/menu/tape.lua
@@ -45,13 +45,13 @@ local function tape_exists(index)
   if type(index) == "number" then
     index = string.format("%04d",index)
   end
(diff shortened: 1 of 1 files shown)


lua/core/menu/tape.lua
```diff
@@ -50,48 +50,30 @@ local function tape_exists(index)
...
-    m.fileindex = m.fileindex+1
-  end
-end
-
-local function write_tape_index_v(v)
-  local f = io.open(_path.tape..'index.txt','w')
-  if f == nil then
-    os.execute("mkdir -p ".._path.tape)
-    f = io.open(_path.tape..'index.txt','w')
+  tape = util.os_capture("ls ".._path.audio.."tape/", true)
```
tehn, line 53: could you add a check here that if the tape folder isn't found, the folder is automatically created?

otherwise looks fantastic! much more robust. thank you.

tehn (changes requested):
this is great--- way more robust. thank you!

lua/core/menu/tape.lua
stvnrlly: It's safe to run `mkdir -p` even if the folder exists, so I popped it in that way. I can wrap it in a conditional if that's preferred, though.

tehn (approved):

PR #1520, opened 2022-02-27
add engines, lua/engine folder

tehn:
deprecating `we` repo

including these few basic engines as standard

### DIFF
--- a/lua/extn/sky/unstable/engine/polysub.lua
+++ b/lua/extn/sky/unstable/engine/polysub.lua
@@ -5,7 +5,7 @@
 -- mutate global used by matron to select engine
 engine.name = "PolySub"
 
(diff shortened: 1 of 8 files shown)


lua/extn/sky/unstable/engine/polysub.lua
```diff
@@ -5,7 +5,7 @@
 -- mutate global used by matron to select engine
 engine.name = "PolySub"
 
-local glue = include('we/lib/polysub')
+local glue = include('polysub')
```
ngwese, line 8: does `include` work here? i thought include would only work for files coming from `dust`. iirc everything coming out of the core needs `require` (except for sky stuff which has its own `use` system)

tehn:
ah yes this likely needs to be 'require' will test and fix

PR #1521, opened 2022-03-05
basic image buffer drawing

ngwese:
the existing `screen_display_png(..)` function loads an image from disk, displays, and destroys the image each time which makes it comparatively expensive to use and not well suited to doing sprite animation.

this PR provides a more flexible way of leveraging images within a script ui. png images can be loaded into image buffers (a cairo surface under the hood) and managed as lua values which allows them to be reused. the `screen.display_image(...)` function has the same drawing behavior as `screen.display_png(...)` 

```lua
image = screen.load_png(paths.this.path .. "some.png")
screen.display_image(image, x, y)
```

displaying sprite style animation can be done with `screen.display_image_region(...)` to draw a subset of the image at a given location on screen.

```lua
digits = screen.load_png(paths.this.path .. "horizontal_row_of_digits.png")
width, digit_height = digits:extents()
digit_width = width / 10
which = 3
screen.display_image_region(digits, digit_width * which, 0, digit_width, digit_height, x, y)
```

the image buffer is a lua userdata type so buffers will be garbage collected when the go out of scope.


tehn (approved):
such a huge improvement, and the new lua class utilities are very welcome as well.

reviewed the code, and can test later today on device, but i can't immediately think of any concerns.

ngwese:
this script i was using to test the functionality can be found here https://github.com/ngwese/image_demo

the next step is to provide a way for screen operations to draw into an image so that ui elements can be drawn offscreen once and reused (should save cpu for complex uis)

ngwese:
i caught one minor error in one of the function prototypes and added to additional commits to do some minor clean up in `screen.c`

catfact:
thanks @ngwese looks real thorough and yeah the lua object interface is nice. (and thanks for the cleanup)

have you confirmed that the lifecycle works as expected,`_image_free` called by lua GC?

i'm alsao curious about offscreen drawing i suppose swapping out the current cairo surface i thought we actually had a version of that at one point, maybe it was not merged

catfact (approved):

ngwese:
i will confirm the gc is working before merging. the object setup is a refinement of the code which is in the soundplane integration and the lviv bindings i did.

ngwese:
looking good. the images are getting released on script clear, etc.

PR #1522, opened 2022-03-06
screen: draw to image support

ngwese:
adds the ability to direct screen drawing to an image

the api is straightforward:
```lua
image = screen.create_image(32, 32)
screen.draw_to(image, function()
  local w, h = image:extents()
  screen.clear()
  screen.level(15)
  screen.aa(0)
  screen.rect(0, 0, w, h)
  screen.stroke()
  screen.update()
end)
```

under the hood a screen "context" (aka cairo_t) is allocated and attached to the image object when drawing to the image is first attempted. this is done lazily for simplicity of user experience and to avoid allocating additional data for images which are never draw to. during the execution of the `screen.draw_to(...)` function the cairo context associated with the screen swapped out for the context associated with the image, the contexts are independent so any changes made to settings like `aa` or `level` and/or transformation only affect the image drawing (and are incidentally carried across `screen.draw_to(...)` calls with the same image).  the `screen.draw_to(...)` function uses `pcall` internally to ensure that errors raised while drawing to the image do not prevent restoration of the screen context.

various tests (https://github.com/ngwese/image_demo/tree/main/test) were used to validate correctness, specifically
- performing drawing outside of any function (toplevel)
- nesting calls to `screen.draw_to(...)`
- raising errors within `screen.draw_to(...)`
- pushing the system to memory exhaustion

this change also includes some small polish on the image handling support such as raising lua errors when failures occur.

### DIFF
--- a/matron/src/hardware/screen.c
+++ b/matron/src/hardware/screen.c
@@ -9,6 +9,7 @@
 #include <cairo-ft.h>
 #include <fcntl.h>
 #include <linux/fb.h>
+#include <math.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -74,6 +75,7 @@ static cairo_operator_t ops[NUM_OPS] = {
 static cairo_surface_t *surface;
 static cairo_surface_t *image;
 static cairo_t *cr;
+static cairo_t *cr_primary;
 
 static cairo_font_face_t *ct[NUM_FONTS];
 static FT_Library value;
@@ -83,7 +85,7 @@ static double text_xy[2];
 
 void screen_init(void) {
     surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 128, 64);
-    cr = cairo_create(surface);
+    cr = cr_primary = cairo_create(surface);
 
     status = FT_Init_FreeType(&value);
     if (status != 0) {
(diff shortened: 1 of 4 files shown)


matron/src/hardware/screen.c
```diff
@@ -529,5 +545,56 @@ void screen_surface_display_region(screen_surface_t *s,
...
+screen_context_t *screen_context_new(screen_surface_t *target) {
+    cairo_surface_t *image = (cairo_surface_t *)target;
+    // NOTE: cairo_create increases the ref count on image which avoids a double
+    // free and allows the image and context to be freed in any order
+    cairo_t *context = cairo_create(image);
+    cairo_status_t status = cairo_status(context);
+    if (status == CAIRO_STATUS_SUCCESS) {
+        return (screen_context_t *)context;
+    }
+    fprintf(stderr, "context_new: %s (%d)\n", cairo_status_to_string(status), status);
```
tehn, line 557: do you want to leave this in? (debug print?)

tehn (commented):
excellent

matron/src/hardware/screen.c
ngwese: yes, this one and that other one in `screen_surface_new` new where intentionally left in; it will help diagnose the source of problems with the higher up operations in weaver throw a more generic lua error.

tehn (approved):

PR #1524, opened 2022-03-10
V3 wip/replace osc

catfact:
ok, everything seems to be basically working...?

(though i am noticing a lot of xruns when swapping engines/scripts and haven't A/Bd this behavior with main.)

what has been done here:

- sidecar process added with socket IPC; the entire purpose of this is to allow child processes/shells to be launched (by the sidecar proc) without causing the main process to hang and stall audio when it is briefly halted and copied. (which is apparently what happens otherwise when a JACK client forks.) `_norns.execute` and `_norns.system_cmd` have been updated to use this mechanism. (the former is blocking, the latter is async, with same callback and lua-side response queue.)

- OSC is bypassed when communicating between `matron` and `crone`. since these are no longer separate processes, but just two jack clients in the same process, they can use direct function calls, now provided by new `crone.h`/`.cpp` API on one hand, and some new callbacks in `oracle.h`/`.cc` on the other.

-----

TODO / known issues:
- function signatures in `crone.h`/`.cpp` were autogenerated from OSC format strings, and i havent gone and renamed all the arguments to meaningful names in the header.
-  needs A/Bing for dropout/glitches and maybe bugs.
- i'm concerned about concurrent access to the sidecar client state from `stat.cc` worker threads and so on. but so far so good?

on a more fun note, one thing we can now add very easily is direct access from lua to softcut buffer contents (and maybe whatever other state seems useful.)


catfact:
oh right - i'll rebase this and fix the conflicts before merge.

PR #1536, opened 2022-03-21
More opt flags

catfact:
- changed `crone` opt level to `O3` and added a couple more flags on top for release builds. this squeezes another 2-3% of audio load headroom according to `baseline`. (not dramatic since we already had auto-vec stuff enabled at O2.)

- also propagated the CPU and FPU selections to `matron` for release builds. so that is probably a little faster too but we dont have as convenient a profiling mechanism.

the minor downside is that a few seconds have been added to compiation time.

### DIFF
--- a/crone/wscript
+++ b/crone/wscript
@@ -25,7 +25,7 @@ def build(bld):
     ]
 
     crone_flags = [ '-std=c++14',
-                    '-O2',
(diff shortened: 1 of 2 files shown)


tehn (approved):
added a fix, `matron_cflags` was unused

PR #1538, opened 2022-03-22
Broken windows

catfact:
- `jack_capture` is redundant, removed
- i've been yelling about resource leaks in supercollider, so let's not have any in the core repo!


tehn (approved):
are pargroups necessary for vanilla sc? i have some weird recollection that we added them for supernova (multicore support)

catfact:
with scsynth, a ParGroup is just a Group on the server. but the sclang code still signals the intention that you don't care about the processing order of the group's children. so may as well leave it i guess

PR #1541, opened 2022-03-28
Print incoming osc IP and port properly

zjb-s:


ngwese (commented):
this change places much greater pressure on the lua garbage collector because it forces the construction of a formatted string. 

the previous implementation did not have that behavior.

zjb-s:
That makes sense, I'm not too familiar with the garbage collection. It's the concatenation that makes it slow? If so, I'm happy to rewrite this into multiple simpler prints

ngwese:
I’m assuming that the primary desire is to see the sending address and port number, if so i’d suggest simply:
```lua
print("incoming osc message from:”, from[1], from[2], path)
```

catfact:
sorry to barge in, but i have to wonder if printing stuff from the default handler is really what we want to be doing. (this will fire in response to _any_ incoming OSC when the script hasn't set up its own handler.)

for development, printing everything makes sense; for normal use its a potentially very large amount of I/O being performed in the background..

zjb-s:
I don't know enough about the stack to say whether it's worth it to print off the default handler. But if we're going to, I'd love to have the print be more useful :)

greg's rewrite is fine with me.

ngwese:
on the whole i am of a similar mind and agree that the default behavior should have probably  been to do nothing.

since this is existing code which as implemented prints a table address, changing it to print more useful values in a manner which is no worse performance wise seems reasonable. 

if we want to remove this print statement (which again i would support) i think it would more sense to scrub the code more broadly for cases of excessive logging.

tehn:
i agree the default should be to do nothing-- this is a remnant from debugging/development and should've been removed (apologies for setting up a bad pattern by having this persist)

tehn (approved):

ngwese:
given the desire to remove the print: @zjb-s do you want to modify this PR to just remove the line?  …it would morph into a PR from you to harden the norns stack and improve a potential performance problem when a script is not using OSC 💪

zjb-s:
heh. already done, and merged! unless I’m missing something?

PR #1542, opened 2022-03-28
combined, redone softcut fixes (update submodule)

catfact:
through the combined action of blind commits, github outages, and cosmic rays, the softcut submodule has become a mess of broken conflicted branches since last norns release.

so: `main` and `norns-latest` have been forcibly reset; they are now the same thing and this PR points us at `norns-latest`.

following changes have been rebuilt:

### crucial fixes

- `Voice` now zeroes its output bus in the no-op sample update routine. this avoids possible very nasty init glitches

### minor fixes and updates

- `Svf` now resets its internal states along with other `Voice` components. might avoid some less bad init glitches

- `Svf` is now pure c++ and doesn't manipulate an internal struct for filter state

- filter cutoff freq is clamped more reliably (in the DSP setter, not just the IPC/API) and to values derived from actual samplerate

- minor opt: frequency warping function isnt called with resonance change

- minor opt: a couple instructions have been saved on frequency update.

this has been built on CM3 factory norns and (finally) tested with several softcut-based scripts on the device. (if briefly.)


tehn (approved):
working beautifully

PR #1543, opened 2022-03-31
matron: embed lua-cjson module

ngwese:
this is a quick pass at building and statically linking the `cjson` module directly into matron. i've minimally tested this by calling `json.encode(...)` and `json.decode(...)` in the repl.

details to consider:
- this registers the module as `json` in the global environment
- embedded modules can't be accessed via `require(...)` since there is nothing to search for, this is maybe confusing for people
- ~~if a scripts assigns a value to `json` it effective overwrites the module registration making it inaccessible until matron is restarted~~

my recollection was that we did something to guard against global modules getting overwritten but in looking around i couldn't find anything that appeared to protect (or restore) the say `screen` or `grid` after a script ran. i did notice that we have this https://github.com/monome/norns/blob/main/lua/core/script.lua#L33-L40 - which if i'm understanding it correctly will prevent new globals from leaking into the environment but it doesn't seem to prevent overwriting of globals.

EDIT:  i've pushed an additional commit which instead registers the modules as `_json` global and then assigns that to `json` at startup and script cleanup. this doesn't prevent the table itself from being altered but it does help to recover from accidental assignment to `json` 

EDIT2: if we wanted to retain the use of `require`ing the module i could also simply add a trivial `json.lua` file which returns the `_json` global or a copy of the table.

/cc @schollz

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "crone/lib/readerwriterqueue"]
 	path = crone/lib/readerwriterqueue
 	url = https://github.com/cameron314/readerwriterqueue.git
+[submodule "third-party/lua-cjson"]
+	path = third-party/lua-cjson
+	url = https://github.com/mpx/lua-cjson.git
--- a/lua/core/script.lua
+++ b/lua/core/script.lua
@@ -27,6 +27,9 @@ Script.clear = function()
     package.loaded['asl'] = nil
   end
 
+  -- reset embedded modules
+  json = _json
+
   -- script local state
   local state = { }
 
--- a/lua/core/startup.lua
+++ b/lua/core/startup.lua
@@ -29,6 +29,9 @@ paramset = require 'core/paramset'
 params = paramset.new()
 norns.pmap = require 'core/pmap'
 
+-- embedded global modules
+json = _json
+
 -- load menu
 require 'core/menu'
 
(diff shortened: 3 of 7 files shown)


catfact:
looks good to me. no trouble building+running but i didnt do any functional testing (sorry!)

re: protecting globals,
i also havew a memory of the top-level globals being somehow write-protected, although their fields were not. maybe this was reverted for some reason, or never merged to main.

catfact (approved):

tehn:
`tab.readonly` protects elements from write, but not the table itself.

the script reset should do a full reset of the global environment, so i'm not sure any extra json resetting is required? i'm fine having it be a perpetual global if others are ok with that

tehn (approved):
looks good, and nice to see the implementation

ngwese:
i added the reset logic in part because i wrote two trivial scripts. one printed the module table and then assigned a different table to the variable. the second just printed the table. running them in succession showed that the `json` module was not getting restored.

since one of the contexts in which json is useful is saving state (in mods and scripts if not other places) that seemed like a bad thing to break until restart.

tehn:
hmm, that suggests the global reset on script load isn't doing what we
want/think

will investigate

ngwese:
in the interest of building momentum towards the next release i've opted to put this change into the `norns-converged` branch

PR #1544, opened 2022-03-31
add PSET number to PSET actions + add delete action

dndrks:
while writing up examples for saving table data with PSETs (building off of @schollz 's incredible work on this front), i realized that PSET index number is a helpful reporting callback -- since PSETs made via the PSET UI can share names, the index number is often the only differentiator. rather than force authors to extract this from the loaded filename, it made sense to add the PSET's index number (if available) to the actions.

i also realized that a PSET deletion action would be helpful for external data management, so i added that as well.

here's a quick example of usage:
```lua
MusicUtil = require 'musicutil' -- see https://monome.org/docs/norns/reference/lib/musicutil
s = require 'sequins' -- see https://monome.org/docs/norns/reference/lib/sequins

function init()
  base_note = math.random(30,50)
  my_seq = s{}
  notes_array = MusicUtil.generate_scale_of_length(base_note, "dorian", 16)
  note_data = {}
  for j = 1,64 do
    -- auto-generate 64 steps of notes:
    note_data[j] = MusicUtil.snap_note_to_array(math.random(base_note, base_note+28),notes_array)
  end
  my_seq:settable(note_data) -- send this table of notes to the sequins
  play = false
  
  -- here, we set our PSET callbacks:
  params.action_write = function(filename,name,number)
    print("finished writing '"..filename.."' as '"..name.."'", number)
    os.execute("mkdir -p "..norns.state.data.."/"..number.."/")
    tab.save(note_data,norns.state.data.."/"..number.."/note.data")
  end
  params.action_read = function(filename,silent,number)
    print("finished reading '"..filename.."'", number)
    note_data = tab.load(norns.state.data.."/"..number.."/note.data")
    my_seq:settable(note_data) -- send this restored table to the sequins
  end
  params.action_delete = function(filename,name,number)
    print("finished deleting '"..filename, number)
    norns.system_cmd("rm -r "..norns.state.data.."/"..number.."/")
  end
end
```

in the example:
- when a user saves a PSET, its index is passed to create a numeric directory in the script's data folder -- the note table is saved there as `note.data`
- when a user reads a PSET, its index is passed to load the note data into a sequins
- when a user deletes a PSET, its index is passed to delete the directory which holds the additional tables, etc

this is testing well on my side, but please lmk if there's anything i'm missing!

### DIFF
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -449,11 +449,7 @@ function ParamSet:write(filename, name)
     end
     io.close(fd)
     if self.action_write ~= nil then
-      if pset_number ~= nil then
-        self.action_write(filename,name,pset_number)
-      else
-        self.action_write(filename,name)
-      end
+      self.action_write(filename,name,pset_number)
     end
   else print("pset: BAD FILENAME") end
 end
@@ -499,11 +495,7 @@ function ParamSet:read(filename, silent)
       end
     end
     if self.action_read ~= nil then 
-      if pset_number ~= nil then
-        self.action_read(filename,silent,pset_number)
-      else
-        self.action_read(filename,silent)
-      end
+      self.action_read(filename,silent,pset_number)
     end
   else
     print("pset :: "..filename.." not read.")
@@ -522,11 +514,7 @@ function ParamSet:delete(filename, name, pset_number)
   print("pset >> delete: "..filename, name, pset_number)
   norns.system_cmd("rm "..filename)
   if self.action_delete ~= nil then
-    if pset_number ~= nil then
-      self.action_delete(filename, name, pset_number)
-    else
-      self.action_delete(filename, name)
-    end
+    self.action_delete(filename, name, pset_number)
   end
 end
 
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -499,10 +499,10 @@ function ParamSet:read(filename, silent)
       end
     end
     if self.action_read ~= nil then 
-      if pset_number == nil then
-        self.action_read(filename,silent)
-      else
+      if pset_number ~= nil then
         self.action_read(filename,silent,pset_number)
+      else
+        self.action_read(filename,silent)
       end
     end
   else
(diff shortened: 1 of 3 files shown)


lua/core/paramset.lua
```diff
@@ -445,8 +448,12 @@ function ParamSet:write(filename, name)
       end
     end
     io.close(fd)
-    if self.action_write ~= nil then 
-      self.action_write(filename,name)
+    if self.action_write ~= nil then
+      if pset_number ~= nil then
```
tehn, line 452: i don't think you need to check for nil: just sending nil is fine

(just simplifies the code, not functional)

lua/core/paramset.lua
```diff
@@ -490,13 +499,37 @@ function ParamSet:read(filename, silent)
       end
     end
     if self.action_read ~= nil then 
-      self.action_read(filename,silent)
+      if pset_number ~= nil then
```
tehn, line 502: same here

lua/core/paramset.lua
```diff
@@ -490,13 +499,37 @@ function ParamSet:read(filename, silent)
...
+function ParamSet:delete(filename, name, pset_number)
+  if type(filename) == "number" then
+    local n = filename
+    filename = norns.state.data .. norns.state.shortname
+    filename = filename .. "-" .. string.format("%02d",n) .. ".pset"
+  end
+  print("pset >> delete: "..filename, name, pset_number)
+  norns.system_cmd("rm "..filename)
+  if self.action_delete ~= nil then
+    if pset_number ~= nil then
```
tehn, line 525: and here

tehn (changes requested):
cool! super into this. just a little cleanup (and quick test) then this is good to go.

needs doc update

dndrks:
rad, thank you @tehn! super helpful review, which led to a good cleanup and additional testing.
changes are now all documented, all testing well -- plz lemme know if there's any loose ends from your perspective!

tehn:
aha, the html files are actually autogenerated from the lua via luadoc, so no worries about those. i should've been clearer: i meant https://monome.org/docs/norns/reference/params in case required.

great looking edit, thank you!

tehn (approved):

PR #1547, opened 2022-04-02
fix: params:delete with no filename

dndrks:
quick fix for `params:delete` -- i realized that `params:read` has a failsafe to use the last-loaded pset if no filename is provided, so i added this to `:delete` as well


tehn:
thanks for being thorough!

i feel like calling delete without a param should just bail, as i can't think of a UI situation where this wouldn't be prone to mis-deletion.

but i'm up to discuss this if you have some specific examples?

dndrks:
makes sense! down to leave it bailin'

tehn:
you actually found a bug; this doesn't gracefully bail, it just does `os.execute('rm ')` which is an error.

best replace your added line with `if filename==nil then return`

PR #1559, opened 2022-04-07
replace nanomsg with nng

ngwese:
the `nanomsg` library is in maintenance mode and has been superseded by  `nng`. additionally `nng` supports secure websockets which will be required in order to (eventually) add authentication to maiden.

_in order to build with this change the `libnng-dev` package must be installed._ 

while `nng` does offer compatibility headers for `nanomsg` the sidecar, maiden-repl, and ws-wrapper code has been updated to use current `nng` api in order to maintain consistency once authentication is implemented.

some minor touchups to maiden-repl are included as well (build cleanup and one crash fix)

### DIFF
--- a/maiden-repl/src/io.c
+++ b/maiden-repl/src/io.c
@@ -6,8 +6,9 @@
 #include <string.h>
 #include <unistd.h>
 
-#include <nanomsg/nn.h>
-#include <nanomsg/bus.h>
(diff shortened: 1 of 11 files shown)


tehn (approved):
excellent, thank you!

i'll add nng dep note to the readme

PR #1561, opened 2022-04-07
Update pmap.lua w/ clearer report

zjb-s:
https://llllllll.co/t/tapedeck/51919/35?u=zbs

I've seen this confusion a couple of times and thought a slightly clearer message would be better.


catfact (approved):
ironically, i visisted github to open a PR with this very change, and found this one instead.

Issue #1557, opened 2022-04-06
Midi.remove doesn't return device info


jaseknighter:
Midi.remove is supposed to return a device object according to the docs, but it returns nil.

i believe the fix is in this file: /norns/blob/main/lua/core/midi.lua

line 403 should be changed from: `Midi.devices[id].remove()`

to something like:

```
local dev = {
  Midi.devices[id].id
  Midi.devices[id].name
  Midi.devices[id].dev
}
Midi.devices[id].remove(dev)
```

jaseknighter:
i'll see if i can fix this myself and submit a PR (i haven't tried to change norns code before).

ngwese:
i don’t have the code in front of me but i’d recommend validating which of those fields in the table being returned are safe/valid to even return after the device itself is gone. for instance i’m not sure the `dev` handle is valid or useful at that point.

jaseknighter:
good point @ngwese. perhaps it would be better to just return the device's name and id, like:

```
local dev = {
  id = Midi.devices[id].id
  name = Midi.devices[id].name
}
Midi.devices[id].remove(dev)
```

jaseknighter:
i've committed a fix to my local repo to address this issue and submitted a pull request. also, in the fix i updated the autodocumentation code to indicate the return object includes midi id, name, and port.

catfact:
@jaseknighter does this change do what you need?
https://github.com/monome/norns/pull/1590/files

jaseknighter:
i've not tested but the change looks fine to me. i didn't really have a specific need for this change other than to address the discrepancy between the docs and the code (as mentioned earlier).

catfact:
id like to close this issue. it;s confuzing to me. firstply i think it's supposed to say that thep per-device callback should be getting device info as a parameter - which is what the comment illustrates, and which i actually don't agree with.

looking at this area the only issue i saw was that the glohbal customizable device removal callback wasn't firing, as the (inline) documentaiton suggests it should - hence the PR

i don't know of any other discrepancies with docs but please point out if you see

PR #1562, opened 2022-04-07
fix to add midi data when device is removed (issue #1557) 

jaseknighter:
midi id, name, and port is passed back when a device is removed. this info is also printed (same as when device is added). also, the luadoc code has been updated so people can see in the docs what data is returned when a device is removed.

i tested by recompiling the docs locally and i verified the print data looked correct when removing a midi device. 

this is my first pull request for making an update to norns code. if there's anything i should do process wise to improve this or future submissions, please let me know.


dndrks:
omg apologies for the delay here!
thank you for closing the loop on this feature-set -- it's all testing well on my end!

apologies for the late bump, but i'm curious if there's any benefit to more strictly following the `_norns.midi.add` formatting for `_norns.midi.remove`, where we're calling the generic `Midi.remove` function and passing the device info, rather than specifying the device's aliased version? eg:

```lua
-- remove a device.
_norns.midi.remove = function(id)
  if Midi.devices[id] then
    if Midi.remove then
      local dev = {
        id = Midi.devices[id].id,
        name = Midi.devices[id].name,
        port = Midi.devices[id].port
      }
      print(string.format("_norns.midi.remove: %d, %s, %s",dev.id,dev.name,dev.port))
      Midi.remove(dev)
    end
  end
  Midi.devices[id] = nil
  Midi.update_devices()
end
```

everything should work the exact same, i think it just clarifies that there aren't `.remove` callbacks for each device, but rather for any :)

lmk if i've misunderstood / if you have any concerns!

ngwese:
Sorry. This slipped past my view. Will take a look tonight.

tehn (approved):

tehn:
ack, i sortof plowed ahead and merged after reading the code and understanding it from a particular angle, but @dndrks 's comment also makes sense. let me know if this should be reverted, apologies for the mess

jaseknighter:
@ngwese, do you have a recommendation here? i’m pretty sure i don’t understand the code well enough to decide whether @dndrks' suggestion is the way to go and @tehn's merge should be reverted.

catfact:
um, sorry for not weighing in on this i guess? i am generally taking a long break from norns.

i'm not sure i understand what is solved by the proposed change. the idea here is that after connecting to a device, a script author can set the midi device's `remove` callback to do something useful when the device is unplugged. i don't really see the point of adding the device table as a parameter to that callback, but i suppose it doesn't do any significant harm.

jaseknighter:
ezra, i hope you are well. it has been a while since i submitted this pr, but if i remember correctly, i created it because the api docs states that there is a static callback when any midi device is removed but that callback was not actually implemented so i tried to implement it.

catfact:
ah i see. you are referring to this in the lua source i think?

https://github.com/monome/norns/blob/main/lua/core/midi.lua#L75-L79

i think you're right! that callback is not executed. probably at some point we switched to the per-device removeal callback scheme and forgot that the global custom callback was stilll hangin around.

here are two possible fixes:

- add a line with `Midi.remove(device)` to the weaver-controlled `norns.midi.remove` function, so that `Midi.remove` works as expect3ed.

- delete `Midi.remove` from midi.lua and dont' worry about it anymore. (nobody is using it now because it  evidently does not work.)

catfact:
i would like to revert the merge of this PR since it does break things for me by actually removing the per-device calllback. then i will implement option (1) above to fix the static callback.

PR #1563, opened 2022-04-10
provide the ability to manipulate audio routing

ngwese:
- adds low level wrappers for various `jack` functions to introspect and manipulate ports
- adds high level api to alter audio routing by port, by tables of ports, or by component
- adds a hook for tweaking audio routing in mods
- ensures supercollider is connected on restart
- removes all system calls out to `jack_*` commands

### DIFF
--- a/lua/core/audio.lua
+++ b/lua/core/audio.lua
@@ -2,6 +2,7 @@
 -- @module audio
 
 local cs = require 'controlspec'
+local hook = require 'core/hook'
 
 local Audio = {}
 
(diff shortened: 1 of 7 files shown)


lua/core/audio.lua
```diff
@@ -208,6 +209,209 @@ Audio.level_cut_rev = function(value)
...
+
+--- default audio connections for system (hardware)
+Audio.default_system_routes = {
+  ["system:capture_1"] = "crone:input_1",
+  ["system:capture_2"] = "crone:input_2",
+  ["crone:output_1"]   = "system:playback_1",
+  ["crone:output_2"]   = "system:playback_2",
+}
+
+Audio.default_system_routes_reversed = {
```
tehn, line 240: rather than keep a separate table for reversed routes, would it be bad form to just modify the default routes table in-place? that way we skip all of the extra checking and comparisons later? (by this i mean, the routes table gets modified upon detection of the reverse.txt at startup)

but as-is has the advantage of being very clear and less error prone

tehn (approved):
code looks great, haven't tested on device but good to merge!

lua/core/audio.lua
ngwese: that is a totally fair observation. one vague thought i had around the tables was tweaking them in a mod such that an alternative "default" configuration was setup by the system w/o additional code but i couldn't decide on how to do that while also staying compatible with the reverse stereo case. the one concern i have about programmatically modifying them is that it implies the jack client names and routings for the "system" paths are fixed. 

fwiw an earlier version of this didn't expose the tables at all.

tehn: mmmm i sortof like the idea of a MOD just to flip the channels, given this is just a narrow batch of shields affected (and, curiously, many people don't seem to notice?) ---- just to keep the core codebase small

ngwese: a mod sounds like an interesting path, will revise this PR.

catfact (approved):
havent run it on device yet but all looks shipshape to me!

ngwese:
okay - i've pulled out the handling of the reverse case and added a more elegant means of disabling the change tracking. when combined with this mod: https://github.com/ngwese/reverse_io/blob/main/lib/mod.lua all the needed functionality should available.

PR #1568, opened 2022-04-22
update readme.md

tyleretters:
should we update this here @catfact ?


catfact:
Sure...  Maybe a note as to what that does (builds explicitlt for pi3/pi4 architectures with aggressive optimization)

tyleretters:
awesome! thank you!!

PR #1571, opened 2022-05-02
Add default "custom value" field in `tab.gather`

zjb-s:
This allows copying a table by calling `tab.gather(myTab)`. Previously you needed to call it as `tab.gather(myTab,{})`

### DIFF
--- a/lua/lib/tabutil.lua
+++ b/lua/lib/tabutil.lua
@@ -245,6 +245,7 @@ end
 -- @treturn table composite table
 function tab.gather(default_values, custom_values)
   local result = {}
+  if type(custom_values) == 'nil' then custom_values = {} end
   for k,v in pairs(default_values) do 
     result[k] = (custom_values[k] ~= nil) and custom_values[k] or v
   end


ngwese:
given that the stated purpose of the function is to combine two tables it seems odd to me to change it to make the second parameter optional. 

it seems more appropriate for that caller of the function to handle the possibility of a passing a nil value by doing `tab.gather(defaults, custom or {})`. so some usage of the gather function it may be more appropriate to fail if the second argument is nil because it represents an error in the script itself.

zjb-s:
i tend to use this function to return a plain copy of a table (not trivial in lua, unless i'm missing something big) rather than combine tables, but maybe that's more of an edge-case than I thought?

if that's the case, we can just close this. it's a minor point anyway. and your point about errors makes sense.

thank you greg!

ngwese:
if you haven’t seen it i might suggest reading through this page http://lua-users.org/wiki/CopyTable, in particular it speaks to the ambiguity around the notion of copying. 

this basic clone function at the top of the page would give you a more performant way to achieve the same behavior as calling `table.gather` with an empty table. 

```lua
function table.clone(org)
  return {table.unpack(org)}
end
```

zjb-s:
that's really helpful, thank you!!

PR #1572, opened 2022-05-08
use abl_link C API for link clock

artfwo:
ableton link has official C bindings now, this PR updates link clock to use them.


artfwo:
quick testing on hardware doesn't seem to expose any problems.

tehn (approved):

PR #1573, opened 2022-05-09
Allow chaining of actions

awwaiid:
This saves the existing action and when invoked calls the old one and the new one.

This came up as a discussion on the norns study group discord. It is not extensively tested. At all. But here for consideration!

### DIFF
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -352,15 +352,27 @@ function ParamSet:set_action(index, func)
   param.action = func
 end
 
---- set additional action
+--- run an action after the existing action
 -- @param index
--- @tparam function new_action set an additional action for this index
+-- @tparam function new_action the new action to run
 -- Chain a new action onto a paramset (even if one isn't already set)
-function ParamSet:chain_action(index, new_action)
+function ParamSet:append_action(index, new_action)
   local current_action = self:lookup_param(index).action or function() end
   self:set_action(index, function()
     current_action()
     new_action()
+  end)
+end
+
+--- run an action before the existing action
+-- @param index
+-- @tparam function new_action the new action to run
+-- Chain a new action onto a paramset (even if one isn't already set)
+function ParamSet:prepend_action(index, new_action)
+  local current_action = self:lookup_param(index).action or function() end
+  self:set_action(index, function()
+    new_action()
+    current_action()
   end)
 end
 
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -352,6 +352,18 @@ function ParamSet:set_action(index, func)
   param.action = func
 end
 
+--- set additional action
+-- @param index
+-- @tparam function new_action set an additional action for this index
+-- Chain a new action onto a paramset (even if one isn't already set)
+function ParamSet:chain_action(index, new_action)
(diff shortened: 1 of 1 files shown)


awwaiid:
hmm. maybe it should be named `add_action`.

ngwese (changes requested):
Took a quick look. One thing which needs to be accounted for is that action functions for some or all parameter types are passed the value of the parameter. Chaining functions in this manner would need to maintain that behavior.

It would be great to know a bit more about the use cases which motivate the feature.

ngwese:
@awwaiid is this still something you are interested pushing forward on?

awwaiid:
The idea here is that a library or a user-script, can chain on top of existing/shared param actions. So if a library and a script are aware of the same param, either one can extend the other. This can also be a multi-action code organization scheme, or a debugging scheme (chain in print statements before or after the existing action). The current version of this also handles the case where there isn't an existing action.

I don't see any unit tests for lua libs, there aren't any are there? I built a local test to verify the `...` syntax usage here for lua 5.1.

Here is the test script:

```lua
local callback = function(name, value)
  print("name", name)
  print("value", value)
end

function append_action(new_action)
  local current_action = callback
  callback = function(...)
    current_action(...)
    new_action(...)
  end
end

function prepend_action(new_action)
  local current_action = callback
  callback = function(...)
    new_action(...)
    current_action(...)
  end
end

callback("a", "b")

append_action(function()
  print("Do this after")
end)

print("----------------")
callback("c", "d")

prepend_action(function(n)
  print("Do this before. Name:", n)
end)

print("----------------")
callback("e", "f")
```

awwaiid:
I should note -- I do not personally have anything that requires this functionality; if y'all think it adds unneeded complexity no problem closing it until the use-case comes up more explicitly if ever. The conversation I invented it for is on the study group discord at https://discord.com/channels/765746584582750248/825387870873190410/973030210633097266

ngwese:
FWIW I'm still a bit conflicted as to whether this is _the_ abstraction to integrate directly into the param system for a few reasons

- The first is it provides only one of potentially multiple forms of function composition - sequencing. If we have two functions `f` and `g` and `x` is the parameter value this chaining abstraction provides `f(x), g(x)` or `g(x), f(x)` but equally useful would be `f(g(x))` or `g(f(x))`.

- The second is that it makes assumptions about whether an error occurring in once of the callbacks should be ignored or whether it should immediately fail the entire chain

- The third is that the proposed API; by providing append and prepend methods, implies that order is important but if scripts and mods have these calls sprinkled about then as script code is refactored or the system evolves the order in which the callback chain is built up is hard to determine.

 I guess if others find it useful I'm not overly opposed, it just feels like something which could be handled equally as well outside of the param system while also introducing people to the utility of functional composition.

catfact:
i agree with the concerns: this doesn't seem as powerful as a composition system needs to be; it doesn't seem to provide a way of "undoing" the monkeying of a param function; and i don't know why this syntax needs to be added to parameters in particular. (it seems like a very minor shortcut for the generic idiomatic syntax for monkeying a function in lua.)

awwaiid:
I'll go ahead and close this PR so it's not hanging out here. Let me know if y'all would like to pursue some variant of this!

Issue #1569, opened 2022-04-25
ensure sidecar response is null terminated


ngwese:
the `norns-converged` branch exhibits stability issues around script selection. sometimes the menu hangs, other times it shows garbage at the end. looking at the code my current guess is the buffers returned from the sidecar are not null terminated

https://github.com/monome/norns/blob/421ca6ea96742827a511bbeca3b08fa0b40da82e/norns/sidecar.cpp#L153

catfact:
perhaps enough to make it `buf = (char *)realloc(buf, nb+1);` (adding extra byte for null)

ngwese:
i've been poking at this and i believe that there is more than one thing going on. in the case that we read less than `CMD_CAPTURE_BYTES - 1` (which seems like the most common case) then the last byte of the the command output is replaced with `\0` which stops that byte.
https://github.com/monome/norns/blob/421ca6ea96742827a511bbeca3b08fa0b40da82e/norns/sidecar.cpp#L38-L39
since we never allow the full size of CMD_CAPTURE_BYTES to be read it is safe to `nb++` and then terminate. unfortunately that didn't fix the problem.

didn't deeper things stability improved when i disabled the background stat collection. looking at the code that makes sense. the `stat.cc` code is creates its own thread and then calls `sidecar_client_cmd` in a loop. invocation of norns.system_cmd in turn is also calling `sidecar_client_cmd` but from the lua thread so we have potentially overlapping requests to run commands. since every call to `sidecar_client_cmd` is doing a synchronous send/recv using the same nanomsg/nng socket i'm not sure that is anything which guarantees that the response coming back from the sidecar is getting matched with the correct requester thread.

looking at the nng docs on req sockets: https://nng.nanomsg.org/man/v1.3.2/nng_req.7.html

> Furthermore, only a single receive operation may be pending at a time. Attempts to post more receive operations concurrently will result in NNG_ESTATE.

> Requests may be canceled by sending a different request. This will cause the requester to discard any reply from the earlier request, but it will not stop a replier from processing a request it has already received or terminate a request that has already been placed on the wire.

this all leads me to believe the some of all of the following would be worth considering:
- replace the lua level async command/capture queue and one shot thread created in `system_cmd.cc` with a dedicated thread consuming consuming from another nng `inproc` socket and servicing requests to the sidecar serially. this would allow any thread to run system_cmd safely.
[...]

PR #1574, opened 2022-06-11
sidecar: serialize command invocation requests

ngwese:
- fix capture buffer termination (issue #1569)
- moves async command capture logic from lua to c
- ensures sidecar requests from `_norns.execute` and `_norns.system_cmd`
  are not interleaved when called from different threads
- adds progressive capture buffer allocation in server up to
  maximum of 10MB

additionally this change adds a sidecar client side "request queue" which is derived from the core event queue. while overkill for this particular fix the intension is to lay part of the infrastructure for moving the `stat.cc` logic out of the main norns process and into the sidecar which will reduce the overhead of gathering stats.

### DIFF
--- a/lua/core/norns.lua
+++ b/lua/core/norns.lua
@@ -200,33 +200,13 @@ norns.is_norns = norns.platform == 2
 --- true if we are running on norns shield (PI3)
 norns.is_shield = norns.platform == 3
 
--- Util (system_cmd)
-local system_cmd_q = {}
-local system_cmd_busy = false
-
---- add cmd to queue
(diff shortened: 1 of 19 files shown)


ngwese:
i've added a bonus change which adds names to many of the threads so running `info threads` in gdb is more informative.

tehn (approved):
fantastic! and great bonus feature with the thread labels, thank you

catfact (approved):
awesome. love the cleaned up and expanded sidecar API with sync/async.

PR #1575, opened 2022-06-15
add script-definable action for when clock tempo changes

dndrks:
during scripting, sometimes it's helpful to push changes to `PARAMETERS > CLOCK > tempo` across other parts of the script which work within the bounds of timing but perhaps don't use `clock.sync`. currently, this requires re-patching the system tempo change handler inside of the script, as described here: https://llllllll.co/t/norns-scripting/14120/1010

to avoid development pitfalls / weird states if a script doesn't reset the system handler, this PR adds `clock.tempo_change_handler`, which is tied into the system-defined `clock_tempo` parameter action, and passes the `bpm` and `source` into a user script, eg:

```lua
function clock.tempo_change_handler(bpm,source)
  print(bpm,source)
  engine.delay_time(clock.get_beat_sec() * bpm/128) -- imagining a clock-synced delay mechanism inside of an engine
end
```

the above will print `91  internal` (for example) and send the engine the correct timing value.

the `clock.tempo_change_handler` function is reset during clock cleanup, so there's a clean slate in between scripts.
please lmk if there's anything else to consider on this! tyty!


lua/core/clock.lua
```diff
@@ -88,6 +94,7 @@ clock.cleanup = function()
 
   clock.transport.start = nil
   clock.transport.stop = nil
+  function clock.tempo_change_handler(bpm,source) end
```
artfwo, line 97: for consistency with clock.transport callbacks cleared here this could be set to `nil`, given that the check below will only call tempo_change_handler if it's not nil anyway.

```suggestion
  clock.tempo_change_handler = nil
```

lua/core/clock.lua
```diff
@@ -51,6 +51,12 @@ clock.sync = function(...)
   return coroutine.yield(SCHEDULE_SYNC, ...)
 end
 
+--- static callback when clock tempo is adjusted via PARAMETERS > CLOCK > tempo;
+-- user scripts can redefine
+-- @static
+-- @param bpm : the new tempo
+-- @param source: the clock source
+function clock.tempo_change_handler(bpm,source) end
```
artfwo, line 59: maybe this definition could be moved together with the other callbacks (clock.transport) and declared as `clock.tempo_change_handler = function...` for consistency with the rest of the module?

also see the comment about resetting this function to `nil` below, it could be initialized as `nil` value too.

dndrks:
rad, thank you @artfwo !! folded the changes in -- much clearer, appreciate your eyes + insights!!

lua/core/clock.lua
```diff
@@ -210,6 +218,9 @@ function clock.add_params()
       if source == "internal" then clock.internal.set_tempo(bpm)
       elseif source == "link" then clock.link.set_tempo(bpm) end
       norns.state.clock.tempo = bpm
+      if clock.tempo_change_handler ~= nil then
+        clock.tempo_change_handler(bpm, source)
```
artfwo, line 222: What's the reason for passing source as an argument here? It could be easily retrieved as a param value if necessary, could it?

artfwo (commented):
Thanks @dndrks , I've got a couple more comments, wdyt?

lua/core/clock.lua
dndrks: for sure! i think i might actually have subconsciously been angling for adding a `clock.source_change_handler` as well -- would that be too much, or would it be a good completion of the callback set?

lua/core/clock.lua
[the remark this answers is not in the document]
dndrks: oh, awesome! thank you for pointing this out, super helpful -- i think it'd be great to add the other callbacks with this method, yes!

lua/core/clock.lua
artfwo: > i do think i might actually have subconsciously been angling for a clock.source_change_handler as well -- would that be too much to add, or would it be a good completion of the callback set?

yep, i've thought about it too, but can't imagine a use-case for the source-changed kind of callback at the moment, as the clock API is largely source- and tempo-agnostic by design anyway.

so i'd suggest sticking with a more minimalist approach here and leaving the tempo-changed callback having only the bpm parameter, and if a source-changed callback is necessary in the future - add it later at some point, wdyt?

awwaiid:
https://github.com/monome/norns/pull/1573 might be another approach, it allows chaining on top of existing param actions. The example on the lines forum would be something like:

```lua
params:append_action("clock_tempo", function ()
  -- Do some other stuff
  -- or even here call do_some_other_stuff()
end)
```

dndrks:
@artfwo , finally got the brain space to make these proposed changes -- please lmk if there's anything i'm misunderstanding?

artfwo (approved):
:+1:

artfwo:
> @artfwo , finally got the brain space to make these proposed changes -- please lmk if there's anything i'm misunderstanding?

approved, looks good!

Issue #1576, opened 2022-06-19
util.wrap ldoc: clarify integer-only (or add float support)


dndrks:
this was brought up in the norns study group discord -- [`util.wrap` and `util.wrap_max`](https://github.com/monome/norns/blob/b0e40ba7d475189947345737c0618da58d8cab5d/lua/lib/util.lua#L249-L291) are designed for integer input, but since Lua only supports one kind of _number_  (as floating point), this can confuse usage -- eg. `util.wrap(3.4,0,1)` returns `1.04`.

either the ldoc could specify that these utilities are for integer input only, or the functions could be expanded to support `numbers` as either integers or floats.

PR #1577, opened 2022-06-19
Notate that wrap accepts integers; reimplement to be constant-time

sixolet:
fixes https://github.com/monome/norns/issues/1576

### DIFF
--- a/lua/lib/util.lua
+++ b/lua/lib/util.lua
@@ -1,4 +1,4 @@
---- Utility module
+-- Utility module
 -- @module lib.util
 
 util = {}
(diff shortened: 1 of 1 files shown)


tyleretters:
@sixolet initial review looks great! i can test in the next 24 hours or so.

are the crow commits supposed to be part of a separate PR?

sixolet:
Okay, I did a little bit of git surgery to remove the weird merge commits or the commits that I already PRed or whatever happened there.

tyleretters:
my test results look good! the behavior is 1:1. i approve the merge. (i don't have the permissions actually approve it on github.)

@ngwese there are two concerns - one is clarifying is for integers. i will let @sixolet speak to the constant-time / performance.

catfact:
i provided the `while` in demonstration code as an alternative to a recursive version of the same thing. i don't claim its optimal and it is indeed only for integers (thank you.) 

regarding performance i'll note that the looped `wrap` is faster when it only needs to wrap once or twice, which is most of the time. lua's modulo operator has several hidden costs: metamethod lookup, two branches, and operating on doubles.

sixolet:
I think they were merged already. I can do some git surgery to extract
them. I just switched to ff-only, I think the problem was I was merging
from upstream.

tehn (approved):

PR #1578, opened 2022-06-19
Modulation

sixolet:
ROUGH DRAFT ROUGH DRAFT ROUGH DRAFT

This is a rough draft of what I mean in https://github.com/monome/norns/issues/1248#issuecomment-1159587877

It allows you to map to a "modulation", which will be additive with the pset's setting for the parameter and all other modulations on the same parameter. This will allow you to keep your midi mappings as you edit the value of the parameter, and it will also allow you to more easily make mods that do things like apply LFOs to arbitrary parameters. In fact, if we merge something like this, the next thing I'll do is make an awesome LFO mod.

### DIFF
--- a/lua/core/menu/params.lua
+++ b/lua/core/menu/params.lua
@@ -371,12 +371,24 @@ m.enc = function(n,d)
           max = r[2]
         end
         if m.mpos == 8 then
-          pm.out_lo = util.clamp(pm.out_lo + d, min, max)
+	  if pm.modulation then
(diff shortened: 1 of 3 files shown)


sixolet:
Closing this, as it has been basically entirely implemented by the [matrix](https://github.com/sixolet/matrix) mod. If there's a path forward for incorporating some of that code into core (which is currently doubtful?) it should use the code there, which has a nice API and everything.

PR #1583, opened 2022-07-07
Screen gamma + "DISPLAY" setting menu.

tlubke:


tlubke:
see: https://github.com/monome/linux/pull/47

matron/src/hardware/screen.c
```diff
@@ -277,6 +277,54 @@ void screen_aa(int s) {
...
+    char s[75];
+    char cmd[75 + 40];
+    sprintf(s, "%04x %04x %04x %04x %04x %04x %04x "
+               "%04x %04x %04x %04x %04x %04x %04x %04x",
+               grayscale_table[ 1], grayscale_table[ 2], grayscale_table[ 3],
+               grayscale_table[ 4], grayscale_table[ 5], grayscale_table[ 6],
+               grayscale_table[ 7], grayscale_table[ 8], grayscale_table[ 9],
+               grayscale_table[10], grayscale_table[11], grayscale_table[12],
+               grayscale_table[13], grayscale_table[14], grayscale_table[15]);
+    sprintf(cmd, "echo '%s' > /sys/class/graphics/fb0/gamma", s);
```
ngwese, line 310: An upcoming architectural change to matron/crone requires that we avoid shelling out to run command in order to prevent xruns from occurring. This use of `system(...)` should be replaced with calls to simply `open(...)`, `write(...)`, and `close(...)` the file.

Additionally it would be good to emit a warning and fail gracefully if the expected path under `/sys` doesn't exist.

matron/src/hardware/screen.c
```diff
@@ -277,6 +277,54 @@ void screen_aa(int s) {
...
+    if (v < 0) {
+        v=0;
+    }
+    if (v > 31) {
+   	v=31;
+    }
+
+    char cmd[47 + 1];
+    sprintf(cmd, "echo '%04x' > /sys/class/graphics/fb0/precharge", v);
+    system(cmd);
```
ngwese, line 325: Similar to above. Error checking and file I/O instead of `system(...)`

ngwese (changes requested):
I have not tested due to the kernel dependencies.

One thing I would like to better understand is what the strategy is regarding compatibility. For better or worse existing scripts and the menu system have many hard coded "level" values. If this becomes a static system setting (gamma curve) then a few questions come to mind:

* To what degree is the main menu system affected visually by these settings?
* Is it possible to change the display settings to such a degree that it would be difficult/impossible for someone to navigate the menu to undo the changes? 
* If scripts are written which expect a specific gamma curve and those scripts are shared what are people expected to do, adjust the gamma curve ahead running the script and restore it after running?

Additional notes:
* The `SYSTEM > RESET` logic removes all settings files and restores the system configuration. The settings file for DISPLAY should also be removed
* @tehn - I'm thinking strategically it would probably be good to do a maintenance release and then possibly rebase this change on top of the converged branch before testing on device.

tehn:
> I'm thinking strategically it would probably be good to do a maintenance release and then possibly rebase this change on top of the converged branch before testing on device.

I agree, and I apologize for the delay on the release. Will prioritize it.

More broadly I'm starting to question the cost/benefit balance of this proposal. I absolutely agree that the current gamma is not ideal, which is why I initially jumped at this opportunity.

tlubke:
@ngwese 

> To what degree is the main menu system affected visually by these settings?

The main menu system is only affected when changing the "low" setting. This leads into your next question

> Is it possible to change the display settings to such a degree that it would be difficult/impossible for someone to navigate the menu to undo the changes?

I have considered this, and tested it. Again, if we chose to keep the precharge setting, changing "low" low enough makes the lower screen levels almost invisible. Not to the point that I wasn't able to get back to the page, because of the highlighting. I think a more reasonable range for "low" is 16-31, and then it should functionally be considered a "brightness" setting.

> If scripts are written which expect a specific gamma curve and those scripts are shared what are people expected to do, adjust the gamma curve ahead running the script and restore it after running?

I've also considered this because it was brought up as a backwards compatibility argument. I initially added screen function for norns script to control gamma, and for the system menu to stash/pop the settings. However, it quickly became more complicated than I think would be useful. And then, what if the user wants to stop scripts from changing the gamma? I think this conversation kind of pits users against script owners.

@tehn 
> More broadly I'm starting to question the cost/benefit balance of this proposal. I absolutely agree that the current gamma is not ideal, which is why I initially jumped at this opportunity.

I still believe the benefit of the proposal outweighs the cost. I've personally dealt with the frustration having to fine-tune screen levels and feeling like there is a lack of fidelity, especially when compared to grid LEDs. The same thing has been echoed by others in the discord, and the original lines topic. Personally, I think changing the driver to always have a 2.14 gamma would be the best approach because of how much that reduces the scope of the changes. I don't think any scripts would look worse or different to the point of undermining script owners' visions. However, I understand the concerns about compatibility.

matron/src/hardware/screen.c
tlubke: Will do. Same with below.

tehn:
On older devices we had a knob to control brightness--- I do think trusting the user is the right move.

I do support this path forward. In hopes of minimal disruption, perhaps we consider postponing a merge until the rest of the breaking changes are completed? (You mentioned 60fps!) By breaking changes I mean kernel update (not exactly the correct term)

tlubke:
> In hopes of minimal disruption, perhaps we consider postponing a merge until the rest of the breaking changes are completed? (You mentioned 60fps!) By breaking changes I mean kernel update (not exactly the correct term)

I agree. I might open a GitHub issue on screen improvements to keep progress organized. Should I close my PRs for now?

matron/src/hardware/screen.c
tlubke: Done in latest commit.

tlubke:
@ngwese 
> An upcoming architectural change to matron/crone requires that we avoid shelling out to run command in order to prevent xruns from occurring.

Does this hold true for the lua side as well? Thought about using something like the line below to control disabling the DISPLAY settings menu.

`os.execute("test -w /sys/class/graphics/fb0/gamma")`

ngwese:
@tlubke - yes. `os.execute` should be avoided but the `norns.system_cmd` function can be used as an alternative. Both commands are "expensive" in that ideally shelling out is a last resort. In the case of testing if a file such as `sys/class/graphics/fb0/gamma` exists and is writable it seems far more appropriate to simply use `io.open("/sys/class/graphics/fb0/gamma", "w")`?  I quickly tested on a random linux system and opening a non-existent file under `/sys/...` returned nil. If I was able to write to it I got a file object back.

tlubke:
Closing in favor of userspace SPI screen.

PR #1584, opened 2022-07-13
params: align 'add_separator' and 'add_group' flow with other paramtypes

dndrks:
while scripting tonight, i realized https://github.com/monome/norns/pull/1539 did not anticipate situations where an author might do the following:

```lua
params:add_number('my var', 'my var', 1, 127)
[...]
params:add_separator('my var')
```

though it's convention to use underscores within parameter ID's, the system supports ID instantiation with spaces, so there's a chance someone will accidentally clobber their useful parameter with a separator or group name. since separators and groups do not have ID's, only names, this seems like a good guardrail.

### DIFF
--- a/lua/core/params/group.lua
+++ b/lua/core/params/group.lua
@@ -6,10 +6,11 @@ Group.__index = Group
 
 local tGROUP = 7
 
-function Group.new(name, n)
+function Group.new(id, name, n)
   local g = setmetatable({}, Group)
-  g.name = name or "group"
-  g.n = n or 1
+  g.name = type(name) ~= "number" and name or (id or "group")
+  g.id = id or g.name
+  g.n = type(name) == "number" and name or (n or 1)
   g.t = tGROUP
   g.action = function() end
   return g
--- a/lua/core/params/separator.lua
+++ b/lua/core/params/separator.lua
@@ -6,9 +6,10 @@ Separator.__index = Separator
 
 local tSEPARATOR = 0
 
-function Separator.new(name)
+function Separator.new(id,name)
   local s = setmetatable({}, Separator)
-  s.name = name or ""
+  s.name = name or (id or "separator")
+  s.id = id or s.name
   s.t = tSEPARATOR
   s.action = function() end
   return s
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -142,7 +142,8 @@ function ParamSet:add(args)
     end
   end
 
-  if self.lookup[param.id] ~= nil and (param.id ~= "group" and param.id ~= "separator") then
+  local overwrite = true
+  if self.lookup[param.id] ~= nil and param.t ~= 0 and param.t ~= 7 then
     print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
     print("!!!!! ERROR: parameter ID collision: ".. param.id)
     print("! please contact the script maintainer - this will cause a load failure in future updates")
@@ -152,6 +153,12 @@ function ParamSet:add(args)
     else
       print("! BEWARE! clobbering a script or mod param")
     end
+  elseif self.lookup[param.id] ~= nil and param.t == 0 and params:lookup_param(param.id).t ~= 0 then
+    print("! separator ID <"..param.id.."> collides with a non-separator parameter, will not overwrite")
+    overwrite = false
+  elseif self.lookup[param.id] ~= nil and param.t == 7 and params:lookup_param(param.id).t ~= 7 then
+    print("! group ID <"..param.id.."> collides with a non-group parameter, will not overwrite")
+    overwrite = false
   end
 
   param.save = true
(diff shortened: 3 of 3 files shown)


tehn:
is this the right fix? if there is a copy of the name, then it works but doesn't have an entry in the lookup? rather should it just fail and issue an error?

dndrks:
for sure! i was conflicted on what would be the easiest most-right fix, honestly. all the other param types use the `:add` method, where these conflicts are caught, but `add_separator` and `add_group` manipulate state external to the `:add` method.

though i'm a little gun shy about adding any fails / error messages to the params after the conflicts earlier this year, i popped a few into the latest commit. full fail doesn't feel necessary, since most cases of `params:lookup_param` seem to be motivated by `.action` manipulation anyway -- having an error and no overwrite to the previous instance of the action-oriented ID when a separator or group tries to use it feels a lot more informative + actionable!

dndrks:
> i wonder if the option to add a unique ID to separators and groups could help clear up these cases?

got to spend some time on this and it works! i ended up changing `:add_separator` and `:add_group` to use the standard `:add` method (like all the other parameter builders), so that ID conflicts could be accurately tallied and errors printed. this required adhering to the `name` and `id` requirement of all other paramtypes, which required placing some guardrails for expected historical invocation:

## separators
- if a separator is added using `params:add_separator(<name string>)`, then the separator's ID will be the `<name string>`
  - if another separator shares the same string, or another variable uses that string as its ID, maiden will post the collision warning
- if a separator is added using `params:add_separator(<id_string>, <name string>)`, then it registers like any other parameter to the lookup **(preferred)**
- if a separator is added using `params:add_separator()`, then the separator's ID will register as `'separator'`
  - if norns finds a conflicting `'separator'` ID entry in the lookup, it will _not_ post the collision warning (i'm assuming not identifying a separator in any way means it won't be manipulated)

## groups
- if a group is added using `params:add_group(<name string>, <count>)`, then the group's ID will be the `<name string>`
  - if another group shares the same string, or another variable uses that string as its ID, maiden will post the collision warning
- if a group is added using `params:add_group(<id_string>, <name string>, <count>)`, then it registers like any other parameter to the lookup **(preferred)**
- if a group is added using `params:add_group()`, then the group's ID will register as `'group'` and it will have a default group of `1`
  - if norns finds a conflicting `'group'` entry in the lookup, it will _not_ post the collision warning (i'm assuming not identifying a group in any way means it won't be manipulated)

<details>
<summary> example script </summary>

```lua
function init()
  
  print("1 --------")
  -- will not error, historical use:
  params:add_group('test group one',1)
  params:add_separator('1')
  print('historical use will not error')
  
  print("2 --------")
  -- will error, as ID's are same
  params:add_group('name with spaces') -- if just name is provided, this becomes the ID
  params:add_number('name with spaces', 'name with spaces') -- ID is same as group!
  print("   ^ parameters with identical ID's will cause error")
  
  print("3 --------")
  -- will error, as group ID's conflict:
  params:add_group('test group two', 1)
  params:add_separator('three')
  params:add_group('test group two', 1)
  params:add_separator('four')
  print("   ^ group tests with identical names *did* error because they do not have proper ID's")
  
  print("4 --------")
  -- will not error, group ID's provided:
  params:add_group('g_test_1', 'test group one',1)
  params:add_number('number_1a','1a', 0, 100, math.random(100))
  params:add_group('g_test_2', 'test group one',1)
  params:add_number('number_1b','1b', 0, 100, math.random(100))
  print("group tests with identical names did not error because they have proper ID's")
  
  print("5 --------")
  -- will error, as separator id's conflict:
  params:add_separator('five')
  params:add_separator('five')
  print("   ^ separator tests with identical ID's error like any other parameter")
  
  print("6 --------")
  -- will not error, as separator id's are different:
  params:add_separator('sep_fiveA', 'five')
  params:add_separator('sep_fiveB', 'five')
  print("separator tests with identical names did not error because they have proper ID's")
  
  print("7 --------")
  -- will not error, as default 'separator' ID is skipped:
  params:add_separator()
  params:add_separator()
  print("separators without ID's are given 'separator' as an ID, but norns will not call it out as an error")
  
  print("8 --------")
  -- will not error, as default 'group' ID is skipped:
  params:add_group()
  params:add_separator('example_8a','hello')
  params:add_group()
  params:add_separator('example_8b','good to see you')
  print("groups without ID's are given 'group' as an ID, but norns will not call it out as an error")
end
```
</details>

please feel free to lmk if this all feels aligned or if there's more to consider! i just wanna get all param types using the same basic building blocks, totally open to any redirects / feedback -- tyty!

dndrks:
ok, ready for real now -- sorry for the extra traffic!

tehn (approved):

dndrks:
had a quick thought, @tehn -- entering MAP mode on params surfaces the param ID for each parameter, but the current code [excludes](https://github.com/monome/norns/blob/2e2164e0ffc713ccbae5316fa83f5e254ee60952/lua/core/menu/params.lua#L468-L476) separators and groups from this changeover. do we want to retain this behavior now that groups and separators _can_ have ID's? if only a single string is provided at instantiation (eg. historical use of passing a name only), the ID defaults to the name, so shouldn't cause any backwards compatibility issues and i could see an author using the MAP mode to double-check their parameter IDs populate correctly.

just lmk which way you feel we should go and i'll propose the small change in another PR if you wanna go ahead and adjust!

PR #1585, opened 2022-07-13
add parameter-based lfo scripting library

dndrks:
there have been a _ton_ of wonderful approaches to infusing LFOs into scripts, eg. @markwheeler's `changes` and @justmat's `hnds`, which has prompted a recurring feature request for 'global LFOs' in the parameters menu. supporting this with a mod or something similar seems like a good solution, but gets a little fuzzy when considering how the mod would narrow to the most 'meaningful' parameters, rather than presenting the artist with _all_ of the parameters added by a script.

so, as an alternative solution, i built out a scripting library for parameter-based LFOs which an author can quickly deploy after adding their script's parameters. i've been sitting with this for a while and gone through a lot of different syntax decisions -- i think what's here feels familiar to the parameters system, but absolutely open to any and all notes from y'all!

### general flow

the general flow is that once a script's parameters are built, you can register up to 8 parameter ID's to a single LFO group and assign custom callbacks to the returned values (optional).

<details>
<summary> example script </summary>

in this example, the `my_number_<i>` parameters will have their actions called, but their params menu / `params:get()` values will remain the same. this is desirable in situations where you might want to return to the current state when disengaging the LFO.

```lua
-- my script!

lfo = require 'param-lfo'

function init()
  params:add_separator('numbers')
  number_values = {}
  for i = 1,8 do
    params:add_number('my_number_'..i, 'num '..i, 0, 128, 63)
    params:set_action('my_number_'..i,
      function(val)
        number_values[i] = val
        screen_dirty = true
      end
    )
    
    -- register control of 'my_number_<i>' to an LFO in the 'number LFOs' group:
    -- lfo:register(param, parent_group, fn [optional])
    lfo:register('my_number_'..i, 'number LFOs')
    
  end
  
  -- add the 'number LFOs' group to the UI parameters menu:
  -- lfo:add_params(parent_group, separator_name [optional], silent [optional])
  lfo:add_params('number LFOs')
  
  redraw_timer = metro.init(
    function()
      if screen_dirty then
        redraw()
        screen_dirty = false
      end
    end,
    1/15,
    -1
  ):start()
end

function redraw()
  screen.clear()
  for i = 1,8 do
    screen.move(5+(15*(i-1)),util.linlin(0,128,64,10,number_values[i]))
    screen.text(i)
  end
  screen.update()
end
```

</details>

### retro-fitting

the library will dynamically format `min` / `max` ranges, presentation, and awareness of any additional labels based on the registered parameter. this makes adding LFOs to other scripts pretty approachable.

for example, if i wanted to add LFOs to `awake`'s synthesis parameters, i'd just need to add:

- `lfo = require 'param-lfo'` at [the top](https://github.com/tehn/awake/blob/0de3c1480fb173a4fc421972d9d4840ad754b323/awake.lua#L34)
- after [the script's parameters are built](https://github.com/tehn/awake/blob/0de3c1480fb173a4fc421972d9d4840ad754b323/awake.lua#L330):
  ```lua
  lfo:register('amp', 'SYNTH')
  lfo:register('pw', 'SYNTH')
  lfo:register('cutoff', 'SYNTH')
  lfo:register('pan', 'SYNTH')
  lfo:add_params('SYNTH', 'LFOs')
  ```

this will create an `LFOs` separator at the bottom of the parameters list, with a`SYNTH` group which holds LFOs for `amp`, `pw`, `cutoff`, and `pan`. once you engage an LFO's `depth` parameter past 0%, you'll hear its affect.

### additional notes

- each LFO group consumes a [metro](https://monome.org/docs/norns/reference/metro) -- so if the script doesn't use any other metros then you can feasibly instantiate up to 240 LFOs (8 per group x 30 metros)
- though a group's LFOs share a single metro, they each have different rates -- `clocked` rates go from 1/16th to 1024 while `free` rates go from 0.001hz to 4hz.
- three shapes: `sine`, `square` and `random`
- each LFO can center its `position` on up to four different states:
  - `from min`: `depth` calculates from the `lfo min`
  - `from center`: `depth` calculates from the center point of the `lfo min` and `lfo max`
  - `from max`: `depth` calculates from the `lfo max`
  - `from current`: `depth` calculates from the registered parameter's current value -- this allows the registered parameter to be MIDI-mapped and change _as_ the LFO runs
- the `reset lfo` action can be used to reset the lfo's position to its lowest or highest point (`floor` or `ceiling`)
- by default, the LFO will not affect the parameter's stored value -- use `'map param'` as the optional `fn` argument in the `:register` command to enable this
  - if the LFO is turned off (or its `depth` is set to 0%) in `'map param'` state, it will not be able to restore the parameter's pre-LFO value, as it will be overwritten by the LFO movement
  - `'map param'` mode will remove `from current` as a `position` option, since the parameter's current value will constantly change
- the following param types are supported: `number`, `option`, `control`, `taper`, `binary`

oh! and since all the LFO distribution management is done inside of the library, the scripting author just has to initiate it once and register the parameters -- versus something like `pattern_time` where the author needs to allocate variables for the different instances. this means that the module's state needs a hard-reset between script loads, so i added the following to `script.lua`:

```lua
if package.loaded['param-lfo'] then
  package.loaded['param-lfo'] = nil
end
```

works very well in testing, but open to other ideas!

### DIFF
--- a/lua/core/params/group.lua
+++ b/lua/core/params/group.lua
@@ -6,10 +6,11 @@ Group.__index = Group
 
 local tGROUP = 7
 
-function Group.new(name, n)
+function Group.new(id, name, n)
   local g = setmetatable({}, Group)
-  g.name = name or "group"
-  g.n = n or 1
+  g.name = type(name) ~= "number" and name or (id or "group")
+  g.id = id or g.name
+  g.n = type(name) == "number" and name or (n or 1)
   g.t = tGROUP
   g.action = function() end
   return g
--- a/lua/core/params/separator.lua
+++ b/lua/core/params/separator.lua
@@ -6,9 +6,10 @@ Separator.__index = Separator
 
 local tSEPARATOR = 0
 
-function Separator.new(name)
+function Separator.new(id,name)
   local s = setmetatable({}, Separator)
-  s.name = name or ""
+  s.name = name or (id or "separator")
+  s.id = id or s.name
   s.t = tSEPARATOR
   s.action = function() end
   return s
--- a/lua/core/paramset.lua
+++ b/lua/core/paramset.lua
@@ -97,40 +97,6 @@ function ParamSet.new(id, name)
   return ps
 end
 
---- add separator.
--- name is optional.
--- separators have their own parameter index and
--- can be hidden or added to a parameter group.
--- @tparam string name
-function ParamSet:add_separator(name)
-  local param = separator.new(name)
-  table.insert(self.params, param)
-  self.count = self.count + 1
-  self.group = self.group - 1
-  self.hidden[self.count] = false
-  if name ~= nil then
-    self.lookup[name] = self.count
-  end
-end
-
---- add parameter group.
--- groups cannot be nested,
--- i.e. a group cannot be made within a group.
--- @tparam string name
--- @tparam int n
-function ParamSet:add_group(name,n)
-  if self.group < 1 then
-    local param = group.new(name,n)
-    table.insert(self.params, param)
-    self.count = self.count + 1
-    self.group = n
-    self.hidden[self.count] = false
-    self.lookup[name] = self.count
-  else
-    print("ERROR: paramset cannot nest GROUPs")
-  end
-end
-
 --- add generic parameter.
 -- helper function to add param to paramset
 -- two uses:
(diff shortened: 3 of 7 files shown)


tehn:
this looks great!

i have a minor concern with naming. this is a deeply param-centric LFO library, whereas i could imagine/anticipate a more generic scripting-focused LFO library.

would you be up for considering calling it param-lfo or paralfo or plfo or plolfo plololofo (or something) in order to reserve the minimal "lfo" for a prospective very-basic lib?

my second idea was to extend this lib for some more non-param-centric uses, but i think that'd clutter it up.

dndrks:
> would you be up for considering calling it param-lfo or paralfo or plfo or plolfo plololofo (or something) in order to reserve the minimal "lfo" for a prospective very-basic lib?

for sure! just pushed changes to rename it `param-lfo`, which feels a lot clearer.
i also changed the naming for the auto-added parameters to `p_lfo_[...]` instead of `lfo_[...]`. hopefully that'll avoid any conflicts with a future non-param general purpose lfo library :)

also, updated the code to take advantage of the separator and group ID specifications that https://github.com/monome/norns/pull/1584 added, so there will be no startling messages when retro-fitting this library to an existing script in the future.

thank you for taking a look at everything + for the positive feedback -- please lmk if i can adjust anything else!

PR #1586, opened 2022-07-22
Call the global Midi.remove function on removal

ngwese:
This change adjusts #1562 to align with what is understood to
be the original intent of the `Midi.add` and `Midi.remove` callbacks.


ngwese:
Actually - looking at this it could represent a breaking change so more investigation is needed.

PR #1591, opened 2022-07-30
add 'lfo'

dndrks:
`lfo` is a general-purpose scripting library for establishing LFOs, with optional parameter UI control.
it was inspired by the fantastic contributions over the years from @markwheeler , @justmat + @sixolet.

@sixolet brought the lib into a lattice structure, which means it only consumes one system clock to instantiate as many LFOs as you want, with settable `ppqn` per LFO to manage processing.

### basic use in a script

instantiating `lfo` and building LFOs should feel pretty similar to scripting with any of the current libraries:

```lua
_lfos = require 'lfo' -- assign the library to a general variable
engine.name = 'PolyPerc'
s = require 'sequins'

function init()
  hz_vals = s{400,600,200,350}
  sync_vals = s{1,1/3,1/2,1/6,2}
  clock.run(iter)
  
  -- establish an LFO variable for a specific purpose:
  cutoff_lfo = _lfos.new(
    'saw', -- shape
    200, -- min
    1300, -- max
    1, -- depth (0 to 1)
    'clocked', -- mode
    6, -- period (in 'clocked' mode, represents beats)
    -- pass our 'scaled' value (bounded by min/max and depth) to the engine:
    function(scaled, raw) engine.cutoff(scaled) end -- action, always passes scaled and raw values
  )
  
  cutoff_lfo:start() -- start our LFO, complements ':stop()'
end

function iter()
  while true do
    clock.sync(sync_vals())
    hertz = hz_vals()
    engine.hz(hertz)
  end
end
```

### advanced use in a script

to go further, there are additional `:set` and `:get` methods, which connect to the individual LFO's current state.

`:set` and `:get` options:
- 'enabled': 1 or 0
- 'shape': 'sine', 'saw', 'square', 'random'
- 'min': number
- 'max': number
- 'depth': number 0.0 to 1.0
- 'offset': number -1.0 to 1.0
- 'mode': 'clocked' or 'free'
- 'period': number (in 'clocked', beats; in 'free', hz)
- 'baseline': 'from min', 'from center', 'from max'
- 'reset_target': 'floor' or 'ceiling' (essentially should the reset start at min or max?)
- `ppqn`: the resolution of the LFO, defaults to 96 but can be brought down (ideally in equal divisions of 96) to reduce consumption
- 'action': the callback function, which receives both the scaled (to min/max * depth) and raw value (0 to 1)

special `:get` options:
- 'scaled': returns the scaled value of the LFO (scaled to min and max, adjusted by depth and offset)
- 'raw': returns the 0 to 1 raw value of the LFO, without considering the other adjustments

```lua
_lfos = require 'lfo' -- assign the library to a general variable
engine.name = 'PolyPerc'
s = require 'sequins'

function init()
  hz_vals = s{400,600,200,350}
  sync_vals = s{1,1/3,1/2,1/6,2}
  clock.run(iter)
  
  screen_dirty = true
  
  -- establish an LFO variable for a specific purpose:
  cutoff_lfo = _lfos.new()
  cutoff_lfo:set('shape', 'sine')
  cutoff_lfo:set('min', 200)
  cutoff_lfo:set('max', 5000)
  cutoff_lfo:set('depth', 0.3)
  cutoff_lfo:set('mode', 'free')
  cutoff_lfo:set('period', 2)
  cutoff_lfo:set('action', function(scaled,raw) engine.cutoff(scaled) screen_dirty = true end)
  
  redraw_screen = metro.init(check_dirty,1/15,-1)
  redraw_screen:start()
end

function iter()
  while true do
    clock.sync(sync_vals())
    hertz = hz_vals()
    engine.hz(hertz)
    cutoff_lfo:set('depth', math.random())
  end
end

function check_dirty()
  if screen_dirty then
    redraw()
    screen_dirty = false
  end
end

function redraw()
   screen.clear()
   screen.level(15)
   screen.move(64,40)
   screen.font_size(20)
   screen.text_center(util.round(cutoff_lfo:get('scaled'),0.01)..'hz')
   screen.update()
end

-- press K3 to start/stop:
function key(n,z)
  if n == 3 and z == 1 then
    if cutoff_lfo:get('enabled') == 1 then
      cutoff_lfo:stop()
    else
      cutoff_lfo:start()
    end
  end
end

-- turn E3 to adjust cutoff when LFO is inactive:
function enc(n,d)
  if n == 3 and cutoff_lfo:get('enabled') == 0 then
    local current = cutoff_lfo:get('scaled')
    local change = util.clamp(current + d*100, cutoff_lfo:get('min'), cutoff_lfo:get('max'))
    cutoff_lfo:set('scaled', change)
    engine.cutoff(cutoff_lfo:get('scaled'))
    screen_dirty = true
  end
end
```

## parameter UI

to simplify the process of changing variables during play, you can also just instantiate a parameter menu entry for any of the LFOs you construct. in order for the parameter menu to scale appropriately, it simply needs the `min` and `max` bounds for the LFO set ahead of executing `:add_params`:

```lua
_lfos = require 'lfo' -- assign the library to a general variable
engine.name = 'PolyPerc'
s = require 'sequins'

function init()
  hz_vals = s{400,600,200,350}
  sync_vals = s{1,1/3,1/2,1/6,2}
  clock.run(iter)
  
  screen_dirty = true
  
  -- establish an LFO variable for a specific purpose:
  cutoff_lfo = _lfos.new()
  -- IMPORTANT! set your LFO's 'min' and 'max' *before* adding params, so they can scale appropriately:
  cutoff_lfo:set('min', 200)
  cutoff_lfo:set('max', 5000)
  -- now we can add params:
  cutoff_lfo:add_params('cutoff_lfo', 'cutoff', 'LFOs')
  
  cutoff_lfo:set('action', function(scaled, raw) engine.cutoff(scaled) end)
  
  redraw_screen = metro.init(check_dirty,1/15,-1)
  redraw_screen:start()
end

function iter()
  while true do
    clock.sync(sync_vals())
    hertz = hz_vals()
    engine.hz(hertz)
  end
end

function check_dirty()
  if screen_dirty then
    redraw()
    screen_dirty = false
  end
end

function redraw()
   screen.clear()
   screen.move(64,40)
   screen.font_size(20)
   screen.text_center(util.round(cutoff_lfo:get('scaled'),0.01)..'hz')
   screen.update()
end
```

### DIFF
--- a/lua/lib/lfo.lua
+++ b/lua/lib/lfo.lua
@@ -62,6 +62,15 @@ function LFO.new(shape, min, max, depth, mode, period, action)
   return i
 end
 
+--- construct an LFO via table arguments
+-- eg. my_lfo:add{shape = 'sine', min = 200, max = 12000}
+-- @tparam[opt] string shape The shape for this LFO (options: 'sine','saw','square','random'; default: 'sine')
+-- @tparam[opt] number min The minimum bound for this LFO (default: 0)
+-- @tparam[opt] number max The maximum bound for this LFO (default: 1)
+-- @tparam[opt] number depth The depth of modulation between min/max (range: 0.0 to 1.0; default: 0.0)
+-- @tparam[opt] string mode How to advance the LFO (options: 'clocked', 'free'; default: 'clocked')
+-- @tparam[opt] number period The timing of this LFO's advancement. If mode is 'clocked', argument is in beats. If mode is 'free', argument is in seconds.
+-- @tparam[opt] function action A callback function to perform as the LFO advances. This library passes both the scaled and the raw value to the callback function.
 function LFO:add(args)
   local shape = args.shape == nil and 'sine' or args.shape
   local min = args.min == nil and 0 or args.min
--- a/lua/lib/lfo.lua
+++ b/lua/lib/lfo.lua
@@ -1,5 +1,5 @@
 -- LFOs for general-purpose scripting
--- @module lib.lfospec
+-- @module lib.lfo
 -- inspired by contributions from @markwheeler (changes), @justmat (hnds), and @sixolet (toolkit)
 -- added by @dndrks + @sixolet
 
@@ -15,6 +15,8 @@ local lfo_rates_as_strings = {"1/16","1/8","1/4","5/16","1/3","3/8","1/2","3/4",
 local params_per_entry = 14
 
 local rand_values;
+local percentage;
+local scaled_min, scaled_max, mid, centroid_mid;
 
 --- construct an LFO
 -- consumes one clock per LFO (norns has 100 clocks available for scripting)
@@ -23,9 +25,9 @@ local rand_values;
 -- @tparam[opt] number max The maximum bound for this LFO (default: 1)
 -- @tparam[opt] number depth The depth of modulation between min/max (range: 0.0 to 1.0; default: 0.0)
 -- @tparam[opt] string mode How to advance the LFO (options: 'clocked', 'free'; default: 'clocked')
--- @tparam[opt] number period The timing of this LFO's advancement. If mode is 'clocked', argument is in beats. If mode is 'free', argument is in Hz.
--- @tparam[opt] function fn A callback function to perform as the LFO advances. This library passes both the scaled and the raw value to the callback function.
-function LFO.new(shape, min, max, depth, mode, period, fn)
+-- @tparam[opt] number period The timing of this LFO's advancement. If mode is 'clocked', argument is in beats. If mode is 'free', argument is in seconds.
+-- @tparam[opt] function action A callback function to perform as the LFO advances. This library passes both the scaled and the raw value to the callback function.
+function LFO.new(shape, min, max, depth, mode, period, action)
   local i = {}
   setmetatable(i, LFO)
   i.scaled = 0
@@ -50,8 +52,19 @@ function LFO.new(shape, min, max, depth, mode, period, fn)
     wrap = false,
     formatter = nil
   }
-  i.action = fn == nil and (function(scaled, raw) end) or fn
+  i.action = action == nil and (function(scaled, raw) end) or fn
   return i
+end
+
+function LFO:add(args)
+  local shape = args.shape == nil and 'sine' or args.shape
+  local min = args.min == nil and 0 or args.min
+  local max = args.max == nil and 1 or args.max
+  local depth = args.depth == nil and 1 or args.depth
+  local mode = args.mode == nil and 'clocked' or args.mode
+  local period = args.period == nil and 4 or args.period
+  local action = args.action == nil and (function(scaled, raw) end) or args.action
+  self.new(shape, min, max, depth, mode, period, action)
 end
 
 -- PARAMETERS UI /
@@ -125,6 +138,48 @@ end
 
 -- SCRIPTING /
 
+local function scale_lfo(target)
+  if target.baseline == 'min' then
+    scaled_min = target.min
+    scaled_max = target.min + percentage
+    mid = util.linlin(target.min,target.max,scaled_min,scaled_max,(target.min+target.max)/2)
+  elseif target.baseline == 'center' then
+    mid = (target.min+target.max)/2
+    local centroid_mid = math.abs(target.min-target.max) * (target.depth/2)
+    scaled_min = util.clamp(mid - centroid_mid,target.min,target.max)
+    scaled_max = util.clamp(mid + centroid_mid,target.min,target.max)
+  elseif target.baseline == 'max' then
+    mid = (target.min+target.max)/2
+    scaled_min = target.max * (1-(target.depth))
+    scaled_max = target.max
+    mid = math.abs(util.linlin(target.min,target.max,scaled_min,scaled_max,mid))
+  end
+end
+
+local function change_bound(target, which, value)
+  target[which] = value
+  percentage = math.abs(target.min-target.max) * target.depth
+  scale_lfo(target)
+end
+
+local function change_baseline(target, value)
+  target.baseline = value
+  scale_lfo(target)
+end
+
+local function change_depth(target, value)
+  target.depth = value
+  percentage = math.abs(target.min-target.max) * value
+  scale_lfo(target)
+end
+
+local function change_ppqn(target, ppqn)
+  target.ppqn = ppqn
+  if target.pattern then
+    target.pattern:set_division(1/(4*ppqn))
+  end
+end
+
 local function process_lfo(id)
   local _lfo = id
   local phase
@@ -133,33 +188,26 @@ local function process_lfo(id)
   if _lfo.mode == "clocked" then
     phase = _lfo.phase_counter / _lfo.period
   else
-    phase = _lfo.phase_counter * clock.get_beat_sec() / (1/_lfo.period)
+    phase = _lfo.phase_counter * clock.get_beat_sec() / _lfo.period
   end
   phase = phase % 1
 
   if _lfo.enabled == 1 then
 
-    local current_val = (math.sin(2*math.pi*phase) + 1)/2
-    if _lfo.shape == "saw" then
+    local current_val;
+    if _lfo.shape == 'sine' then
+      current_val = (math.sin(2*math.pi*phase) + 1)/2
+    elseif _lfo.shape == 'saw' then
       current_val = phase < 0.5 and phase/0.5 or 1-(phase-0.5)/(0.5)
+    elseif _lfo.shape == 'square' then
+      current_val = phase < 0.5 and 1 or 0
+    elseif _lfo.shape == 'random' then
+      current_val = (math.sin(2*math.pi*phase) + 1)/2
     end
 
     local min = _lfo.min
     local max = _lfo.max
 
-    if min > max then
-      local old_min = min
-      local old_max = max
-      min = old_max
-      max = old_min
-    end
-
-    if min == -inf or max == -inf then
-      min = min == -inf and -2^1023.999999999 or min
-      max = max == inf and 2^1023.999999999 or max
-    end
-
-    local percentage = math.abs(min-max) * _lfo.depth
     if _lfo.shape ~= 'random' then
       _lfo.raw = current_val
     end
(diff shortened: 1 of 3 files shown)


lua/lib/lfospec.lua
```diff
@@ -0,0 +1,419 @@
...
+local rand_values;
+
+--- construct an LFO
+-- consumes one clock per LFO (norns has 100 clocks available for scripting)
+-- @tparam[opt] string shape The shape for this LFO (options: 'sine','saw','square','random'; default: 'sine')
+-- @tparam[opt] number min The minimum bound for this LFO (default: 0)
+-- @tparam[opt] number max The maximum bound for this LFO (default: 1)
+-- @tparam[opt] number depth The depth of modulation between min/max (range: 0.0 to 1.0; default: 0.0)
+-- @tparam[opt] string mode How to advance the LFO (options: 'clocked', 'free'; default: 'clocked')
+-- @tparam[opt] number period The timing of this LFO's advancement. If mode is 'clocked', argument is in beats. If mode is 'free', argument is in Hz.
```
sixolet, line 25: This might be tradition, but it feels a little strange to have a time-measure in one mode and a 1/time measure in the other. Maybe beats & seconds is better, since they're both time measures? I know for slow LFOs I always find myself taking the reciprocal in my head anyway if I'm trying to specify hz.

lua/lib/lfospec.lua
```diff
@@ -0,0 +1,419 @@
...
+-- inspired by contributions from @markwheeler (changes), @justmat (hnds), and @sixolet (toolkit)
+-- added by @dndrks
+
+local LFO = {}
+LFO.__index = LFO
+
+local lfo_rates = {1/16,1/8,1/4,5/16,1/3,3/8,1/2,3/4,1,1.5,2,3,4,6,8,16,32,64,128,256,512,1024}
+local lfo_rates_as_strings = {"1/16","1/8","1/4","5/16","1/3","3/8","1/2","3/4","1","1.5","2","3","4","6","8","16","32","64","128","256","512","1024"}
+
+local update_freq = 96
```
sixolet, line 12: If update_freq is constant, it might make sense (like lattice) to have all LFOs share a clock. Or... use a lattice?

Separately, it might make sense to have update_freq settable. I know that 96ppqn can be a lot if the LFO is triggering a lot on each step.

sixolet (commented):
I like this! It'll be generally useful, and then specifically useful, with adding params.

lua/lib/lfospec.lua
dndrks: ah yeah, both points make a lot of sense -- thank you! using a lattice to manage the complexity of having groups of LFOs share a clock in the library itself (which the first draft of this that was reverted did, but at the expense of scripting legibility) would be good

lua/lib/lfospec.lua
dndrks: def pulling from tradition here -- @tehn , what do you think?

lua/lib/lfospec.lua
sixolet: Yeah, I can probably give it a try later today. Shouldn't be that much of a change. There's some options here, tho, the design space is pretty big:
* Reintroduce the concept of an LFO group, LFO groups share a clock/lattice and a ppqn?
* Leave LFOs on separate clocks, each one picks its ppqn?
* One global ppqn, but settable by the scripter, all LFOs share the same clock?
* One global "base" ppqn, all LFOs on the same clock, (96 ppqn, perhaps not even settable) but each LFO can pick a "bang divider" which allows it to update less frequently — you might pick 2 for a 48ppqn option, or even something like 96 for an LFO that has a period of several minutes.
* ~~LFOs will pick a "bang divider" depending on their frequency, aiming for a resolution that makes sense for them?~~ (This one seems like it'd be a little hard to pick a "goldilocks" value for every context, maybe let's not pick this) 

Which seems to fit best? I think I gently favor the "bang divider" option in that you can just use the lib and it'll bang at 96 like the current PR, but if you find it is overloading your system you can slow its roll pretty easily. And I don't imagine many situations where you'd want to slow its roll by a factor of 1.5 or whatever, slowing the roll by a factor of 2 or 3 or 4 makes a lot more sense. My one worry is what to call the "bang divider" argument, "bang divider" isn't immediately obvs what it means.

dndrks: love your solution of leveraging lattice and using `:set_division` to manage unique ppqn's per LFO -- and this way, just consumes one clock!

dndrks:
rad, thank you @sixolet for all the help migrating to a `lattice` scaffolding + adding the `ppqn` management -- feels _really_ solid + capable of running pretty lean! super exciting <3

quick q for the core devs: there isn't really a precedent for using `lattice` within other main libraries and since clocks are all cleaned up during script clear, this required naomi to get creative with nuking + re-establishing the `lfospec` lattice: https://github.com/monome/norns/blob/lfospec/lua/lib/lfospec.lua#L228-L235

i found an alternate approach, which follows @pq's forced unloading of `asl` during script clear: https://github.com/monome/norns/blob/13bccdd3fcd3f0cb99aaadd87e72be6027842b6e/lua/core/script.lua#L24-L28

eg:
```lua
if package.loaded['lattice'] ~= nil then
  print('lattice still loaded')
  package.loaded['lattice'] = nil
end
```

this change to `script.lua` allows us to remove naomi's `hook`s, but it feels mysterious.
which path is preferred for a main repo library?

catfact:
i see why `asl` needs reloading - because it isn't designed as a proper module but instead is just free global functions which may have been clobbered by scripts. since the `require` mechanism will not re-execute the sourcce, it cannot be "unclobbered" by re-requiring.

i don't quite see why you need to do that for the `lattice` module or how it would avoid the necessity of re-initializing the lattice _object_ used by the mod. in other words, it seems perfectly clear to use the hooks for that.

catfact:
... yeah, although i haven't tried this on device yet (will tonight if childcare allows), i am still not seeing how re-executing the lattice source would work around the issue solved by the hooks, which is (as you say) that the lattice's "superclock" has been destroyed.

if @sixolet desires an alternate and maybe simpler workaround which preserves the object, i might try simply setting the lattice's `superclock_id` to `nil` so that the clock is recreated by these lines in `lattice:start`:
https://github.com/monome/norns/blob/13bccdd3fcd3f0cb99aaadd87e72be6027842b6e/lua/lib/lattice.lua#L33-L35

(again this is just theory.)

... in fact - i don't know exactly how this would be accomplished, but it would be _nice_ if a clock's destructor somehow alterted any `lattice` objects which are using it.

sixolet:
I’m perfectly happy using the hooks, as long as it’s safe and reasonable to
use them outside the mod framework it seems like they were designed for.

dndrks:
rad, thank you for the guidance and additional gut-checkin' @catfact !! i'm totally happy with the hooks -- they feel super legible and it's nice to model those scripting approaches in the main libraries.

catfact:
ah sorry i am being a doofus. saw this and somehow assumed i was looking at a mod (because of the hooks.)

yeah i agree - using the mod hooks is actually a little weird. at minimum it would need to deregister the hooks at some point.

but once you need to do that... why is a _library_ trying to do stuff automatically around the script's lifecycle? mods require the ability to inject themselves into the system logic independently of scripts, but libraries do not - they are used explicitly by scripts, and i think its best practice for libraries to have minimal global side effects.

so although it is very slightly more verbose, i think it is proper that `lfospec` follow the same struction as other library modules: 
- executing the library code should not directly effect system state;
- instead it should simply wrap some reusable functionality in a table.
- in this specific case i think it's best for a script to explicitly call whatever setup methods are needed for thse LFOs in its `init`, and teardown methods in its `cleanup`. 

(BTW it would be nice if `asl` followed the same patterns/assumptions as other library modules.)

the other option is to make this a core feature that is always-on, and simply call its setup/teardown from `core/script.lua`.

dndrks:
rad, thank you for the extra eyes + attention!!

> executing the library code should not directly effect system state;

just to confirm, this definitely isn't purposeful + i'm unsure where system state is being affected -- i tried to model it after `pattern_time`, originally using a clock-per-lfo approach, and the process was cleared with each script load since all clocks are nuked during cleanup. the `lattice` version allows for a lot more extensibility, but it seems like the superclock its assigned to at first boot is extending past the lib's intended lifecycle / scope.

i'm happy to put setup + cleanup onto the script author, but are there any obvious spots where the lib oversteps its intended 'load me when you want me, i'll die off when the script is cleared' functionality?

lua/lib/lfospec.lua
```diff
@@ -0,0 +1,449 @@
...
+
+    if min > max then
+      local old_min = min
+      local old_max = max
+      min = old_max
+      max = old_min
+    end
+
+    if min == -inf or max == -inf then
+      min = min == -inf and -2^1023.999999999 or min
```
tehn, line 158: use `math.huge` and `-math.huge` instead for legibility

also there is redundancy by having the `if` above this line (just remove the if check)

lua/lib/lfospec.lua
```diff
@@ -0,0 +1,449 @@
...
+      _lfo.action(_lfo.scaled, _lfo.raw)
+      if _lfo.parameter_id ~= nil then
+	params:set("lfo_scaled_".._lfo.parameter_id,util.round(_lfo.scaled,0.01))
+	params:set("lfo_raw_".._lfo.parameter_id,util.round(_lfo.raw,0.01))
+      end
+    end
+  end
+end
+
+hook['script_pre_init']:register("lfospec lattice setup", function()
```
tehn, line 228: I do think using the mod hook system here is peculiar. I think explicit `init`/`cleanup` functions would be preferable.

honestly I also think I need to understand the rationale behind using lattice vs. just the clock system? what does that provide additionally?

lua/lib/lfospec.lua
```diff
@@ -0,0 +1,449 @@
+-- LFOs for general-purpose scripting
+-- @module lib.lfospec
```
tehn, line 2: forgive me, why would this lib not just be called `lfo`? is there something in supercollider that suggests this should have `spec` added?

lua/lib/lfospec.lua
```diff
@@ -0,0 +1,449 @@
...
+  lb('shape')
+  lb('reset')
+  lb('reset_target')
+end
+
+-- / PARAMETERS UI
+
+-- SCRIPTING /
+
+local function process_lfo(id)
```
tehn, line 128: generally this function needs to be combed through to make sure you're not doing math-work that doesn't need to be done since there is a lot of per-mode calculation

catfact: +1. for example...

lua/lib/lfospec.lua
catfact: - in this per-tick `process_lfo` i see many calulations involving these `min`/`max` deriverd from `_lfo.min` etc., but that does not seem to be updated here. these intermediate calculations should really just be performed whenever the lfo bounds change....

lua/lib/lfospec.lua
```diff
@@ -0,0 +1,449 @@
...
+	mid = math.abs(util.linlin(min,max,scaled_min,scaled_max,mid))
+	value = util.linlin(0,1,scaled_max,scaled_min,current_val)
+      end
+
+      raw_value = util.linlin(min,max,0,1,value)
+
+      if _lfo.shape == "sine" then
+	value = util.clamp(value,min,max)
+	_lfo.scaled = value
+      elseif _lfo.shape == "saw" then
```
catfact, line 199: - i would refactor this whole conditional change into spearate functions per mode, and one function handle that is assigned when the mode changes.

lua/lib/lfospec.lua
catfact: - whoa! these `2^x` calls with constants don't work like they would in a compiled environemnt - it's really gonna recalculate these on each tick with 64b precision. just use constants

catfact:
>  i'm unsure where system state is being affected

the only one that is obvious to me is the mod hook which is never deregistered. but `LFO` has global state (the singular `lattice`) and its worth just being sure that this is really in a predictable condition when a script first asks to use the library. 

like, going back to my previous point, i'm not sure why there needs to be global `lattice`. most modules (including `pattern_time` which you mention) don't have global state, they are OOP style and only provide factory functions. exceptions are when we need global state to hook into other parts of the system that are long-running, but this should rarely if ever apply sto stuff outside of `lua/core`. 

it would seem more obvious to me to make a new lattice for each script that wants to use the functionality - though it does again require some care in the teardown step.

sixolet:
So the reason I rewrote it to use a lattice was so that all the lfos could
share a clock, and a place on the queue, without basically re-creating that
same functionality that lattice already offers. I’m… not completely sure
this is valuable, but I read a little of the discussion around the creation
of lattice, and was given to understand that that was one of the factors
that made lattice valuable. I used the script init and cleanup hooks
because I wanted the lattice lifecycle to match that of a script, and felt
like it might be “cleaner” than modifying Script, because the lifecycle
management could be placed in the same file as the library. The lattice is
started lazily upon lfo start, and I think I wrote it to stop at the last
lfo stop, so as not to take resources when not being used.

Anyway that’s the reasons I made a PR to Dan’s code; I’m not particularly
attached to any given way and I love learning from this conversation.

lua/lib/lfospec.lua
dndrks: agreed! changed

lua/lib/lfospec.lua
```diff
@@ -0,0 +1,449 @@
...
+--- construct an LFO
+-- consumes one clock per LFO (norns has 100 clocks available for scripting)
+-- @tparam[opt] string shape The shape for this LFO (options: 'sine','saw','square','random'; default: 'sine')
+-- @tparam[opt] number min The minimum bound for this LFO (default: 0)
+-- @tparam[opt] number max The maximum bound for this LFO (default: 1)
+-- @tparam[opt] number depth The depth of modulation between min/max (range: 0.0 to 1.0; default: 0.0)
+-- @tparam[opt] string mode How to advance the LFO (options: 'clocked', 'free'; default: 'clocked')
+-- @tparam[opt] number period The timing of this LFO's advancement. If mode is 'clocked', argument is in beats. If mode is 'free', argument is in Hz.
+-- @tparam[opt] function fn A callback function to perform as the LFO advances. This library passes both the scaled and the raw value to the callback function.
+function LFO.new(shape, min, max, depth, mode, period, fn)
```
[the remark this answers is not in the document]
dndrks: raaad, ty!

lua/lib/lfospec.lua
dndrks: oh duhhhh, thank you!

lua/lib/lfospec.lua
dndrks: omg, yes! changed all the min/max/depth/percentage calculations to static functions which only execute when changes actually occur.

dndrks:
hihi! i got all the changes discussed above into the library -- most important is removing the constant recalculation of static variables (like `min` and `max`) from the LFO loop.

i also made a (perhaps extreme) shift with how to manage the global `lattice` invocation -- when the library is run the library creates a temporary `norns.lfo` table, where the `lattice` lives. in `script.lua`, i added a check to see if this temporary variable exists and, if so, to deregister the lattice and make `norns.lfo = nil`. this seems like a ~decently non-invasive solution to running the single `lattice` meta-structure around all the lfo's a script might create, while not allowing it to extend past the script's life, while also not using mod functionality.

all of the feedback has been incredibly helpful -- i'm definitely walking away having learned the _immense_ importance of ensuring optimization goes past 'oh, but the thing works fine!'. thank you all so much!! please lmk if there are other spots i'm not seeing (or not yet able to see) <3

lua/lib/lfo.lua
```diff
@@ -15,6 +15,12 @@ local lfo_rates_as_strings = {"1/16","1/8","1/4","5/16","1/3","3/8","1/2","3/4",
 local params_per_entry = 14
 
 local rand_values;
+local percentage;
+local scaled_min, scaled_max, mid, centroid_mid;
+
+function LFO.init()
+  norns.lfo = {lattice = lattice:new()}
```
tehn, line 22: seems this should check if norns.lfo exists prior to creating it again

dndrks: oh! i just assumed the `script.lua` clearout would always ensure it didn't exist, but def good to protect against weirdness. i'll adjust now!

tehn:
indeed you're correct. it's a safeguard in case someone tries to make multiple lfo object classes (which would spawn multiple lattice, leaking the previous clock per). maybe not of concern as clocks get cleared

tehn (approved):
i think this looks good! let's merge and beta. we can fix up any issues if they come up.

great job!

dndrks:
eeee! tytyty @tehn + @catfact + @sixolet !!

PR #1592, opened 2022-07-30
Use lattice for LFOs, so they share clocks, and allow setting ppqn

sixolet:
This suggestion does one of the things I reccommended:

* Uses lattice instead of a raw clock, so the LFOs can easily share a clock
* Allows the user to specify the ppqn per-LFO. Values that divide 96 evenly are going to be better, because they won't leave residual phase in the lattice, but other values are okay — they just won't bang quite _exactly_ on the dot every time.
* I hope I have not been misusing the `hook` library, but it was useful.


dndrks (approved):
omg YES, you are a brilliant sweetheart, holy hell

PR #1606, opened 2022-10-16
expand acceptable accum CC values

zjb-s:
a tiny change that allows controllers to send any CC value above 64 for "up" and any below for "down". previously it had to just be 65 and 63 which isn't supported by all controllers.

### DIFF
--- a/lua/core/menu/params.lua
+++ b/lua/core/menu/params.lua
@@ -677,7 +677,7 @@ norns.menu_midi_event = function(data, dev)
         local d = norns.pmap.data[r]
         local t = params:t(r)
         if d.accum then
-          v = v - 64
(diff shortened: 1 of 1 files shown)


tehn (commented):
i'm perplexed at this--- do incremental controllers not send values beyond +1 and -1? for example, sending the value 66 should mean +2.

effectively the code change just clamps increments to values of 1, is that the correct expected behavior?

zjb-s:
mornin'! yeah, my faderfox ec4 by default sends 127 and 0 for +1 and -1 respectively. I know the Deluge expects those values for incremental control as well. I've never seen a controller send values beyond +1 and -1 but I'm no midi expert

PR #1607, opened 2022-10-18
add TAPE previewing to fileselect

### LINKED ISSUE #1581: Sample preview for lib.fileselect (FEATURE)
I was thinking that a sample preview feature for lib.fileselect would be a great addition to a lot of scripts. Especially Timber.

My idea of the process workflow would be:

-open load file with lib.fileselect
-turn E3(Or E1?) clockwise to initiate sample preview
-turn E3(Or E1?)  counterclockwise to stop sample preview
-everything else as is, K2 to select sample, K3 to cancel
-Might be cool to have a "turn Ex to preview" in the upper right corner

As far as the max length goes of sample preview I think 20 to 30 seconds should be enough? 
For what I'm doing musically I'd be fine with just previewing longer samples in Matron and just being able to preview shorter samples(under 20 seconds maybe) in Norns. But that's just for my purposes.

See this thread if you want to see the Lines thread that prompted me to post this feature request: 
https://llllllll.co/t/norns-sample-load-workflow-improvement/27296/18

ryleelyman:
This would close #1581

The operation is basically as proposed: E3 will turn on and off previewing, which hijacks TAPE.

Currently it's very bare-bones; so for instance if TAPE is muted, no sound will be heard.

Thoughts / suggestions / comments welcome :)

### DIFF
--- a/lua/lib/fileselect.lua
+++ b/lua/lib/fileselect.lua
@@ -142,11 +142,13 @@ fs.enc = function(n,d)
   elseif n==3 and d > 0 then
     fs.file = fs.list[fs.pos+1]
     if fs.lengths[fs.pos+1] then
-      fs.previewing = fs.pos
-      audio.tape_play_stop()
-      audio.tape_play_open(fs.path .. fs.file)
-      audio.tape_play_start()
-      fs.redraw()
+      if fs.previewing ~= fs.pos then
+        fs.previewing = fs.pos
+        audio.tape_play_stop()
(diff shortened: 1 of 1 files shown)


tehn (approved):
untested, but code looks good and love this feature!

ryleelyman:
added a couple of tweaks: previewing always stops on key now, since either we're exiting or we're changing folders, and always stops on `E3` CCW.

tehn:
very good observation, i recall thinking about that months ago and totally forgot.

PR #1614, opened 2022-10-24
Ugen install

ryleelyman:
Adds a `Script.ugen_helper()` function that iterates over the table of strings `engine.ugens`, checking for each whether a file of the form `name.sc` and `name_scsynth.so` is present at `~/.local/share/SuperCollider/Extensions/name/` and if not, attempts to copy from `/ignore` or `/lib/ignore`. If files are successfully copied, script load is blocked and an error message encouraging a `system > restart` is displayed.

Potential gotcha: if a "UGen" is just an `.sc` file without a corresponding `.so` file, using this method will prevent script load.  I don't know how frequently these "pseudo-UGens" get used qua UGens as opposed to bespoke script components, but it's something to be aware of.

I figured `Script` was a better place for this than `Engine`, but I'm happy to refactor if we don't agree.

### DIFF
--- a/lua/core/engine.lua
+++ b/lua/core/engine.lua
@@ -11,6 +11,8 @@ local Engine = {}
 Engine.names = {}
 -- currently loaded name
 Engine.name = nil
+-- ugens to be installed on first script load
+Engine.ugens = {}
 -- current command table
 Engine.commands = {}
 -- flag if there is a load request pending
--- a/lua/core/menu/home.lua
+++ b/lua/core/menu/home.lua
@@ -101,7 +101,7 @@ m.redraw = function()
     local line = string.upper(norns.state.name)
     if(_menu.scripterror and _menu.errormsg ~= 'NO SCRIPT') then
       line = "error: " .. _menu.errormsg
-      if util.string_starts(_menu.errormsg,"missing") then
+      if util.string_starts(_menu.errormsg,"missing") or util.string_starts(_menu.errormsg,"installed") then
         screen.level(8)
         screen.move(0,25)
         screen.text("try 'SYSTEM > RESTART'")
--- a/lua/core/script.lua
+++ b/lua/core/script.lua
@@ -220,7 +220,12 @@ Script.run = function()
   -- allow mods to do initialization
   hook.script_pre_init()
   -- attempt to install missing ugens
-  Script.ugen_helper()
+  local restart_flag = Script.ugen_helper()
+  if restart_flag then
+    norns.scripterror("installed new UGens")
+    Script.clear()
+    return
+  end
   print("# script run")
   if engine.name ~= nil then
     print("loading engine: " .. engine.name)
@@ -291,11 +296,7 @@ Script.ugen_helper = function()
       end
     end
   end
-  if flag then
-    norns.scripterror("installed new UGens")
-    Script.clear()
-    return
-  end
+  return flag
 end
 
 return Script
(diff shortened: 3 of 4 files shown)


tehn:
cool!

i'm curious about the choice of the name `ignore`? somehow i figure it'd make sense to just "install" things found in `lib` if needed?

looking through the rest...

ryleelyman:
oh, `ignore` is chosen for this reason: SuperCollider will avoid looking in those directories at startup, so this allows every script that wants to use a particular UGen to ship it with the script without getting the Duplicate Engines error on startup.

tehn:
apologies for the delay on this one. anyone else have comments?

catfact:
this has been languishing a long time. i have no problem with it but haven't really tested personally.

tehn:
likewise conceptually i have no problem with this, but would want a handful of interested parties to test it first

ryleelyman:
@sixolet @schollz sorry to ping—any interest in having something like this?

schollz:
Yes, definitely interested in this. Do you have a little demo example of it in action? I can test it out with some of my scripts

ryleelyman:
yeah, if I remember correctly, to test it out you would do something like this:
```lua
-- in your script
engine.name = "YrGreatNornsEngine"
engine.ugens = { "FormantTriPTR", "TrianglePTR", "AnotherUGen" }
```
and then in your script folder, you would place `FormantTriPTR.sc`, `FormantTriPTR_scsynth.so` and so on in a folder named `ignore`. the file structure needs to be either like so
```
dust/code/my_script
 |- lib
 |- ignore
 |     |-  FormantTriPTR.sc
 |     |-  FormantTriPTR_scsynth.so
```
or like the above, but with `ignore` inside of `lib`.
(in particular, the structure inside `ignore` must be completely flat.)
anyway, the UGen helper then attempts to copy any named UGens which are not present in, for example, `.local/share/SuperCollider/Extensions/FormantTriPTR/` into that location.

Notably, if somebody has installed a UGen into a different folder inside of `Extensions` than the one provided, people will get SuperCollider fails... would happily accept suggestions on how to be a bit more robust in that regard!

catfact:
Hey thanks for the description. So, actually reviewing with that in mind, it does seem a bit fragile. 

Not really a problem when you're distributing your own ugens. But for redistributing anything that might have come from somewhere else or been built + installed following some other instructions, it could fail:
- people can and do reorganize extensions arbitrarily under the support directory (mine is pretty messy for example)
- its easy to end up with things installed under /usr/local too

So ultimately it would be good to (1) use a recursive search, (2) check the system extensions location too.

Might as well do (2) I think. for the sake of forward momentum, i think it would be ok to hold off on (1), and maybe create a GH issue preemptively calling out the known limitations. Or otherwise document that this is basically a new convention that needs to be followed.

schollz:
I can PR a fix (unless someone gets to it first). I've implemented (1) and (2) in my own sc installer that I've been bundling in scripts. (relevant lines: https://github.com/schollz/scinstaller/blob/main/scinstaller.lua#L73-L95). it uses `find` to do a recursive search over these folders:

```
/usr/local/share/SuperCollider/Extensions
/home/we/dust/code
/home/we/.local/share/SuperCollider/Extensions
```

then pattern matching to look for relevant files.

sixolet:
The find command is what I do, too.

ryleelyman:
A PR to fix (1) and (2) would be very welcome, @schollz!

catfact:
hi i'm going to close this pending rework and because it is >1yr old. but i still think it is a good idea!

Issue #1612, opened 2022-10-20
clock.tempo_change_handler not responding to ableton link


sonocircuit:
When changing the tempo in ableton live norns follows the tempo change as expected, however, the `clock.tempo_change_handler` is not called. If the tempo is changed via norns it works as expected.

I used this test script to register if the handler get called:

```
-- test clock.tempo_change_handler()

function init()
end

function clock.tempo_change_handler(tempo)
  print("tempo is "..tempo.."bpm")
end
```

Also, if I print the tempo with ` print(clock.get_tempo())` within the handler, the bpm is offset by 1. However, printing the current tempo with `clock.get_tempo()` with the maiden REPL returns the correct value. Not sure if this is a bug or me overseeing something obvious.

sonocircuit:
thanks dan! this is great. looking forward to updating my norns later today.

PR #1617, opened 2022-11-09
send external source tempo to clock.tempo_change_handler

dndrks:
addresses issue #1612

adds a call to the script-defined `clock.tempo_change_handler` function whenever the `clock_tempo` parameter is changed in the parameters from an external source. tested with the snippet included in the issue.


artfwo (approved):
looks good dan!

PR #1620, opened 2022-11-19
proposal: prevent global menu shortcut messing up w/ sub-menu's

p3r7:
**before:** if a sub-menu defines its own `_menu.keycode` callback, it gets evaluated **in addition and after** all the **default global keycode mappings**.

these default mappings are nice for most menus but can prevent making custom specialized interactions for sub-menus that have a slightly different logic (mod screens, typically).

**after:** if a sub-menu defines its own `_menu.keycode`, it **takes precedence** over all of the default menu mappings (except the `F1`..`F4` shortcuts).

**only small drawback**: if this sub-menu wants to use some of the default mappings for itself, it needs to redefine them in its own `_menu.keycode` callback.

### DIFF
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -210,16 +210,33 @@ end
 
 -- global menu keys
 function _menu.keycode(c,value)
-  if value==1 then
-    if c=="F1" then _menu.set_page("MIX")
-    elseif c=="F2" then _menu.set_page("TAPE")
-    elseif c=="F3" then _menu.set_page("HOME")
-    elseif c=="F4" then _menu.set_page("PARAMS")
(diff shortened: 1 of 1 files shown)


p3r7:
full disclosure of my use-case: 

i have a fully working REPL for both maiden & sc as a mod.

i made the REPLs accessible through a norns mod sub-menu (instead of hijacking the script `redraw` function, too finicky) and you get to interact w/ it through the keyboard.

but the default keybindings for `LEFT` / `RIGHT` and `UP` / `DOWN` are messing up some of those it felt natural to define.

repl

tehn (approved):

tehn:
onboard repl?? wow!!

PR #1624, opened 2022-11-24
[mergeable] rework `gamepad` to handle more edge-cases

p3r7:
so upon testing w/ `@McNUG?` i discovered that gamepads are even wackier than what i originally anticipated.

indeed:
- analog **sensor ranges** are **not standardized** (a single controller can even use different sensors w/ different ranges)
- i knew that button **keycodes** where **all other the place**, but it turns out that **axis** too!

this slight rework makes the following change:
- we configure & **track each individual analog sensor range** & calibration (origin point, amount of noise)
- `gamepad.axis` and `gamepad.astick` now return a **per-sensor axis** as its 1rst arg (`dpady`/`dpadx`, `lefty`/`leftx` & `righty`/`rightx` instead of generic `X`/`Y`, `RX`/`RY` & `Z`/`RZ`) \*
- `gamepad.astick` now has its `val` **center value** normalized **around `0`** & takes a 3rd param `half_reso` corresponding to the max value for each half of the travel (i.e. `val` can take value between `[-half_reso ... half_reso]`)
- **new profile** built from an [8BitDo SN30 Pro](https://www.8bitdo.com/sn30-pro-usb-gamepad/) which is a high quality & readily available controller (unlike those of the previous profiles)

\* i reused the same "per-sensor axis" [naming convention](https://love2d.org/wiki/GamepadAxis) as the `LÖVE` lua game framework.

the [accompanying interactive gamepad profile generator script](https://github.com/p3r7/gamepad_wiz) has also been updated & robustified.

i still plan on:
- adding support for **analog shoulder buttons** (generally associated w/ either `L1`/`R1` or `L2`/`R2`).
- **renaming `gamepad.astick`** into something fitting any analog sensor (e.g. shoulder buttons). maybe simply `gamepad.analog`?
- adding support for loading **user-configured profiles** (stored maybe somewhere in `~/dust/data`?)

what edge-cases i don't plan to support but i'm pretty sure some crazy manufacturer has implemented:
- dpads using buttons keycodes instead of axis event
- analog stick using 2 axis keycode for each half axis


p3r7:
tl;dr: so i've got a cleaner way to identify controllers in the C layer, a GUID.

i still have a small bug: i expected to grab this GUID as a new arg to `Hid.new` (lua-side) but i must have missed something in the interop declaration as i get `nil` (while it's well set C-side when calling `w_handle_hid_add`). 

<details>
 <summary>details</summary>

> so i did some more research on how to best identify a controller.
>
> SDL has a notion of [controller GUID](http://wiki.libsdl.org/SDL2/SDL_JoystickGetGUIDString) which i hoped was something explicit in the HID protocol but is in fact [more of a hash](https://github.com/libsdl-org/SDL/blob/9a6bcca6b8db999217f157dab8c44bf27b0a8eb4/src/joystick/SDL_joystick.c#L1887).
>
> it turns out that SDL get the fields it uses to build this GUID from `libevent`. so we have access to those same fields & can [easily reconstruct this GUID](https://meghprkh.github.io/blog/posts/handling-joysticks-and-gamepads-in-linux/) (i've tested it on norns, it works) and adhere to the same standard as many other platforms.
>
> one promise would be to support a bunch of controllers by importing a [db of profiles](https://github.com/gabomdq/SDL_GameControllerDB) (but sadly this one doesn't reference the analog sensor ranges nor their noisiness...).
</details>

p3r7:
alright, added support for analog buttons (left/right shoulders typically).

i did a [semblance of doc](https://github.com/p3r7/gamepad_wiz#about-norns-gamepad-support).

i still need to do one or two test & add handling of analog buttons in my wizard but basically we're 99% there.

p3r7:
did a bunch of test. everything seem to be working as expected.

i tweaked some sub-menu integration code to feel more intuitive.

i'm not 100% positive about the menu callback names though (`_menu.button`, `_menu.dpad`, `_menu.axis`, `_menu.analog`). maybe should i rename them into something more "fully-qualified" (e.g. `_menu.gamepad_button`)?

p3r7:
ok, done for the prefixed callbacks.

minor inconvenience i just discovered: gamepad won't always work in textentry & filesect menus (depends if the button/axis callbacks are implemented for the calling script/menu).

i could make it work independently of the calling context like i did w/ `keyboard`.

tehn:
i'd say this is good as is and if we wanted to revisit the context-sensitivity for textentry/fileselect we can do that further on?

tehn (approved):

Issue #1637, opened 2023-01-09
lattice transport value gets increased in steps of 5


kasperbauer:
@ryleelyman
@tyleretters  

Hey there, i think i just found an issue with lattice 2.0, due to the new sprocket ordering logic.
Heres a little script to demonstrate the problem. The action should be (is?) executed every pulse, but the transport value is increased in steps of 5.

```
lattice = require('lattice')

function init()
    lat = lattice:new()
    sprocket = lat:new_sprocket({
        action = function(t)
            print(t)
        end,
        division = 1 / lat.ppqn / 4, -- do every ppqn
        order = 1
    })
end

function key(n, z)
    if z == 1 then
        if n == 2 then
            lat:start()
        elseif n == 3 then
            lat:stop()
        end
    end
end
```

The output is the following:
```
> 0
5
10
15
20
25
30
35
40
45
50
55
60
65
70
75
80
85
90
95
100
105
...
```

If I'm correct, the transport value should be increased outside the for-loop. PR incoming.

dndrks:
closed via #1638

PR #1638, opened 2023-01-09
fix lattice transport value getting increased in steps of 5

kasperbauer:
Here's the corresponding issue: #1637

Lattice transport values should be increased outside the for-loop. 
In the current state, the transport values are increased in steps of 5, due to the sprocket ordering logic.

### DIFF
--- a/lua/lib/lattice.lua
+++ b/lua/lib/lattice.lua
@@ -118,9 +118,9 @@ function Lattice:pulse()
           flagged = true
         end
       end
-      if flagged then
-         self:order_sprockets()
-      end
+    end
+    if flagged then
+      self:order_sprockets()
     end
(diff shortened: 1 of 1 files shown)


ryleelyman (approved):
Thanks for catching this! I wouldn't mind also moving the `if flagged then` statement also out of the inner loop, just so that it's reached once instead of potentially five times per clock tick.

kasperbauer:
Right, just added a commit with that.

ryleelyman:
Wonderful, thanks for catching my mistake!

ryleelyman (approved):

tehn (approved):

tyleretters:
thank you!

PR #1642, opened 2023-01-10
allow sending midi clock to all devices simultaneously

zjb-s:
Added an option to the "midi clock" param under `clock.lua`, "all", which sends clock to every device whose name isn't "none".

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -248,10 +248,10 @@ function clock.add_params()
       norns.state.clock.link_start_stop_sync = x
     end)
   params:set_save("link_start_stop_sync", false)
-  local clock_table = {"off"}
+  local clock_table = {"off","all"}
   for i = 1,16 do
     local short_name = string.len(midi.vports[i].name) < 12 and midi.vports[i].name or util.acronym(midi.vports[i].name)
-    clock_table[i+1] = "port "..(i)..""..(midi.vports[i].name ~= "none" and (": "..short_name) or "")
+    clock_table[i+2] = "port "..(i)..""..(midi.vports[i].name ~= "none" and (": "..short_name) or "")
   end
   params:add_option("clock_midi_out", "midi out",
       clock_table, norns.state.clock.midi_out)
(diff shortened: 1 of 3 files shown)


tehn:
hello! thanks for taking this on.

i have two concerns:

1. i suspect the next feature request will be selective clocking to the list, ie, clock to a few midi devices but not all. this is not an unreasonable request, since we're pretty far down the menu-config rabbit-hole as is
2. the existing midi clock-out code was "expeditiously" inefficient (with regard to the number of operations that get called per 1/24 clock), and i think this clock-all solution puts it into bad territory: it's a _lot_ of redundant processing per clock tick. it makes me concerned that as we pile in more solutions like this to the core that we're stealing cpu capabilities from scripts themselves.

recommendation, which requires some UI design:

- each midi device should have clock-out enable as a setting
- a function `build_midi_clock_table` which scans plugged in midi devices, if they have clock-out enabled, add their `clock` function to a table. this build function should get called on all midi reconnects and state changes for clock-out
- a single function `midi_clock_out` which traverses the table with the clock functions, calling them sequentially. this is the thing that gets called by the clock tick.

the solution above has zero conditionals. a clock tick would then do the minimum required operations.

counter proposals welcome!

zjb-s:
I noticed the fixme about conditionals and your explanation makes it super-clear, thanks for that. sounds like @dndrks is on this but either of you let me know if you want a hand or anything!

dndrks:
oh! if we're touching MIDI clocking, i'm sure that folks might also want to specify which device is clocking when `source` is set to `midi`, instead of having to open settings for every single box they might have connected.

> some UI design

if it helps to chat through, i think it could be useful to keep things in the clock menu + add some separators to chunk things a bit. maybe remove the single 'midi out' option and just build a 'MIDI' section of the clock params where all the connected device names populate (so nobody needs to remember which device is in which port) via the dynamic show/hide methods.

tehn:
SOURCE i do not think is a problem in its current implementation, since we only want a single source selected.

midi submenu makes sense with toggles.

on that note, crow clock out should probably just be a toggle then as well.

(again trying to anticipate future requests)

dndrks:
>since we only want a single source selected

totally, apologies for the murkiness, i just meant allowing the user to say 'this is the single source i want to listen to'. but yeah, outside of this scope.

dndrks:
added some commits to align with @tehn 's recommendation!

- any currently-ported midi device will populate in the 'midi clock out' section of 'PARAMS > CLOCK'
- each visible entry has a toggle to receive norns clock
- toggles get saved/restored as part of `system.state` during clean reboots

following the suggested approach actually saves us from doing any if's during each clock tick!

zjb-s:
Tested with norns>mmtrs>interface>macbook + norns>syntakt. Works great. Nice work on the UI!

dndrks:
great call @zjb-s !
i also aligned the formatting with `SYSTEM > DEVICES > MIDI`, which allows for up to 20 characters of space before aliasing is helpful

tehn (approved):

PR #1648, opened 2023-01-24
[untested] attempt fix for dropped sysex bytes

catfact:
blind fix attempt for dropped sysex bytes. (we don't actually want to start a new message on the stop byte.)

attn @okyeron


catfact:
oops i typoed

okyeron:
Testing - this appears to fix the dropped bytes!  Sending random sysex now all of it comes through. 

16n Controller test with sines is now properly getting the full 90 byte sysex string.

Looks good to me!  👍

tehn (approved):

PR #1650, opened 2023-02-14
(#1650) Screen Improvements

tlubke:
- Added ssd1322(.c/.h)
 - Added display.lua
 - Added new settings menu: SYSTEM > DISPLAY
 - Added new thread: ssd1322
 - Removed fbdev.c
 - NOTES:
   * The major change in this commit is the move from a kernel-space screen driver via an SPI overlay, to a user-space driver using a thread model. In this change, the screen is treated as a generic linux "spidev" device. This facilitates in some ways a simplification of the screen operations, or at least makes it a bit more clear and brings maintenance into the hands of norns maintainers. The kernel driver previously in use is part of fbtft, which has ceased development. Notably, the use of a framebuffer is no longer needed and so there is no need for an extra cairo surface to be mapped to system memory.
   * Default settings in the DISPLAY menu were chosen based on legacy. For example, the gamma value default is 1.0 which copies the original linear white balance. It would probably be valuable to consider changing these defaults if the userbase manages to settle on generally agreeable settings.
   * Considerable effort was put into trying out smaller cairo formats (A8, A4, G8, C8, ARGB16) but most were fruitless. The 4-bit per color formats do not seem to be any more performant than 8-bit. And the alpha-only and grayscale formats did not work with blending modes. The remaining promising formats (C8, YUY2, YUV12) did not work with the former having a bug in calculating stride, and the two latter being unsupported (yet defined?) in pixman which is the backend of cairo and an ancient library from the 1980s.
   * I do not like the way I have implemented saving/loading DISPLAY settings in lua. There is too much repeated and hardcoded. It sort of follows the practices of other settings, but I think it needs to be improved.


tehn:
amazing! looking forward to testing this out and will review the code.

tlubke:
Been noticing some slight flickering using this a little more. Reverted the 75hz refresh rate to ~120hz and it improved. However there is still some weird subtle strobe effect. It seems to be because of using a dedicated thread vs my branch that just does all the updates in `screen.c`.

tehn:
@tlubke i'm back from crow production land, looking forward to checking this out! anything i should be aware of? comment above suggests everything is good?

has anyone else had a chance to test this beyond my early testing?

tlubke:
@tehn 

> anything i should be aware of?

Just two things:
1. This comment still applies for all of this code to work: https://github.com/monome/norns/discussions/1615#discussioncomment-4180802
2. I gave the thread a higher priority than default since timing can mess with the visual feedback a lot. I haven't seen or heard any issues, but I also didn't push anything too hard or run anything that maxed out the CPU. Do you think there should be other threads with higher priority? Maybe anything involving audio.

> has anyone else had a chance to test this beyond my early testing?

Not since my changes moving to a thread model. But before that there were a handful of users from the discord that did a little testing. I have a few tests scripts I made for my own verifications if you would like to try those out.

tlubke:
Pulling some discussion over from lines.

### recent thoughts from myself

* > [...] I think I respect the responsibility of norns users enough to not enforce a frame limit per-se. Already, it can be very easy to freeze up the system by drawing too often, even with a frame limit in place (I think the cost of transferring the screen data over SPI is minimal compared to the cost of its surface rendering).
* As a matter of personal taste, I think I am also in favor of the above because it makes the behavior of the driver closer to how the code describes it. If I were a first-time-reader I'm not sure if I would expect `ssd1322_init()` to spawn a thread for example.
* I can't remember for certain, but I think `hello.c` would have to be modified if this approach is taken. I think there were 1000s of calls `screen_update()` happening and slowing down the startup animation.

### questions

@tehn 
> i regret that it’s been in staging for almost a year now, as it’d be a substantial update— perhaps a completely new disk image, which is rather disruptive

**What are the capabilities of the update system?** I know disk images are a huge headache, and I don't think these changes would require them at all since the branch can work without any changes to the kernel.

catfact:
@tlubke can you humor me and just briefly describe the threading design here?

is the idea to just move framebuffer updates to the new thread? i agree with discussion elsewhere that the time spent blocking on SPI transactions is probably much less significiant than time actually spent in cairo rendering.

tlubke:
> @tlubke can you humor me and just briefly describe the threading design here?
>
> is the idea to just move framebuffer updates to the new thread? i agree with discussion elsewhere that the time spent > blocking on SPI transactions is probably much less significiant than time actually spent in cairo rendering.

@catfact 
Sure! Yeah, you've pretty much got it down with frame buffer updates going to a new thread. Although, with this PR the actual linux kernel frame buffer and memory map goes away. It's essentially a straight shot from a cairo image surface to the hardware. The main reason for going to a threaded model was because of an expressed interest for a frame rate limit that happened in previous discussions. So, there is essentially a dedicated high-priority thread running at 60 hz that will transfer the surface data over SPI *only if* `screen_update()` has been called. This is a slight improvement from the current kernel driver for the SSD1322 screen which is *always* pushing across that data at a rate of 20hz.

>  what i would actually want to do is move all cairo calls to a separate thread. this is the gist of this even older PR, which i would be delighted if anyone could assist with testing / finishing / landing; if i understand correctly your “low level screen improvements” don’t include that kind of change (which heavily impacts matron structure and some lua APIs.)

To be honest it hadn't really crossed my mind to do this, but I can see the benefits of having a dedicated thread for cairo surface drawing. I'll take a look at https://github.com/monome/norns/pull/1309 and see if I can merge that work easily with mine.

tlubke:
Merged via #1717.

Issue #1657, opened 2023-02-18
keyboard.code() and keyboard.char() clarifications


niksilver:
A question: What are the intended behaviours of [`keyboard.code()`](https://github.com/monome/norns/blob/main/lua/core/keyboard.lua#L92) and [`keyboard.char()`](https://github.com/monome/norns/blob/main/lua/core/keyboard.lua#L94)? I am willing to clarify the documenation and make any code changes (if we decide there should be any), but importantly I'd like to understand the intended behaviour of each.

tl;dr - Maybe `keyboard.code()` should be recommended for non-visible characters only (maybe not); `keyboard.char()` should be recommended for visible characters only; it would be useful to add a code-to-character converter function, but its logic will depend on the intended logic of `keyboard.code()`.

# Observations

Here are some specific observations...

## keyboard.code()

This is intended to capture individual key presses. For example, when the key for letter `A` is pressed then this function will receive the same code regardless of whether the `SHIFT` key is being pressed to capitalise it. But this function also captures whether the key has gone up or down or is held (via the second parameter being 0, 1 or 2 respectively).

## keyboard.char()

This is intended to capture the character generated from key presses, and is only called for visible characters. For example, if the key for letter `A` is pressed then this function will receive the character `A` or `a` depending on whether `SHIFT` is currently pressed. But this function is only called when the key is pushed down or held. We cannot use this function to see when a key is released.

## Both functions

When a key is pushed down, and also when it is held, both the above functions are called.

## Key remapping

Keyboards will get remapped for different countries' keyboard layouts. By looking at `keyboard.process()` we learn that this key mapping happens after `keyboard.code()` is called, and therefore is only relevant for `keyboard.char()`.

This therefore leads to a question: What should we expect for `keyboard.code()` if someone presses the key for letter `A`? Currently, if the user has a US keyboard this function receives `code = 30`, but if they have a French keyboard it receives `code = 16`. This kind of difference also exists for non-alphabetic characters (ie `*`, `#`, `<`, etc).

And the reverse is therefore also true. For example, if `keyboard.code()` receives `code = 30` then it's not at all clear what character has been pressed. But we sort-of know the physical placing of the key on the keyboard!

So in general we must presume that all keys for visible characters may get remapped. I'm going to _assume_ that country-specific layouts do _not_ remap non-visible keys (ie `ENTER`, `LEFTSHIFT`, `F7`, etc).

[...]

PR #1659, opened 2023-02-21
Implement character/key conversion

niksilver:
As discussed in #1657, enable developers to work with printable characters from within `keyboard.code()`, plus supporting documentation.
- Includes new function `keyboard.code_to_char()`. 
- Includes a new function `keyboard.char_to_code()`.
- Replaced some code from `keyboard.process()`, to use `code_to_char()` - which just uses that original code.

I suggest further discussion on this topic takes place here, not in the originating issue.

This PR does not include updates to the `docs/` directory because when I generated the docs it made too many changes (most notably updating every external reference in the docs to the Lua 5.3 manual). I presume the docs will be generated before an official release. However, I have tested locally using `ldoc`.

This PR does not include the suggested `keyboard.char_lift()` as it duplicates functionality (a bit) and I thought it might just be too much.

I have tested this manually using [the script over here](https://github.com/niksilver/norns-kcode-test) which can be installed via the maiden REPL using
```
;install https://github.com/niksilver/norns-kcode-test
```

### DIFF
--- a/lua/core/keyboard.lua
+++ b/lua/core/keyboard.lua
@@ -89,9 +89,24 @@ function keyboard.clear()
   keyboard.char = function() end
 end
 
---- key code callback, script should redefine
+--- key code callback, script should redefine.
+-- use this primarily for responding to physical key presses ('LEFTSHIFT',
+-- 'ENTER', 'F7', etc) not printable characters, as it does not adjust to
+-- the setting of the user's keyboard layout, including national variant.
+-- if you want to detect the character, use @{char} or @{code_to_char}.
+--
+-- @tparam string key    a code representing the physical key pressed.
+-- @tparam number value    0 = released, 1 = pressed, 2 = pressed and held.
+--
 function keyboard.code(key, value) end
---- key character callback, script should redefine
+
+--- key character callback, script should redefine.
+-- will only be called for printable characters such as letters, numbers
+-- and punctuation - and space. will take account of the setting of the user's keyboard
+-- layout, including national variant.
+--
+-- @tparam string ch    the printable character intended to be generated.
+--
 function keyboard.char(ch) end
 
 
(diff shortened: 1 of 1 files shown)


p3r7:
`code_to_char` implem is looking mighty fine.

on the other hand i'm not convinced by the usefulness of `char_to_code` nor its robustness.

don't get me wrong. it's not a issue of code quality but the problem is that `code_to_char` is **certainly not bijective** (not a 1 to 1 code / char mapping)!
(also `ikm` should be recomputed on change of `keyboard.selected_map`, inside of `keyboard.load_enabled()`)

niksilver:
On your points, well spotted on these two:
* Yes, `char_to_code` is certainly not bijective. I meant to add something to that effect in the documentation, but forgot.
* Yes, `ikm` should be recomputed. That error is due to laziness on my part. 😞 

As to the usefulness of `char_to_code`... I don't know. Mainly it seemed it might a reasonable countpart to `code_to_char`. It might be used in the situation where the developer wants to test for a character in `keyboard.code()` because that's where the rest of their logic is. Eg:

```lua
QUIT_BUTTON = keyboard.char_to_code('Q')

function keyboard.code(code, value)
  if code == QUIT_BUTTON and value > 0 then
    do_quit()
  end
  -- etc
end
```

That seems a bit nicer than this:
```lua
function keyboard.code(code, value)
  local is_q = keyboard.code_to_char(code) == 'Q' or keyboard.code_to_char(code) == 'q')
  if is_q and value > 0 then
    do_quit()
  end
  -- etc
end
```

Meanwhile this is simple, but only if the developer is happy to use `keyboard.char()` - which might not be the case if the rest of their logic is in `keyboard.code()`:
```lua
function keyboard.char(ch)
  if ch == 'Q' or ch == 'q' then
    do_quit()
  end
  -- etc
end
```

So... I think it has a bit of utility. But I don't feel strongly about it.

p3r7:
the issue w/ `char_to_code` not being bijective is not a documentation one.

it's that `code_to_char` is not guaranteed to work if several key combination in a layout can result in the same char. in that case a simple map for reverse lookup may not cut it.

likewise, `char_to_code` would have to take into account current modifiers (like `char_to_code`). untested but i predict that w/ current implem if i try to map the `QUIT_BUTTON` to say the `0` on my AZERTY keyboard, it'd get shadowed by either `@` or `à` as they all share code but w/ different modifiers.

the `keyboard.code_to_char` approach may be more verbose but is more correct.

niksilver:
I've now removed code relating to `char_to_code()` and left only `code_to_char()` with the updated documentation. I've also updated my [manual testing script](https://github.com/niksilver/norns-kcode-test).

So this is ready for further review.

tehn:
apologies for the delay. any further thoughts on this one? @niksilver @p3r7

p3r7:
everything looks good to me.

niksilver:
Yes, I'm happy with what we have here.

tehn (approved):

PR #1662, opened 2023-03-13
Sequins v2 & Timeline from crow, on norns

trentgill:
This updates norns to the new `sequins` v2 library, and adds the new `timeline` library.

Both files are identical to their crow counterparts.

If there's any concern about either of these, I'm happy to split them into separate PRs.

### DIFF
--- a/lua/lib/sequins.lua
+++ b/lua/lib/sequins.lua
@@ -1,164 +1,252 @@
 --- sequins
--- nestable tables with sequencing behaviours & control flow
--- TODO i think ASL can be defined in terms of a sequins...
---
--- for the norns port, @tyleretters copy-pasta'd with no changes from:
--- https://github.com/monome/crow/blob/34ce1e455f01fdef65a0d37aa97163b4cd14a115/lua/sequins.lua
 
 local S = {}
 
-function S.new(t)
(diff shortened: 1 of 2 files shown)


tehn (approved):
look great. haven't tested on device, but given these are straight copies i feel good about it.

dndrks:
omfriggingod i love having `timeline` on norns!!

@trentgill , i ran into something while building out norns docs that i wanted to confirm, whenever you have a sec?
while using `sequins` with `timeline` on crow, the following use returns values aligned with what [the docs suggest](https://monome.org/docs/crow/timeline/#sequins-enabled)):
```lua
timeline.loop{ 1, {print, sequins{0,2,4,6,8,10}} }
```
returns:
```bash
0
2
4
6
8
10
```

but on norns, it returns just a print loop of:
```bash
s[1]{0,2,4,6,8,10}
```

my first troubleshooting impulse was to change to:
```lua
timeline.loop{ 1, {print, sequins{0,2,4,6,8,10}()} }
```
but this just endlessly returns `0`, since the sequins has simply fired-off the first value and saved that in the loop step.

is there a way for `timeline` to have closer-to-total parity between crow + norns? or is there a workaround we should highlight in the norns docs? tytyty!

PR #1667, opened 2023-03-23
Adding rx.lua(ReactiveX, FRP) to lua/lib

hankyates:
Hi everyone,

I wanted to see how everyone felt about adding this library that's an implementation of the Obeservable software pattern called rx. This is sometimes called [ReactiveX](https://reactivex.io/) or [Functional Reactive Programming](https://en.wikipedia.org/wiki/Functional_reactive_programming).

If you are curious what a norns use case would look like see https://github.com/hankyates/norns-convolution-reverb/blob/master/convolution.lua#L52

### DIFF
--- a/lua/lib/rx.lua
+++ b/lua/lib/rx.lua
@@ -1,0 +1,2318 @@
+-- RxLua v0.0.3
+-- https://github.com/bjornbytes/rxlua
+-- MIT License
+
+local util = {}
+
+util.pack = table.pack or function(...) return { n = select('#', ...), ... } end
+util.unpack = table.unpack or unpack
+util.eq = function(x, y) return x == y end
+util.noop = function() end
+util.identity = function(x) return x end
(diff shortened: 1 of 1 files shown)


ngwese:
others can weigh in but historically there has been an effort to keep the core repository as minimal as possible and encourage the practice of bundling third party libraries with the scripts that consume them, pulling them in via [`include(...)`](https://monome.org/docs/norns/reference/#libraries). scripts can use the `include` function to load code in sibling script directories and a few community maintained "library" scripts. one such example is [soot](https://github.com/northern-information/soot)

tehn:
i am in agreement--- since such libraries can be distributed with scripts it keeps the norns core focused. this topic has been discussed at length in the past with the core developers.

if there's strong desire to reopen this inquiry let's start a [thread](https://github.com/monome/norns/discussions)

PR #1668, opened 2023-03-23
add system/settings menu, move items, add battery_warning

tehn:
we'd discussed this in the past, but SYSTEM now is getting full enough to merit a submenu of lesser items. i've moved RESET and PASSWORD into there, alongside this new one, BATTERY WARNING.

bonus is RESET now makes more set in the context of SETTINGS, less confusion with RESTART (because RESET commonly means the same thing, ie "i reset my NES")

if there are additional needs to SETTINGS i guess the flood gate is open

but also if there are strong preferences against this i'll close this


lua/core/state.lua
```diff
@@ -111,6 +113,7 @@ state.save_state = function()
   io.write("norns.state.shortname = '" .. state.shortname .. "'\n")
   io.write("norns.state.path = '" .. state.path .. "'\n")
   io.write("norns.state.data = '" .. state.data .. "'\n")
+  io.write("norns.state.battery_warning = '" .. state.battery_warning .. "'\n")
```
dndrks, line 116: should save as number instead of a string, otherwise the `==1` checks fail

lua/core/menu/settings.lua
```diff
@@ -0,0 +1,94 @@
...
+
+local m = {
+  pos = 1,
+  list = {"RESET", "PASSWORD >", "BATTERY WARNING"},
+  pages = {"RESET", "PASSWORD"}
+}
+
+m.key = function(n,z)
+  if n==2 and z==1 then
+    _menu.set_page("HOME")
```
dndrks, line 11: for consistency with the other `SYSTEM` sub-pages, should this be `_menu.set_page("SYSTEM")` ? otherwise, K2 jumps back two menu layers, instead of one.

tehn:
oh dang look at this thorough code review ty!! will fix all

PR #1669, opened 2023-03-23
improved png export

### LINKED ISSUE #1548: better screenshots
currently, people making screenshots typically post-process them in imagemagick.

i fielded a question about adding imagemagick to the base image as follows:

> 
>   IMHO just making screenshots is not sufficient call for installing Magick, which is pretty big (iirc?) and has lots of dependenc ies for dealing with all possible graphics formats. I would either:
> 
> - export a trivial format like PGM direct from Lua
> - if PNG format is preferred, add a dedicated function in C that uses `libpng` (and probably upsamples)
> 
> The required form of upsampling is trivial (no interp), for example to 3x upscale:
> - allocate temp line buffer of 128x3 levels
> - for each line in source buffer:
>   - for each column, copy the source level to the next 3 positions in the line buffer
>   - write the whole tmp line buffer to output file 3 times
> 
> To put it kind of bluntly, I don't like burning 2.5% of every CM3 users system shortage to save us writing ~10 lines of code.

so in other words, maybe it would be nice to update the `screen_export_png` function here:
https://github.com/monome/norns/blob/main/matron/src/hardware/screen.c#L399

to perform upsampling and inversion (on a temporary surface.)

tehn:
fixes #1548

- scales pixels by 4, adds 64 pixel black border
- adds `screen.export_png(file)` to `dust/data/(script)/(file).png`

concerns
- gamma is off. previously using imagemagick `-gamma 1.25`

### DIFF
--- a/matron/src/hardware/screen.c
+++ b/matron/src/hardware/screen.c
@@ -396,7 +396,38 @@ double *screen_text_extents(const char *s) {
 
 extern void screen_export_png(const char *s) {
     CHECK_CR
-    cairo_surface_write_to_png(surface, s);
+    static cairo_surface_t *png;
+    static cairo_t *temp; // for bg fill
+    // width = 640 (128*4 pixels with 64 pixel black border)
+    // hieght = 384 (64*4 pixels plus border) 
+    png = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 640, 384);
+    temp = cairo_create(png);
+    // fill
+    cairo_set_source_rgb(temp, 0, 0, 0);
+    cairo_rectangle(temp, 0, 0, 640, 384);
+    cairo_fill(temp);
+    // copy big pixles
+    uint32_t *src = (uint32_t *)cairo_image_surface_get_data(surface);
+    uint32_t *dst = (uint32_t *)cairo_image_surface_get_data(png);
+    if (!src || !dst) return;
+    dst += 64 + 640*64;
+    for(int y=0;y<64;y++) {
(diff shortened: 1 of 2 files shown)


matron/src/hardware/screen.c
```diff
@@ -396,7 +396,38 @@ double *screen_text_extents(const char *s) {
 
 extern void screen_export_png(const char *s) {
     CHECK_CR
-    cairo_surface_write_to_png(surface, s);
+    static cairo_surface_t *png;
+    static cairo_t *temp; // for bg fill
+    // width = 640 (128*4 pixels with 64 pixel black border)
+    // hieght = 384 (64*4 pixels plus border) 
+    png = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 640, 384);
```
tlubke, line 403: I know I wasn't requested to review this, but I was poking around open PRs anyway. You may be able to use CAIRO_FORMAT_RGB24 and then you wouldn't have to worry about the alpha channel (pixels still fit into a uint32). That could probably allow you to lose the fill area of the code, and the bitwise OR with 0xFF for alpha.

Have you experimented with `cairo_scale()`? You might be able to copy the image data from `surface`, use `create_image_surface_for_data()` with the new dimensions, then apply translate and scale transformations.

tehn: thanks for taking a look! your suggestions make a ton of sense.

honestly all of my experience with graphics is on raw framebuffers... hence the low-level implementation. i'll see if i can find a minute to try out some of these ideas.

PR #1670, opened 2023-04-05
norns.expand_filesystem()

dndrks:
`norns.` helper to avoid [all of these steps](https://monome.org/docs/norns/help/#expanding-storage) after fresh installing.
if 4gb cm3 is installed, expand does not occur and prints a message to matron.

### DIFF
--- a/lua/core/norns.lua
+++ b/lua/core/norns.lua
@@ -260,3 +260,13 @@ end
 norns.rerun = function()
   norns.script.load(norns.state.script)
 end
+
+-- expand the filesystem after a fresh installation
+norns.expand_filesystem = function()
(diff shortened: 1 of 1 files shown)


tehn (changes requested):
i don't believe that anything will break if it's a 4GB model--- it just... won't do anything useful? so it seems we can skip this.

furthermore, 32GB CM3+ models need to use this (not just shield) so i'd skip the shield check

lastly, it'd be better for people to manually restart, so that matron saves settings/etc and has a "clean shutdown"

so... i'd just reduce this function to the one-liner w/o the shutdown?

(please let me know if i'm missing an important point in all of this)

dndrks:
@tehn , makes sense! fixed!

tehn (approved):

PR #1671, opened 2023-04-08
allow user to add gamepad profiles without tempering w/ core

p3r7:
so rn only 2 gamepad models are suported natively (profile bundled w/ norns core).

i made a [small norns script](https://github.com/p3r7/gamepad_wiz) that allows one to generate profile from a pluggued in gamepad and output it under `~/we/dust/data/gamepads/`.

this PR enables sourcing those user-level profiles dynamically.

2 qs:
- is this a sane approach?
- what about the name / location of the user's gamepad profile (`~/we/dust/data/gamepads/`)?


tehn (commented):
i do think this approach is fine--- though it'd also be good if we could get some more profiles into the core.

PR #1677, opened 2023-05-10
small fixes: softcut API

dndrks:
softcut's `voice_sync` command has its arguments reversed in the API docs.

[current documentation](https://monome.org/docs/norns/api/modules/softcut.html#voice_sync): `voice_sync (src, dst, offset)`
[source](https://github.com/monome/softcut-lib/blob/57fb9877e13b991aa642246ced70e7fff00c5b1b/softcut-lib/include/softcut/Softcut.h#L176-L178): `syncVoice(int follow, int lead, float offset)`

(@tehn , this'll need a docs re-gen!)

### DIFF
--- a/lua/core/softcut.lua
+++ b/lua/core/softcut.lua
@@ -85,8 +85,8 @@ SC.loop = function(voice,state) _norns.cut_param("loop_flag",voice,state) end
 
 --- set fade time.
 -- @tparam int voice : voice index
--- @tparam number pos : crossfade time in seconds
-SC.fade_time = function(voice,pos) _norns.cut_param("fade_time",voice,pos) end
+-- @tparam number fade_time : crossfade time in seconds
+SC.fade_time = function(voice,fade_time) _norns.cut_param("fade_time",voice,fade_time) end
 
 --- set record level.
 -- this sets the realtime-modulated record level,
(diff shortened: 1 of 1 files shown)


dndrks:
saw another typo, renamed PR!

catfact:
That's very strange, I could have sworn we already fixes these ages ago... Let's double check the order of arguments for the sync command when you actually dig sown to softcut OSC handler

Also if we're gonna update the comment for fade time, maybe correct the parameter name as well

dndrks:
@catfact , i thought so too! but both on my device and at the [static API](https://monome.org/docs/norns/api/modules/softcut.html#voice_sync), `voice_sync` is listed as `voice_sync (src, dst, offset)`.

re: OSC handler, it seems like it all expects `dst,src,offset`. [this](https://github.com/monome/softcut-lib/blob/57fb9877e13b991aa642246ced70e7fff00c5b1b/clients/softcut_jack_osc/src/OscInterface.cpp#L306-L309) calls [this](https://github.com/monome/softcut-lib/blob/57fb9877e13b991aa642246ced70e7fff00c5b1b/clients/softcut_jack_osc/src/SoftcutClient.cpp#LL188C9-L190C19), which [lands here](https://github.com/monome/softcut-lib/blob/57fb9877e13b991aa642246ced70e7fff00c5b1b/softcut-lib/include/softcut/Softcut.h#L176-L178), yeah?

```h
void syncVoice(int follow, int lead, float offset) {
            scv[follow].cutToPos(scv[lead].getActivePosition() + offset);
        }
```

catfact:
huh, this never got merged? i shall merge it?

PR #1680, opened 2023-05-24
hotfix: midi clock double-up

### LINKED ISSUE #1679: 'send_midi_clock' table can hold duplicates, causing double-clocking (clock.lua)
<details><summary>reported via support</summary>

Hi folks,

I’m experiencing an issue with a phantom (not else what to call this LOL) system clock that keeps running after I disable all ports in parameters:clock. It also seems to get doubled resulting in weird BPM on the receiving end.

Here’s how I’m able to reproduce:

1. System restart with no scripts running
2. Set parameters:clock to Internal, BPM 120, sending out on one port. >> Confirm correct BPM is being detected in my DAW (Ableton, externally synced as follower)
3. Start Awake script. >> BPM is now detected as something around 128 in DAW
4. Disable sending of clock on all ports in parameters:clock. >> Clock continues to run and is picked up as 120 BPM in DAW. Changing the BPM in parameters results in the DAW picking up the new tempo despite all ports being disabled in the Norns UI.

The only workaround I have found:
1. Disable all clock ports
2. System restart
3. Launch script
4. Enable sending of clock on the port in parameters.

I missed some updates prior to stable-230509 update so I’m not sure if it was introduced then. Reproducible with multiple MIDI interfaces.
 </details>

tracked it down to the lifecycle of the [`send_midi_clock`](https://github.com/monome/norns/blob/main/lua/core/clock.lua#L203) table, which is freshly initialized at every script load, but is double-tapped by both the parameter rebuild _and_ the `params:bang` in [`params:default`](https://github.com/monome/norns/blob/36b9bf78dd083ca309953ca15e0a107126c3101d/lua/core/paramset.lua#L554-L558), which many scripts utilize.

fix incoming!

dndrks:
fixes #1679 by adding a check to see if `send_midi_clock` already contains the initiated device, which avoids the double-registration which gets triggered by `params:default()`

### DIFF
--- a/lua/core/clock.lua
+++ b/lua/core/clock.lua
@@ -257,7 +257,9 @@ function clock.add_params()
     params:set_action("clock_midi_out_"..i,
       function(x)
         if x == 1 then
-          table.insert(send_midi_clock,i)
+          if not tab.contains(send_midi_clock,i) then
+            table.insert(send_midi_clock,i)
+          end
         else
           if tab.contains(send_midi_clock,i) then
             table.remove(send_midi_clock,tab.key(send_midi_clock, i))
@@ -314,7 +316,6 @@ function clock.add_params()
   end)
 
   -- executes midi out (needs a subtick)
-  -- FIXME: lots of if's every tick blah
   clock.run(function()
     while true do
       clock.sync(1/24)


artfwo (approved):
LGTM!

PR #1704, opened 2023-09-26
make variable y in util.wrap local

sixolet:
Should fix https://llllllll.co/t/norns-study-zero-first-light/41017/70


dndrks (approved):
ah, thank you!!

PR #1705, opened 2023-09-26
string.char to utf8.char

### LINKED ISSUE #1702: replace / assess `string.char()` usage
rephrasing https://github.com/monome/norns/issues/1701 for more direct action, thanks @256k for bringing the problem to light.

in lua 5.3 at any rate, `string.char(x, ... )` is i think technically undefined for `x > 127`; in practice it can produce a single-element non-ascii byte array which can lead to various problems for string rendering and transmission.

i think either `string.char()` should always have its arguments clamped to [0, 127], or should be replaced by `utf8.char()` which will always produce a legal C string at least.

a quick search reveals usages of `string.char()` in `textentry` and `crow` lua modules.

tehn:
fixes #1702

### DIFF
--- a/lua/core/crow.lua
+++ b/lua/core/crow.lua
@@ -110,8 +110,8 @@ end
 function norns.crow.register_event(fn)
   local n = tab.count(norns.crow.events) + 1
   local c = ""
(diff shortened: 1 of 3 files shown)


tehn:
mixed commit, apologies

PR #1706, opened 2023-09-27
Norns converged update

catfact:
this is just a merge from `main` into `norns-converged` to bring it up to speed with past ~1yr of changes.

i haven't tested this at all yet! (will download to the rpi presently...) but i didn't see any serious conflicts as most changes are in lua.

(... though, now that i write this i am not sure if the merge was really smart enough to migrate all `.c` changes to their `.cc` counterparts (it does seem to have managed a few of those at least.))

also just pinging interest on restarting the effort to fix and land converged.


tehn:
i'll try to get this compiling tomorrow

catfact:
for what it's worth, i am able to build and run, presenrtly on rpi4+shield.

my steps on the `norns-converged` branch

- 1. install nng requirement
`sudo apt install llibnng-dev`

- 2. clean and build
```
./waf configure --release
./waf clean
./waf build --release
```

- stop and start process/services
```
./stop.sh
systemctl start norns-jack.target
./build/norns/norns
```

and i usually run `sclang` in a separate sesssion / screen.

i'll move on to checking out actual functionality on my next sit-down.

tehn:
good to hear. just dug up the notes on the systemd services and updated the branch:

https://github.com/monome/norns-image/tree/norns-converged

(this is not necessary for development, though it's helpful to make sure things boot up right)

i'll get a dev system together now

catfact:
ok then i guess let's merge this so they stay in sync.

(i wonder if we should rebase the big pending screen change on top of converged...)

tehn:
yes i think i'd highly prefer to have converged and the screen update coincide and require a full disk update. it's a pain for everyone but offers a good opportunity for cleaning up everything.

tehn:
@catfact attempting to decipher the `v3` branches--- were these WIP prior to `converged`?

catfact:
yes. `v3` is same changes but with a longer series of commits going in different directions and sometimes with huge diffs that are hard to merge/rebase. the other one was a temp staging branch and can definitely be deleted. it's conceivable that the `v3` history could be helpful to look at but i think not likely.

catfact:
@tehn @ngwese maybe worth knowing that after the merge things seem to basically work.

i can browse and select scripts, even with hundreds installed.

sclang and softcut still getting controlled by OSC and making sound correctly.

the only oddities i'm seeing are:

- right at launch, this single  report from sidecar. i guess it succeeds after this.
```
sidecar: establishing connection failed: Connection refused
attempts left: 4
```

- at one point, i got a number of these, but then they went away. (probably user error or a debug build or something.)

```
liblo error: 26b6; Invalid message received; 
```

ngwese:
the initial sidecar connection failure is likely related to the main process trying to connect to the sidecar before the sidecar is fully running. adding in a small delay or a message that says something less alarming followed by a positive confirmation when a connection is established might avoid confusion.

iirc the two warts i remember off the top of my head with the converged branch are:
- nng starts a ton of communication threads and this was not configurable at the time, ideally we would get rid of that overhead
- the logic to connect/restore the jack wiring (in lua instead of system calls) results in the jack connections being reset at a time when the sc/crone boot tone is playing which causes it to have a momentary dropout

ngwese:
maybe some good news. it looks like the nng thread count is configurable but only at compile time. i think we should reduce the thread count so this means building our own copy of nng as part of the image build (or norns stack itself). 

https://github.com/nanomsg/nng/issues/1572#issuecomment-1718939065

PR #1707, opened 2023-09-27
double the temp buffer size for systemcmd

catfact:
this simply doubles the size of the temp buffer used for shell command results, from 64kB to 128kB.

it's painful to look at this! it would be nice to make something more incremental.

e.g., it would be nice to make better use of POSIX dirent iterators (? or whatever) to make dedicated   C implementations for tasks like listing all files of a given type in a directory.

if that's too much, maybe we can just change this to use `realloc` incrementally at least. i'll update here if i get a chance, but the change as is should at least support current use cases and i don't think the extra temp allocation will break anything.

### DIFF
--- a/matron/src/system_cmd.c
+++ b/matron/src/system_cmd.c
@@ -15,7 +15,7 @@
 #include "events.h"
 
 #if 1 // nice big buffers (allocated in worker thread)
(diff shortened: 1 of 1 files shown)


tehn:
i think this short term fix is fine. i'll PR the improved suggestion for dir listing

tehn (approved):

PR #1710, opened 2023-09-27
Make dev_monitor messages more consistent and informative

ambv:
This is a tiny change but since I was investigating monome/norns-image#101, I found those changes helpful. Should be uncontroversial to merge.

I mostly cared about including name and node information but once I was on it, I also made the other messages more consistent with each other.


catfact (approved):
lgtm!

PR #1712, opened 2023-10-07
Trim away un-needed select menu content

beels:
* Do not scan through .git directories while looking for scripts (optimization of `find` call only).

* Filter out unwanted lua scripts before returning from shell.

This PR is a refinement of the tweaks to the `find` call that have already been merged (i.e., use of `-printf "%P\n"`).

Based on anecdotal tests on my local install, moving the filter for "crow", "lib" and so on from lua to `grep` removes 90% of the data transferred from the system call to crone.  This removes all lag on entering the select menu.  (Tested with an install base of 87 scripts having 555 unwanted internal lua files.)

### DIFF
--- a/lua/core/menu/select.lua
+++ b/lua/core/menu/select.lua
@@ -28,11 +28,7 @@ local function sort_select_tree(results)
 
   local t = {}
   for filename in results:gmatch("[^\r\n]+") do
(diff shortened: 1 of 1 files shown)


tehn (approved):
fantastic!!

PR #1713, opened 2023-10-11
feat: add `lib/reflection.lua`

ryleelyman:
here’s a clocked pattern recorder @dndrks and I have been kicking around for about a year now. some discussion of it exists over at #1609 . i’m not sure the discussion has fully resolved, so i’ll mark this as a draft, but if there’s interest, it would be nice to see it land soon :)

### DIFF
--- a/lua/lib/reflection.lua
+++ b/lua/lib/reflection.lua
@@ -1,0 +1,237 @@
+--- clocked pattern recorder library
+-- @module lib.reflection
+
+local reflection = {}
+reflection.__index = reflection
+
+--- constructor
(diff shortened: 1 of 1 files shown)


ryleelyman:
thanks very much to @sonocircuit for some thoughtful changes and additions!

one thing I took the liberty of doing is having `begin_playback` just call into `clock.sync` directly rather than launching a new clock. since the library itself only calls `begin_playback` within a clock, this seems innocent, but please let me know if you think things should change @dndrks @sonocircuit

tehn (approved):

PR #1715, opened 2023-10-14
changed offset behaviour + bugfixes

sonocircuit:
following up from this https://github.com/monome/norns/discussions/1687 discussion, here's a PR that:
- changes the behaviour of the `offset` parameter and sets its `min` / `max` values according to the `baseline` value.
- adresses a bug where the lfo would not run smoothly when set to free and norns clocked via midi. the clock jitter would lead to constantly changing `clock.get_beat_sec()` values, which is required for the phase calculation and hence led to erratic changes.
- added the missing `lfo_phase_` param to `lfo_params_visibility()`.

let me know if there are any changes required.

best

s


dndrks (approved):
amazing, thank you so much for these beautiful and playable adjustments, @sonocircuit !!

PR #1716, opened 2023-10-15
Improve hooks available to mods

sixolet:
This PR does two things that will be quite useful to matrix/nb/fx and that ecosystem of mods.

* Adds a `script_post_init` hook. 
* Sorts the hooks alphabetically before executing them. This allows mods that add parameters to add them in a consistent order, and mods that need all parameters available before executing to install a hook starting with `~` to ensure they're at the end.


tehn (approved):

ngwese:
seeing this late, two quick comments:

- would suggest doing the sorting on register and deregister instead of repeatedly doing it each time a hook is called
- we have documentation; i think in the example mod code, which states that the callback order is not defined so the documentation should be updated to reflect that.

sixolet:
Are you suggesting doing this for efficiency? It would be more complex to
code, as we’d need to store both the ordering and the map, and update it on
every insert and removal. I judged that there are tens of these hooks at
most, and sorting them adds negligible time to script startup

PR #1718, opened 2023-10-21
add cpu_time

catfact:
adds a very simple module for measuring CPU time deltas accurately from lua. this is useful for benchmarking.

e.g.:

```
local perform_bench = function(nruns, nsteps, fn_step, fn_init)
    local times_ms = {}
    for i=1,nruns do
        fn_init()
        _norns.cpu_time_start_timer()
        for j=1,nsteps do
            fn_step()
        end
        local delta_ns = _norns.cpu_time_get_delta()
        table.insert(times_ms, delta_ns / 1e6)
    end
    return times_ms
end


```

### DIFF
--- a/matron/src/hardware/cpu_time.c
+++ b/matron/src/hardware/cpu_time.c
@@ -1,0 +1,29 @@
+#include "cpu_time.h"
+
+#include <time.h>
+
+struct timespec t0;
+
+static struct timespec diff(const struct timespec *start, const struct timespec *end)
+{
+    struct timespec temp;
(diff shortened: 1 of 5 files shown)


tehn (approved):

catfact:
@tehn I was thinking, maybe it makes sense to add another option using CLOCK_MONOTONIC

(So, better for wall time estimates. Lots of applications for musical logic as well as little things like refining estimates of softcut position change since last poll.)

Does that seem valuable? Id probably rename the source

tehn:
yeah that'd absolutely be useful, agreed makes it a little broader than just testing/optimizing

PR #1719, opened 2023-10-23
add wall-time delta measurement

catfact:
haven't really tested this but added a `wall_time` complement to the `cpu_time` measurement.


tehn:
thank you!

i can get this tested and into the new beta (which i just posted)

tehn (approved):

PR #1720, opened 2023-10-25
screen improvements (combined)

tehn:
continued from https://github.com/monome/norns/pull/1717


tehn:
reminder to compile with `--enable-display-timer-thread`

retesting here now

dndrks:
raaaad!

just to update [the previous list](https://github.com/monome/norns/pull/1717#issuecomment-1779848346):

- [x] ~`text_extents` errors~: fixed with @catfact 's Lua changes!
- [x] ~flickering~: popping everything off timers and into `refresh()` looks great!!! any flickering with `flora` is actually the results of additional `screen.clear()` calls (eg. [on encoder turns](https://github.com/jaseknighter/flora/blob/9500494c815c2593e746c6a359fee79039f4eb42/lib/encoders_and_keys.lua#L31))
- [ ] aliasing and level differences: still at play

matron/src/hardware/screen/ssd1322.c
```diff
@@ -0,0 +1,393 @@
...
+    }
+    else{
+        // If the surface has only been drawn to, we can guarantee that RGB are
+        // all equal values representing a grayscale value. So, we can take any
+        // of those channels arbitrarily.
+        for( uint32_t i = 0; i < tx_len; i += 16 ){
+            const uint8x16x4_t RGB = vld4q_u8((uint8_t *) (data + i));
+            vst1q_u8(spidev_buffer + i, RGB.val[0]);
+        }
+    }
```
tlubke, line 287: If the aliasing issue is driver-related, I would expect it is somewhere in here.

catfact:
i see the aliasing-type artefacts too. like an interference pattern of black dots. 

i don't have a side-by-side setup at the moment so not the easiest to so extensive comparisons. but i noticed a few things just starting with the simplest sort of gradient test:

```
init = function() redraw() end

local draw_rows = function()
    for z=0,15 do
        screen.level(z)
        screen.rect(0, (z-1)*4, 128*2, 10)
        screen.fill()
    end
end

redraw = function()
    screen.clear()
    screen.aa(0)
    draw_rows()
    screen.update()
end
```

i also tried something with `poke` but i am not sure i am using it right:
```

local draw_poke = function()
    local data = { }
    for z=0,15 do 
        for j=1,512 do
            table.insert(data, z)
        end
    end
    screen.poke(1, 1, 128, 64, table.concat(data))
end
```

some messy observations:
- i don't think i see any of the interference stuff on the gradient. i think i only see it in line drawing and some fonts.
- in the driver code, i noticed that `should_translate_color` always seems to be true. i'm not sure why this would happen.
- if i manually toggle between the branches on `should_translate_color`, i don't see any difference in either the gradient or the aliasing.
- oddly enough, i _do_ see a difference if i use the `draw_poke` above and modify it with a positive x,y offset! weird checkerboard stuff shows up with the `should_translate_color==false` branch. hmm.

tlubke:
> in the driver code, i noticed that should_translate_color always seems to be true. i'm not sure why this would happen.

Ooh, nice catch -- I see what looks like a mistake in `screen.c`. Line-475, in `screen_clear()`, `surface_may_have_color` should be set to false on clear. Looks like it was a commit from me a while ago that made that happen.

> if i manually toggle between the branches on should_translate_color, i don't see any difference in either the gradient or the aliasing.
ditto if i use the luminance-scaling branch, but arbitrarily change the luminance values (for example, making them 80/80/80.)

To clarify, is _not seeing a difference_ here a good thing or bad? Just trying to figure out if you mean you're still seeing weird aliasing in those cases.

Doing some testing now, I'll see what I can find.

catfact:
> To clarify, is not seeing a difference here a good thing or bad? Just trying to figure out if you mean you're still seeing weird aliasing in those cases.

i'm seeing approximately the same results in both cases; forcing `surface_may_have_color=false` didn't improve the aliasing as far as i could tell. (the only difference was in the edge case where `poke` had out of bounds arguments. which is... a little alarming and something i'll try and look more closely at later.)

tlubke:
I think we will have to do some additional bit manipulation. In the example [script](https://github.com/monome/norns/pull/1717#issuecomment-1779848346), doing successive calls of `screen.translate()` with some fractional numbers it's possible to get the lines and circles to have black pixels breaking the shapes up. The text always stays the same though.

tlubke:
Side note, `screen.peek()` seems to be crashing matron.

catfact:
peek screen event wasn't getting handled, so weaver was blocking forever. works now. also fixed bug in peek results buffer.

tlubke:
@catfact Thanks for those fixes, should help me debug a little easier tonight after work. I didn't find a fix last night, but potentially a few leads and some paths not to go down. Here are my notable observations:

* It **does not** seem to be related to SPI transfer speed or buffer size.
* ANDing each pixel with `0xF0` or shifting each pixel right by 4 seems to produce the same effect: **Aliasing looks normal, and there are 15 distinct levels, but brightness is reduced drastically**.
* Using direct copy of one color channel per pixel in `data` to `spidev_buffer` has the same result as using the ARM intrinsics.
* Putting two pixels into one byte of the SPI transfer reconfirmed does not work (the data sheet is confusing about RAM writing).

dndrks:
raaaad!! aliasing looking _much_ much better, dang @tlubke !! thank you for the ultra-lightspeed fix.

two things popped up during testing:

### `screen.update()`

with the latest changes, `screen.update()` is not required to draw to the screen upon running a script, which feels like it could cause small trouble. eg i can totally leave it out of this sketch and it'll draw on the `screen` branch, but (as expected) does _not_ draw on `main`:

```lua
-- test script:
function redraw()
  for i = 1,15 do
    screen.level(i)
    screen.rect((i-1)*8,0,8,64)
    screen.fill()
  end
end
```

### low-end gamma tuning?
the lower `screen.level`s (maybe up until 6?) don't have nearly as much distinction on `screen` as they do on `main`. really really hard to capture with a camera, but here's the same sketch running on two shields. left is `screen` branch, right is `main`:

image

-----

lmk if i can help with any other testing! so so stoked on these changes!!!

tlubke:
> ### `screen.update()`
> with the latest changes, `screen.update()` is not required to draw to the screen upon running a script, which feels like it could cause small trouble. eg i can totally leave it out of this sketch and it'll draw on the `screen` branch, but (as expected) does _not_ draw on `main`:
> 
> ```lua
> -- test script:
> function redraw()
>   for i = 1,15 do
>     screen.level(i)
>     screen.rect((i-1)*8,0,8,64)
>     screen.fill()
>   end
> end
> ```
> 
> ### low-end gamma tuning?
> the lower `screen.level`s (maybe up until 6?) don't have nearly as much distinction on `screen` as they do on `main`. really really hard to capture with a camera, but here's the same sketch running on two shields. left is `screen` branch, right is `main`:

@dndrks Thanks for the continued testing! I probably never would have noticed the `screen.update()`. I'll look into tonight, but I'm wondering if there isn't some sort of race condition. Does it redraw when re-initiating a script directly rather than going through the menu system?

Regarding gamma tuning, I'll double check that new defaults in the `screen` branch match what they're set to in the original driver.

> lmk if i can help with any other testing! so so stoked on these changes!!!

If you could play around with the `SYSTEM > SETTINGS > DISPLAY` page and let me know what you think and if there might be better defaults to use. If you happen to match up the gamma between `screen` and `main` that way, let me know what settings you landed on.

tlubke:
One _quality_ concern I've noticed is that the boot screen consistently has 1 or 2 flickers during its run. Might need to take another look at `hello.c`.

tehn:
i'm away from hardware right now, but looking at `hello.c` i see `tick_us = 5000` which is i think attempting to draw at 200fps (5ms) ??? i could see this causing a glitch ;)

`timeout_ticks` needs to be adjusted if this interval is changed

dndrks:
@tlubke :)
>Does it redraw when re-initiating a script directly rather than going through the menu system?

it draws when i initiate the script via maiden's 'play' button, and if i add a `screen.clear()` (still no `screen.update()`, though), it'll draw when i go through the menu system.

>If you could play around with the SYSTEM > SETTINGS > DISPLAY page and let me know what you think and if there might be better defaults to use. If you happen to match up the gamma between screen and main that way, let me know what settings you landed on.

to my eyes, the best 1:1 settings between my two shields are:

- brightness: 0
- contrast: 255
- gamma: 1.35

catfact:
> EDIT: Oh and should I move new screen functions to `screen_events.c` even though they don't interact with cairo? See `screen_gamma()`, `screen_contrast()`, `screen_brightness()`, `screen_invert()`.

looking at `write_command_with_data`, my feeling is that it probably isn't threadsafe, so that would be a good idea. but if you know its threadsafe then no need.

tlubke:
gamma fix PR is up at https://github.com/monome/norns/pull/1724! (remember to revert display settings to default if doing a side-by-side with `main`)

## on script redraw without `screen.update()`

I recompiled without the display timer thread and did some testing. With that option, the screen doesn't really throw away or miss any frames. When pressing the "Run Script" (`⌘+P`) button in Maiden, I am able to see one frame of the menu system before norns goes into the script. Is it possible with the screen_event queue that there could be a leftover `SCREEN_EVENT_UPDATE` from the menu system that gets handled after `redraw()` is reassigned in Lua by the `Script` module? Based on looking at every instance of `Screen.update()`, `ssd1322_update()`, and `screen_update()` that's my only hypothesis so far. I might try adding a "non-emergency" `screen_events_clear()` and give it a Lua binding to call in context switching places.

## on hello.c flickering

Fiddled with reduced drawing rates to no avail.

## on screen events for new non-cairo functions

> looking at write_command_with_data, my feeling is that it probably isn't threadsafe, so that would be a good idea. but if you know its threadsafe then no need.

I don't think I ever tested the threadsafety explicitly, but all the critical functions (like `ssd1322_write_command()`) have mutex locks around them. The functions that set variables don't though, so maybe that is causing issues. I think I'll still add screen events for the new functions to be consistent with the rest of the functions.

catfact:
i'll try and check for the possiblity that there are refresh commands left in the buffer or being handled out of order

catfact:
>  When pressing the "Run Script" (⌘+P) button in Maiden, I am able to see one frame of the menu system before norns goes into the script. Is it possible with the screen_event queue that there could be a leftover SCREEN_EVENT_UPDATE from the menu system that gets handled after redraw() is reassigned in Lua by the Script module?

i took a short look and don't think that's exactly what happening. 

adding print statements to `screen_event_data_push()`, and `handle_screen_event()`, a quick scan doesn't show spurious update events or mismatch I/O sequences. 
admittedly:
- doesn't rule out a race condition that is eliminated by slowing things down with the prints themselves
- should probably make a program to parse and verify the matron output instead of relying on eyeballs

but that said... i don't think there is anything wrong with the screen event queue implementation, it's pretty straightforward.

----

as far as the glitch itself: i'm pretty sure i'm seeing the single frame after launching scripts. like, a single frame (+) of the script UI, then a single frame(+) of the last menu screen (_not_ the script launch confirmation screen, i think), then back to the script UI.

the sequence that makes me suspicious is `screen_save()` and `screen_restore()`, which are both called from `menu` as well as from `script`. 

my hunch is that to keep these continuing to work as expected within the lua program flow, they might need to also be made to block the main thread until they complete.

catfact:
ah well, i did look at it a bit more.

pretty trivially added blocking-on-results to those two functions. 

(though having looked and thought about it more, i would rather / also like to [somehow] add a routine for the main thread to block until all pending screen events are handled. that could be useful from lua, and would certainly be useful as a debug tool to functionally disable the asynchronous screen event stuff.)

anyways... i don't think it helped. but i am starting to think it is actually a bug in lua somehow. i think i am seeing a sequence of actual drawing commands on the stack that would correspond to drawing the script redraw once, then the script intro screen, then the script redraw again.

catfact:
oh right - in case it is useful i pused this debug branch to my fork:
https://github.com/catfact/norns/tree/debug-screen-refresh

catfact:
like ok, here is some output. 

```
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: --- menu.lua: edit mode (page redraw)
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: --- preview redraw ---
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 27
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 12
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 27
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 22
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 1
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: ### pushing SCREEN_EVENT_UPDATE
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 12
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 22
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 1
Oct 28 19:45:20 norns-shield-rpi4 ws-wrapper[4640]: ### handling SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: --- preview redraw ---
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 24
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### pushing SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: # script clear
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 24
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### handling SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> weaver: screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 3
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> waiting for results from screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 3
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> received results, continuing
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 6
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 8
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 4
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 5
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> weaver: screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 2
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> waiting for results from screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 6
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 8
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 4
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 5
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 2
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> received results, continuing
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: # script load: /home/we/dust/code/0_ascreen/grad.lua
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: # script run
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >> reading PMAP /home/we/dust/data/0_ascreen/grad/grad.pmap
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: m.read: /home/we/dust/data/0_ascreen/grad/grad.pmap not read, using defaults.
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: Engine.register_commands; count: 0
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ___ engine commands ___
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ___ polls ___
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: amp_in_l
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: amp_in_r
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: amp_out_l
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: amp_out_r
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: cpu_avg
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: cpu_peak
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: pitch_in_l
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: pitch_in_r
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: # script init
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> weaver: screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 2
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> waiting for results from screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 2
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> received results, continuing
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: --- menu.lua: edit mode (page redraw)
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### pushing SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> weaver: screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 3
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> waiting for results from screen_event_restore...
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 22
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### handling SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 3
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: >>> received results, continuing
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: --- menu.lua: play mode (clearing)
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: --- menu.lua: play mode (update after clear)
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### pushing SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: --- menu.lua: play mode (assigning script redraw)
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: --- menu.lua: play mode (running script redraw)
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: --- script redraw ---
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 6
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 8
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 13
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 20
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: screen_event_data_push: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### pushing SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### handling SCREEN_EVENT_UPDATE
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 27
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 6
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 7
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 8
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 12
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 13
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 20
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: handle_screen_event: 1
Oct 28 19:45:25 norns-shield-rpi4 ws-wrapper[4640]: ### handling SCREEN_EVENT_UPDATE
```

it starts with showing the preview page. (at 19:25:20). 

5 seconds later i press K3 to confirm. then a lot of things happen that i don't totally follow. the preview page is certainly drawn twice more, then save/restore are called, then the script UI at the end.

hm.

-----

OH! `screen_save` is being handled in the midst of handling a sequence of drawing commands from a single redraw. that is wrong. 

so... we have to not (only) block on completion of `save`, we really do have to hold off on saving the surface as long as there are pending drawing commands. so that is the next show stopper fix that needs to be implemented.

catfact:
Actually... Thinking about it I'm not sure how a "torn" rendering routine could happen. Events are always dispatched in order, including save / restore.

Id at least hypothesize that what we're seeing is a real effect of the requested drawing sequences, which we couldn't see before because the screen wouldn't refresh fast enough to show all the surface states resulting from that one key interaction.

tlubke:
@catfact These last few comments have been full of really great information and gave me some ideas. I think you're right in that it is primarily a Lua issue that has arisen because of this big change in process.

I think what may be happening is although the screen events are in order,  because of the timer thread in `ssd1322.c`, a refresh will likely take place _after_ all the screen events are fired. Here's an example below to illustrate.

```
Screen Event Queue
-------------------
SCREEN_EVENT_CLEAR              <- norns.blank redraw()
SCREEN_EVENT_UPDATE
SCREEN_EVENT_LEVEL              <- script redraw()
SCREEN_EVENT_RECT
SCREEN_EVENT_FILL
```

Because `ssd1322.c` will still have a pointer to the surface it will probably transfer the data to the display after it has flushed from the surface. If I spam click the script play button, I am occasionally able to get a partially drawn version of `redraw()`. That behavior would seem to support the theory. I guess in my mind, this would be another reason to support the no-timer build over the one with the thread.

tlubke:
Have a big-ish commit that I'll push to #1724 tomorrow morning. I think it will fix everything that's been discussed so far.

tlubke:
Commit up: https://github.com/monome/norns/pull/1724/commits/747f1ff3eae9953a99791ddaab6b2947ded42cee

tehn:
what's the current thinking with the screen timer? compile with or without for testing?

tehn:
fantastic--- will get on testing!

lua/core/screen.lua
```diff
@@ -477,4 +557,22 @@ Screen.blend_mode = function(index)
...
+-- i == 0 -> normal mode
+-- i < 0 -> inverted mode
+-- i > 0 -> inverted mode
+--
+-- if boolean:
+-- false -> normal mode
+-- true -> inverted mode
+Screen.invert = function(i)
+  if type(i) == "boolean" then
+    _norns.screen_invert(1 and i or 0)
```
dndrks, line 572: oo should this be `_norns.screen_invert(i and 1 or 0)`?

tlubke: > oo @tlubke should this be _norns.screen_invert(i and 1 or 0)?

yes 😔

tehn:
testing is looking good! thank you all for your time on this. i'm still blown away whenever i see oooooo running high frame rate.

it seems there isn't anything blocking going into a beta for this? (main issue being, reverting from the beta is a little fiddly, unless i do an intermediary release which "resets" the boot files.) but i can't decide if that's really necessary--- if there's really a case where someone would need to revert, that's not also capable of hand-fixing the files (since the beta would be for "experienced" testers)

catfact:
seems fine to me, the change back is just commenting a single line in `boot/config.txt`

tlubke:
> testing is looking good! thank you all for your time on this. i'm still blown away whenever i see oooooo running high frame rate.

That's great! Can't believe it's been like 2 years since that original post on the display gamma. Glad to see it finally come to fruition. Was nice to work with you on this, @catfact.

> it seems there isn't anything blocking going into a beta for this? (main issue being, reverting from the beta is a little fiddly, unless i do an intermediary release which "resets" the boot files.) but i can't decide if that's really necessary--- if there's really a case where someone would need to revert, that's not also capable of hand-fixing the files (since the beta would be for "experienced" testers)

I agree it's probably fine just to have people manually edit the config file.

One last request to put out there -- could we add another co-author on one of these commits before at the time it gets merged? @Dewb helped me out quite a bit on the ARM vector intrinsic stuff initially.

lua/core/screen.lua
tehn: fixed

tehn:
added @Dewb ! merging now and will get this into the beta pipeline

PR #1723, opened 2023-10-27
Fix aliasing issues found in ssd1322.c mentioned in (#1720)

tlubke:


tlubke:
@dndrks I got my norns screen looking like [the image on the right](https://github.com/monome/norns/pull/1717#issuecomment-1779848346), but if you could checkout this branch and test again -- nothin' can beat a side-by-side comparison.

catfact:
looking great here, nice work!

(i'd merge it now, but let's leave it separate for A/B convenience until the others get a look.)

tehn:
fantastic!!

i'm going to merge this as the reference can be against `main`

PR #1725, opened 2023-10-30
update softcut to get phase inversion fix

catfact:
so it turns out, softcut has i think a pretty long standing bug where phase is inverted on recording! i don't think it was always there, but somewhere in messing around with fade curves a math error was introduced. 

this pulls in actually a few changes from softcut main branch, but the only significant thing is the single change in `FadeCurves.cpp`. 

apologies, i think people have pointed it out but we always thought it was some polarity issue elsewhere, i never thought to check the sign of the fade curve.

i almost wonder if we even should change the default behavior at this point.


tehn:
good find, thank you. what would you prefer the default behavior to be? (i'm all for changing it if you prefer, since it seems best practice for scripts to explicitly set all params to desired values)

tehn (approved):

PR #1727, opened 2023-11-01
Add `Screen.sleep()` to manually invoke the screensaver

ambv:
The default screensaver time is 15 minutes. I like to manually turn the screensaver on sooner, which I can do from Maiden by calling:
```lua
>> screensaver:event()
```

To be able to do this, `screensaver` needs to be a global.

### DIFF
--- a/lua/core/screen.lua
+++ b/lua/core/screen.lua
@@ -477,4 +477,8 @@ Screen.blend_mode = function(index)
   end
 end
 
+Screen.sleep = function()
+  screensaver:event()
+end
+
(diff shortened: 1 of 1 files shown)


tehn (changes requested):
i'd prefer this to be a more explicit function rather than just global a specialized timer.

can you instead make a function `screen.sleep()` ?

ambv:
Oh, I like this.

ambv:
Done!

tehn (approved):

PR #1728, opened 2023-11-02
reflection: tighten timing

dndrks:
small change, but tightens up the overall experience of a specific use.

@ryleelyman + @sonocircuit, this is an opinionated test case -- i'd be curious to know if your scripts + usage of the library is at all negatively affected by this proposed change?

(oh, right, also 1/48 seems to be the lowest possible quantization, so i changed the API note about that)

<details><summary>test script (from forthcoming 'reflection' extended ref docs)</summary>

```lua
_r = require 'reflection'
my_pattern = _r.new()

g = grid.connect()

function init()
  lit = {}
  my_pattern.process = process_press
  
  params:add_option(
    "demo",
    "demo",
    {"1","2","3"},
    1
  )
  params:set_action("demo",
    function(x)
      if x == 1 then
        params:set("record_duration", 4)
        params:set("hold_rec", 1)
        params:set("rec_sync_value", 3)
        params:set("play_sync_value", 3)
        params:set("loop", 2)
      end
    end
  )
  
  record_duration = 0
  params:add_number(
    "record_duration",
    "record duration",
    0,
    128,
    record_duration,
    function(param) return (
      param:get() == 0 and 'free' or
      param:get()..' beats'
    ) end
  )
  params:set_action("record_duration", function(x) record_duration = x end)
  
  hold_rec = true
  params:add_option(
    "hold_rec",
    "hold rec for first event?",
    {"no","yes"},
    hold_rec and 2 or 1
  )
  params:set_action("hold_rec", function(x) hold_rec = x == 2 end)
  
  rec_sync_value = nil
  params:add_option(
    "rec_sync_value",
    "sync record start",
    {"free","next beat", "next bar"},
    1
  )
  params:set_action("rec_sync_value",
    function(x)
      if x == 1 then
        rec_sync_value = nil
      elseif x == 2 then
        rec_sync_value = 1
      elseif x == 3 then
        rec_sync_value = 4
      end
    end
  )
  
  play_sync_value = nil
  params:add_option(
    "play_sync_value",
    "sync play start",
    {"free","next beat", "next bar"},
    1
  )
  params:set_action("play_sync_value",
    function(x)
      if x == 1 then
        play_sync_value = nil
      elseif x == 2 then
        play_sync_value = 1
      elseif x == 3 then
        play_sync_value = 4
      end
    end
  )
  
  params:add_option(
    "loop",
    "loop playback?",
    {"no", "yes"},
    1
  )
  params:set_action("loop", function(x) my_pattern:set_loop(x-1) end)
  
  params:add_trigger(
    "erase_rec",
    "erase recording?"
  )
  params:set_action("erase_rec", function(x) my_pattern:clear() lit = {} grid_redraw() end)
  
  params:add_trigger(
    "double_rec",
    "double recording"
  )
  params:set_action("double_rec", function(x) my_pattern:double() end)
  
  overdubbing = false
  playback_queued = false
  
  params:bang()
  
  grid_redraw()
end

function my_pattern.start_callback()
  print('playback started', clock.get_beats())
  playback_queued = false
  grid_redraw()
end

function my_pattern.end_of_rec_callback()
  print('recording finished', clock.get_beats())
  grid_redraw()
end

function my_pattern.end_of_loop_callback()
  print('loop ended', clock.get_beats())
  grid_redraw()
end

function my_pattern.end_callback()
  print('playback ended')
  if my_pattern.loop == 0 then
    overdubbing = false
  end
  lit = {}
  grid_redraw()
end

function g.key(x,y,z)
  if x == 1 and y == g.rows then
    if z == 1 then
      if my_pattern.rec == 0 and my_pattern.queued_rec == nil and my_pattern.count == 0 then
        my_pattern:set_rec(hold_rec and 2 or 1, record_duration > 0 and record_duration or nil, rec_sync_value)
      elseif my_pattern.count > 0 and my_pattern.play == 0 then
        if play_sync_value ~= nil then
          clock.run(
            function()
              playback_queued = true
              clock.sync(play_sync_value)
              my_pattern:begin_playback()
            end
          )
        else
          my_pattern:start()
        end
      elseif my_pattern.play == 1 then
        my_pattern:stop()
      else
        my_pattern:set_rec(0)
      end
    end
  elseif x == 1 and y == g.rows - 1 then
    if z == 1 then
      params:set("loop", params:get("loop") == 1 and 2 or 1)
    end
  elseif x == 1 and y == g.rows - 2 then
    if z == 1 then
      if my_pattern.count > 0 and my_pattern.play == 1 then
        overdubbing = not overdubbing
        my_pattern:set_rec(overdubbing and 1 or 0)
      end
    end
  else
    local event = {
      id = x*8 + y,
      x = x,
      y = y,
      z = z
    }
    my_pattern:watch(event)
    process_press(event)
  end
  grid_redraw()
end

function process_press(e)
  if e.z == 1 then
    lit[e.id] = {
      x = e.x,
      y = e.y
    }
    -- print(clock.get_beats())
  else
    if lit[e.id] ~= nil then
      lit[e.id] = nil
    end
  end
  grid_redraw()
end

function grid_redraw()
  g:all(0)
  if my_pattern.queued_rec then
    g:led(1,g.rows,7)
  elseif playback_queued then
    g:led(1,g.rows,8)
  elseif my_pattern.rec == 1 then
    g:led(1,g.rows,15)
    g:led(1,g.rows-1,15)
  elseif my_pattern.play == 1 then
    g:led(1,g.rows,10)
  elseif my_pattern.count > 0 then
    g:led(1,g.rows,5)
  else
    g:led(1,g.rows,2)
  end
  g:led(1,g.rows-1, my_pattern.loop == 1 and 10 or 2)
  g:led(1,g.rows-2, overdubbing and 10 or 2)
  for i,e in pairs(lit) do
    g:led(e.x, e.y, 15)
  end
  g:refresh()
end
```
</details>

**test process**

- load test script
- switch clock to Link and connect norns to a Link session to get an audible metronome from a DAW
- press the last key in the first column to queue recording, which will start on the "1"

currently, starting recording with the expectation of immediate playback after the beat-count has been met pushes event playback timing by 1/96th:

```bash
playback started	92.001814819443
recording finished	96.012302
playback started	96.013124569442
loop ended	100.010620125
...
```

with the changes in this PR, we account for this with a negative 1/96 offset when we fire off the very first recording clock.
the same test script yields the desired results:

```bash
playback started	11.990809791667
recording finished	16.000324013889
playback started	16.00218373611
loop ended	20.001766652776
...
```

this also gives us the added bonus of capturing any hyper-precise entry _on_ the sync beat, as the recording mechanism is technically starting ahead of the sync beat.


ryleelyman:
love this! i'll have a play with it tonight.

while we're making edits, i'd love to get feedback on the following small changes i'm using in `xD2`:

```diff
@@ -22,6 +22,7 @@ function reflection.new()
   p.quantize             = 1 / 48
   p.endpoint             = 0
   p.start_callback       = function() end
+  p.step_callback        = function() end
   p.end_of_loop_callback = function() end
   p.end_of_rec_callback  = function() end
   p.end_callback         = function() end
@@ -211,12 +212,13 @@ function reflection:begin_playback()
           if self.loop == 1 then
             self.step = 0
             self:_clear_flags()
-            self:start_callback()
+            self.start_callback()
           end
         end
       end
     else
       if self.step % q ~= 1 then goto continue end
+      self.step_callback()
       for i = q - 1, 0, -1 do
         if self.event[self.step - i] and next(self.event[self.step - i]) then
           for j = 1, #self.event[self.step - i] do
```

the edit `self:start_callback()` -> `self.start_callback()` is mostly for consistency with the rest of the callbacks. for a second I wanted to make all of these functions into methods, but you can easily recreate that functionality without that by doing things like this:

```lua
local narcissus = reflection.new()
narcissus.start_callback = function()
  -- just use the `narcissus` variable if you need to!
end
```

the new feature, `step_callback()`, allows for things like creating subloops or jumping around.

sonocircuit:
hihi! @dndrks extended api test script looks fantastic, thank you for putting this together! Unfortunately, I won't be able to give it a spin over the next few days as I have a _crushing_ amount of work to get done.


- re negative offset: this might (or not) be related to a small bug I've been wanting to report. basically when syncing recording to a beat sync value, if the first event falls dead on the beat (which would correspond to `self.step == 1`) it wouldn't get recored. I found out that `reflection:watch()` would actually see `self.step == 0`, which is why the event never would get played back. I fixed this by changing `local s = math.floor(step_one == true and 1 or self.step)` to `local s = math.floor(step_one == true and 1 or self.step + 1)` within `reflection:watch()`. If any or both of you get round to test this that would be great.

- @ryleelyman consistency is a good thing! re step_callback: yes!

dndrks:
rad, @ryleelyman got your wonderful change in!! tyty!
@sonocircuit , great call re: first-step being missed! i was worried that adding a `+ 1` offset to all steps would unintentionally shift things, so i just added a condition for step 0:
https://github.com/monome/norns/blob/48f132865308863e310483eddfd22e70e62330ca/lua/lib/reflection.lua#L178

---

generally, things feel very performant and i'd be down to call this 1.0!

in the "would be rad to eventually figure this out" bucket, [pattern playback syncs *before* advancing](https://github.com/monome/norns/blob/48f132865308863e310483eddfd22e70e62330ca/lua/lib/reflection.lua#L193), which means that 'step 1' technically occurs 1/96th after it should. but in `reflection`, simply dropping the `clock.sync(1/96)` down to the end of the advancement function creates a new problem -- steps are in time, but the pattern callbacks are 1/96th _early_.

eg. general case (how `reflection` is set up now):
```lua
function init()
  clock.run(go)
end

function go()
  clock.sync(1)
  clock.run(advance)
end

function advance()
  local step = 0
  print("pattern starting: "..clock.get_beats())
  while true do
    clock.sync(1/96)
    step = step + 1
    if step == 1 then print("step 1: "..clock.get_beats()) end
    if step == 384 then
      step = 0
      print('loop restarting', clock.get_beats())
    end
  end
end
```

will print:
```bash
pattern starting: 8132.001552 -- on time
step 1: 8132.0107841667 -- late
loop restarting	8136.0015766944 -- "on time"
step 1: 8136.0109561667 -- late
```

versus syncing _after_ the action calls the callbacks early:

```lua
function init()
  clock.run(go)
end

function go()
  clock.sync(1)
  clock.run(advance)
end

function advance()
  local step = 0
  print("pattern starting: "..clock.get_beats())
  while true do
    step = step + 1
    if step == 1 then print("step 1: "..clock.get_beats()) end
    if step == 384 then
      step = 0
      print('loop restarting', clock.get_beats())
    end
    clock.sync(1/96)
  end
end
```

which prints:
```
pattern starting: 8183.0002412778 -- on time
step 1: 8183.000301 -- on time
loop restarting	8186.9908284444 -- "early"
step 1: 8187.0015485833 -- on time
```

i'm sure there's a way to rewrite some bits so we can have both on-time steps and on-time callbacks. for now, i guess the big q is how y'all expect the callbacks to be used + whether those need to be _on an exact whole beat_ <333

lua/lib/reflection.lua
```diff
@@ -128,7 +129,7 @@ end
 
 --- quantize playback
 -- @tparam float q defaults to 1/48
--- (should be at least 1/96)
+-- (should be at least 1/48)
```
ryleelyman, line 132: why this change? it seems like 1/96 should behave as expected?

dndrks: oh! the steps don't seem to execute with 1/96, couldn't tell if it was intentional. playback starts and eventually ends with this test script, but no step actions are executed:

```lua
_r = require 'reflection'
my_pattern = _r.new()

function init()
  my_pattern.process = process_pat
  my_pattern.event = {
    [1] = {[1] = {st = 'hello beat 1'}},
    [96] = {[1] = {st = 'hello beat 2'}},
    [192] = {[1] = {st = 'hello beat 3'}},
    [288] = {[1] = {st = 'hello beat 4'}},
  }
  my_pattern.count = 3
  my_pattern:set_length(4)
  my_pattern:set_quantization(1/96)
  clock.run(
    function()
      clock.sync(4)
      my_pattern:begin_playback()
    end
  )
end

function my_pattern.start_callback()
  print('playback started', clock.get_beats())
end

function my_pattern.end_of_rec_callback()
  print('recording finished', clock.get_beats())
end

function my_pattern.end_of_loop_callback()
  print('loop ended', clock.get_beats())
end

function my_pattern.end_callback()
  print('playback ended', clock.get_beats())
end

function process_pat(e)
  print(e.st, clock.get_beats())
end
```

dndrks: mmm:
https://github.com/monome/norns/blob/03f158816346405c4b595f7602ebd66216be2c87/lua/lib/reflection.lua#L209

sets `q` to `1` when `self.quantize` is `1/96`, which fails this check later:

https://github.com/monome/norns/blob/03f158816346405c4b595f7602ebd66216be2c87/lua/lib/reflection.lua#L235

ryleelyman: oh that's bad math on my part ... still not thinking about it!! but excited to dig in later

dndrks: hoho! yeah, `if q ~= 1 and self.step % q ~= 1 then goto continue end` fixes it up!

lua/lib/reflection.lua
```diff
@@ -188,8 +190,22 @@ function reflection:begin_playback()
   self.step = 0
   self.play = 1
   self.start_callback()
+  local queued_start_callback = false
+  local queued_end_of_loop_callback = false
+  local queued_end_playback = false
   while self.play == 1 do
-    clock.sync(1 / 96)
+    if queued_start_callback then
+      self:start_callback()
```
ryleelyman, line 198: typo, should be `self.start_callback()` to match the others.

lua/lib/reflection.lua
```diff
@@ -188,8 +190,22 @@ function reflection:begin_playback()
...
+      queued_start_callback = false
+    end
+    if queued_end_of_loop_callback then
+      self.end_of_loop_callback()
+      queued_end_of_loop_callback = false
+    end
+    if queued_end_playback then
+      self:end_playback()
```
ryleelyman, line 206: interesting: the way this is set up has a new possible edge case where the current step processes one last time _after_ `end_playback()` is called. Maybe better to add a `break` here so that doesn't happen?

lua/lib/reflection.lua
```diff
@@ -212,12 +228,13 @@ function reflection:begin_playback()
...
             self.step = 0
             self:_clear_flags()
-            self:start_callback()
+            queued_start_callback = true
           end
         end
       end
     else
-      if self.step % q ~= 1 then goto continue end
+      if q ~= 1 and self.step % q ~= 1 then goto continue end
```
ryleelyman, line 236: love this, thanks for catching it!

lua/lib/reflection.lua
```diff
@@ -212,12 +228,13 @@ function reflection:begin_playback()
...
+      self.step_callback()
       for i = q - 1, 0, -1 do
         if self.event[self.step - i] and next(self.event[self.step - i]) then
           for j = 1, #self.event[self.step - i] do
```
ryleelyman, line 240: This line and the next could be made shorter, I think, by instead just saying
```lua
for _, event in ipairs(self.event[self.step - i]) do
  if not event._flag then self.process(event) end
end
```

lua/lib/reflection.lua
```diff
@@ -212,12 +228,13 @@ function reflection:begin_playback()
           if self.loop == 1 then
             self.step = 0
             self:_clear_flags()
-            self:start_callback()
+            queued_start_callback = true
           end
         end
       end
     else
```
ryleelyman, line 235: might be nice to add a little comment here saying
`-- not first pass; we'll process`
or something similar; this helps me parse the other comment in this function better

ryleelyman (changes requested):
just a few tweaks, otherwise this looks great! thank you so much for tackling it!

lua/lib/reflection.lua
dndrks: _good_ **call**

lua/lib/reflection.lua
dndrks: yesssss

dndrks:
rad, rad, rad -- @ryleelyman , got it all folded in and testing well!!

@tehn , ready for merge!

ryleelyman (approved):
great work, thanks a ton!!

dndrks:
thank _you_!! this is such an often-asked-for library, i'm really stoked to help how i can and for folks to get to script with it!

tehn (approved):

sonocircuit:
@dndrks wow, this is amazing, thank you so much!

@dndrks @ryleelyman there's one last thing I'd like to address and then I think it's ready to go. currently (and in previous versions) when using `queued_rec` the events recorded after `step_one` were off by 1 step (1/96). Now that recording starts one step in advance they are off by two steps. If it's ok with you I'll work on some proposed changes and test script tonight/tomorrow and report back.

ryleelyman:
oh, thanks for this, @sonocircuit, that would be great !

PR #1729, opened 2023-11-07
Set step for MIDI* controlspec templates to 1

ambv:
Without this change, scrolling through values of a param that uses those controlspecs only hits integer values at the ends of the spectrum.


ambv:
To reproduce, add a param like:
```
    params:add{
        type = "control",
        id = "root_note",
        name = "root note",
        controlspec = controlspec.MIDINOTE,
        formatter = function(param)
            return MusicUtil.note_num_to_name(param:get(), true)
        end
    }
```

Observe in Norns UI how editing the values only shows changes at the end of the spectrum.

Once step is set to 1, everything works well.

dndrks:
hihi! hope all's well :)
oh, good catch that these are not returning useful values!
in testing the changes you proposed, though, i remembered that output quantization (`1`) only _rounds_ the input value -- so we actually end up skipping some steps when we apply +1/-1 deltas, eg:

```lua
MusicUtil = require 'musicutil'

function init()
  params:add{
        type = "control",
        id = "root_note",
        name = "root note",
        controlspec = controlspec.new(0, 127, 'lin', 1, 60, ""),
        formatter = function(param)
            return MusicUtil.note_num_to_name(param:get(), true)
        end
    }
  params:set_action('root_note', function(x) print(x) end )
end
```

results in:

```lua
60 -- default
61 -- 61.27, rounded
63 -- 62.54, rounded
```

when we add the MusicUtil formatter, this skips `D3` altogether.

to remedy, we also need to quantize our controlspec's input value to `1/127` (generally, `1/max-min`), eg:

```lua
controlspec = controlspec.new(0, 127, 'lin', 1, 60, "", 1/127),
```

which gives us the full incremental range!

lua/core/controlspec.lua
```diff
@@ -181,11 +181,11 @@ ControlSpec.PHASE = ControlSpec.new(0, math.pi, 'lin', 0, 0, "")
 --- converts to an exponential RQ range
 ControlSpec.RQ = ControlSpec.new(0.001, 2, 'exp', 0, 0.707, "")
 --- converts to a MIDI range (default 64)
-ControlSpec.MIDI = ControlSpec.new(0, 127, 'lin', 0, 64, "")
+ControlSpec.MIDI = ControlSpec.new(0, 127, 'lin', 1, 64, "")
```
dndrks, line 184: suggested:
`ControlSpec.MIDI = ControlSpec.new(0, 127, 'lin', 1, 64, "", 1/127)`

lua/core/controlspec.lua
```diff
@@ -181,11 +181,11 @@ ControlSpec.PHASE = ControlSpec.new(0, math.pi, 'lin', 0, 0, "")
...
 --- converts to a range for MIDI notes (default 60)
-ControlSpec.MIDINOTE = ControlSpec.new(0, 127, 'lin', 0, 60, "")
+ControlSpec.MIDINOTE = ControlSpec.new(0, 127, 'lin', 1, 60, "")
```
dndrks, line 186: suggested:
`ControlSpec.MIDINOTE = ControlSpec.new(0, 127, 'lin', 1, 60, "", 1/127)`

lua/core/controlspec.lua
```diff
@@ -181,11 +181,11 @@ ControlSpec.PHASE = ControlSpec.new(0, math.pi, 'lin', 0, 0, "")
...
 --- converts to a range for MIDI velocity (default 64)
-ControlSpec.MIDIVELOCITY = ControlSpec.new(1, 127, 'lin', 0, 64, "")
+ControlSpec.MIDIVELOCITY = ControlSpec.new(1, 127, 'lin', 1, 64, "")
```
dndrks, line 188: suggested:
`ControlSpec.MIDIVELOCITY = ControlSpec.new(1, 127, 'lin', 1, 64, "", 1/126)`

dndrks (changes requested):
thank you so much!!!

lua/core/controlspec.lua
ambv: Since it's between 0 and 127, shouldn't it be 1/128?

dndrks: it's a good q, but no, quantum is `1/max-min`.

eg. running this code:
```lua
function init()
  params:add{
        type = "control",
        id = "root_note",
        name = "root note",
        controlspec = controlspec.new(0, 11, 'lin', 1, 0, "", 1/12), -- 1/12 vs 1/11
    }
  params:set_action('root_note', function(x) print(x) end )
end
```

repeated `params:delta('root_note',1)` print:
```
1
2
3
4
5
5
```

whereas following `1/max-min` and using 1/11 for the quantum gets us the expected increments.

ambv:
I noticed that some values were skipped, happy you explained what was going on! I didn't think it was a big deal because you can always K3 + E3 to fine-edit.

Great insight about how to use `quantum` in `controlspec`. Thanks! Can we add this to the docs?

dndrks (approved):
rad, thank you!
and yes, can add to https://monome.org/docs/norns/reference/controlspec

PR #1745, opened 2023-12-03
add Particle font

dansimco:
This pr adds in a font I've created for compact norns ui. Discussed on the study group discord [here](https://discord.com/channels/765746584582750248/822935549659971594/1178444477816389796)

## important

**I've added this font in to the list in it's alphabetically sorted position, which makes it `font_face(3)` this is potentially a breaking change for scripts which had used Roboto-Thin, which would now be font_face(4) and so on down the line. It might be better to add new fonts like this to the end of the list, though that would break the alpha sorting.** 

specimen

The font itself is open source and licensed under SIL OPEN FONT LICENSE for collaboration.


ambv:
Yeah, for backward compatibility I think it shouldn't be `font_face(3)`.

tehn:
i do think we need to change this so that the index is at the end as to not break compatibility. in the end, it's just the code that sees the alphabetization.

separately, we should add a lookup table from lua so we could do `font_face('particle')` etc!

this font looks so good!!

dansimco:
Ok PR updated. Particle is now `font_face(68)` . I'll be away from home for a couple of weeks, I can make changes via github ui but wont be able to test until ~12/17

tehn (approved):

PR #1763, opened 2024-02-21
support i2c adc inf pots for encoders

tehn:
refreshed norns PCB uses inf pots via an i2c ADC.

no config needed for mode switching.


xmacex:
Someone (maybe [me](https://llllllll.co/t/i2c-questions-discussions/13624/41), a fool) will sooner or later be plugging attenuated modular voltages into those i2c ADCs ;)

PR #1768, opened 2024-03-05
fix: always push four quads to libmonome

### LINKED ISSUE #1767: Grid LEDs do not illuminate in vertical orientation
I'm working on a script using an 8x16 grid from the most recent batch on norns 2.8.1/norns 231114 and I'm unable to light all the keys when it's in a vertical orientation. Even with the simplest approach, I can only get the top 8x8 section to illuminate.

A really small script that expresses this problem:

 ```
g = grid.connect()
g:rotation(1)
g:all(5)
g:refresh()
```

I don't believe this to be a mechanical issue because the opposite 8x8 set lights up if I rotate 270 degrees. Similarly, if the rotation is 0 or 180 everything lights up.

When I log key presses after `g:rotation(1)` with  `function g.key(x,y,z)` I get messages with appropriate x and y coordinates so this seems to be limited to illumination.

Additionally, I saw this [post on lines](https://llllllll.co/t/grid-rotation-on-norns/31660/16?u=new_moon) which makes me suspect this is an issue that isn't just affecting me.

ryleelyman:
when grid rotation is nonzero, device_monome.c is writing to quads that end up never being pushed to the device (quad 3, specifically). i'm not familiar with libmonome, and i have yet to test this on my norns, but this is the solution that i used in ryleelyman/seamstress#113 to get grid rotation working as expected.

closes #1767

### DIFF
--- a/matron/src/device/device_monome.c
+++ b/matron/src/device/device_monome.c
@@ -119,7 +119,7 @@ void dev_monome_arc_set_led(struct dev_monome *md, uint8_t n, uint8_t x, uint8_t
 
 // set all LEDs to value
 void dev_monome_all_led(struct dev_monome *md, uint8_t val) {
-    for (uint8_t q = 0; q < md->quads; q++) {
+    for (uint8_t q = 0; q < 4; q++) {
         for (uint8_t i = 0; i < 64; i++) {
             md->data[q][i] = val;
         }
@@ -136,7 +136,7 @@ void dev_monome_refresh(struct dev_monome *md) {
         return;
     }
 
-    for (int quad = 0; quad < md->quads; quad++) {
+    for (int quad = 0; quad < 4; quad++) {
         if (md->dirty[quad]) {
             if (md->type == DEVICE_MONOME_TYPE_ARC) {
                 monome_led_ring_map(md->m, quad, md->data[quad]);


catfact:
good catch, thank you!

however, i note that we originally had that value hardcoded to 4, and changed it in 2021 when the new grids were introduced. this was to avoid some specific issue that i've now forgotten. (my memory is that the latest grids will crash or something if they get data for nonexistent quads, but perhaps a firmware update has addressed this in the meantime, or something.)

catfact:
https://github.com/monome/norns/pull/1358/files#diff-4f52f40200f35bb22c3fa7098c08becae2ed172ad4a92d646d0cbe753b028a84L120-R128

catfact:
... so sure, we can merge this, but brace for some other forgotten issue to re-emerge.

i think the "right" way to fix this is to only transmit the quads that are relevant under the current rotation.

tehn:
firmwares (by now) should be resistant to non-applicable quads, but ideally libmonome shouldn't transmit irrelevant quads to begin with as it quite wasteful sending all that data over serial.

i need to re-familiarize myself with the libmonome rotation implementation, but it seems bizarre that the orientation of the quads wouldn't "follow" the orientation.

i'm somewhat not in favor of merging this, as it'd be much better to fix it properly. which also means deciding how to fix it. i can't believe this hasn't come up during the long life of this project (libmonome, i mean).

catfact:
let's back up and clarify a bit,

there are two changes here:

- first change makes it so that data is copied to all four quads in the internal per-device LED buffer. (which is always 4 quads worth, a trivial waste of RAM.) this is a complete waste of CPU when the device only uses 2 quads.

- second change makes it so that all 4 quads are sent to the device via libmonome. i definitely remember this being a problem with early firmware on ST grids. i haven't looked into libmonome internals to know whether it will actually transmit serial data for unused quads. if it does it is a waste of serial bandwidth for sure.

since serial data takes fixed time per byte, wasting serial bandwidth would be worse.

---

anyways, it doesn't seem like we store the rotation data here in the norns-side "driver," so i think we should, and then it's rather simple to make it correct.

ryleelyman:
cool, happy to work on adding rotation handling here, or closing in favor of another PR, just let me know what would be useful from me :)

catfact:
okay, so @tehn i am looking at libmonome rotation logic implementation. it is pretty nuts and has no documentation so finding usage unclear. but my guess is that it will "just work" if we just send the correct quads and don't worry about anything else.

my first guess is that for a 128 grid, the correct quads are gonna be either [0, 1] (for rotation of 0, 180) or [0, 2] (for rotation of 90, 270)

anyway should be easy enough to test

catfact:
@ryleelyman thank you. i think we should:
-  add a rotation field to `device_monome.h` (could just be an enum with 4 values right?)
- set it from `dev_monome_set_rotation`, instead of just passing the argument through to libmonome
- check the rotation value when copying and transmitting quad data

right?

tehn:
@catfact this i think is the correct solution. i can get to this tomorrow unless someone else jumps on it.

tehn:
will do this the better way.

Issue #1758, opened 2024-02-12
tape playback unresponsive if preview is not stopped


tapecanvas:
I've noticed an issue on my norns shields (all running most recent update):

In Tape, if a file is previewed and the preview is not stopped with e3 before selecting the file, there will be not output from Tape, even though the screen displays the selected file and playback display.
 
After this happens, I am unable to hear the preview of any other files, and any new file selected (if previewed or not) will also not play in Tape.

Upon testing this while I write this issue, I noticed that if I preview a file and leave that preview running and select another file to preview in the same directory, all preview sounds stop. Then if a file is selected to play in Tape, no sound is produced

My method around this for now has been: 
"After turning k3 to the right to preview, turn it back to the left to end preview before going back to preview another file. Likewise if you want to select the previewed file to play in tape, stop the preview before selecting."

[link to lines thread that also mentions this problem](https://llllllll.co/t/tape-playback-not-working-if-previewed/66088)

ryleelyman:
as i mentioned on lines, i'm happy to take a stab at this. it sounds like somehow whatever function turning E3 left is calling just needs to be called in more cases. i have vague memories of attempting that fix already, but i wouldn't be surprised if i missed something...

catfact:
we can't reliably fix this by adding debounce timers in lua.

PR #1759, opened 2024-02-13
fix: more aggressively debounce tape playback

ryleelyman:
closes #1758

there's a lot to dislike here—i added a half-second slowdown to _every_ time fileselect exits—this is because if tape playback is going to start, it needs a buffer of about that long from after being stopped.

that being said, i believe the reported issues are now fixed.

PR #1770, opened 2024-03-10
tape: return false immediately on open() if stream is running

catfact:
this is a quick bandaid to begin addressing https://github.com/monome/norns/issues/1758

previous to the change, calling `open()` on a running stream would bork the state machine. 

with the change, calling `open()` on a running stream does nothing. not ideal but at least it keeps working.

next, i'll try and update such that the same action starts a fadeout and blocks the caller until fadeout is complete. i'd recommend merging this change first and i'll build on it (can't promise a timeline)

### DIFF
--- a/crone/src/Tape.h
+++ b/crone/src/Tape.h
@@ -70,10 +70,16 @@ namespace crone {
             virtual // from any thread
                 void start() {
                 if (isRunning) {
+                    std::cout << "Tape: already running" << std::endl;
                     return;
                 } else {
+                    std::cout << "Tape: starting.." << std::endl;
                     envIdx = 0;
                     envState = Starting;
+
+                /// FIXME: diskLoop() isn't designed to be re-entrant, 
+                // so we need to make very sure here that no existing disk loop thread is running
+
                     this->th = std::make_unique<std::thread>(
                                                              [this]() {
                                                                  this->diskLoop();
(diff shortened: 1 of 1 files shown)


tehn (approved):

tehn:
will create a rollback for https://github.com/monome/norns/pull/1759 and test

tehn:
the fact that it returns and doesn't try to play i believe makes it so the previous bug isn't happening (i tested on hardware, though perhaps i'm clueless as to how the original bug was manifesting)

currently the UI is sometimes not-reflective of the play state (ie, says it's playing but it's not) but that's not what i perceived the original bug to be

catfact:
Yes that's what I think too. Can't say I've ever used `fileselect`  or the preview feature.

But without the change it's very easy to break the state of the tape module. All you need to do is call `open` without first calling `stop` and waiting for the fadeout to finish and file handle to close.

Even with the change, I can imagine it being frustrating to design an interface when you can't know the state of the player directly and - don't know if a given call to `open` succeeds. For example if you trigger playback immediately on selecting a scrolling list entry. If youand on the entry you want and it doesn't play (still busy) then you try and scroll away and it plays an adjacent entry and _still_ doesn't let you select the one you're after.

But timeouts just implement the same behavior with much less precision, so at least I don't see how this is worse

PR #1781, opened 2024-05-13
improve binary-style params

dndrks:
while building docs for `binary`-style parameters, i came across a few API discontinuities, which this PR attempts to address.

- **E3 control**: as discussed in https://github.com/monome/norns/discussions/1752, E3 movement now only affects `"toggle"` behavior. it is ignored for `"momentary"` and `"trigger"`.
- **`params:set` now affects UI:**  previously, the `triggered` and `on` states for this param type did not update with `params:set`.
- **execute `"trigger"` without a value**: since triggers are _actions_, they can alternatively be executed with `params:set('example_trigger')` (this now matches the standard `trigger`-style parameter)

please  let me know if there's any additional questions or angles i missed! <3

### DIFF
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -31,6 +31,7 @@ _menu.errormsg = ""
 _menu.shownav = false
 _menu.showstats = false
 _menu.previewfile = ""
+_menu.binarystates = {triggered = {}, on = {}}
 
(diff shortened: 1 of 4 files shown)


tehn (approved):

tehn:
looks good!

PR #1785, opened 2024-05-30
binary triggers: allow default of '1'

dndrks:
addresses #1784 
@dansimco , I pushed these changes to my local build two weeks ago and _totally_ spaced on PR'ing this to `main`. apologies for the oversight and thank you so so much for reporting the issue!

PR #1787, opened 2024-06-19
fix for paramset:read() not setting values

dstroud:
Hi y'all, this fixes an issue introduced in #1785 which prevents paramset:read() from setting non-binary param values on .pset load. I'm not particularly confident poking around core files, but it seems to do the trick in my testing!


dndrks (approved):
ah, duh, thank you so so so much @dstroud !!

Issue #1436, opened 2021-09-19
TAPE .wav header corruption


catfact:
a couple people have encountered .wav files from TAPE that are missing data-chunk-size field in the .wav header. (byte positions 0x28 - 0x31).

such files will work fine in many programs, but not in a few programs which actually care about this  (notably apple quicktime and Apple Music.)

the most obvious explanation is that norns is shut down in the middle of recording. (same thing happens when my Zoom runs outta batteries.) this is hard to confirm after the fact but i think it's reasonable to assume this is the usual cause. (folllowed by having a full disk, maybe)

`sox` provides a quick fix:
```
sox --ignore-length broke.wav fixed.wav
```

(side question, are we including `sox` on norns base image? feels like a good idea.)

----

i guess i'm raising this for visibility but not sure how to address it. we could do stuff like check for running TAPE and end recording in crone shutdown /signal handlers. but there is no 100% reliable solution for this from inside a system that may lose power at any moment, and a partially-effective solution will still be a good amount of work. (would rather put work towards addressing the general shabbiness of TAPE module, etc.)

maybe just needs a note in documentation with suggested remedy.

tehn:
i spy a TODO: https://github.com/monome/norns/blob/main/lua/core/menu/sleep.lua#L10

i'll make sure `sox` gets on the next image

catfact:
user claims tape recording was stopped as normal, so this remains a mystery / something to watch for i guess. (presently attempting to get checksums to eliminate corruption during transfer as a proximate cause)

tehn:
close issue provided
- fix mentioned TODO
- create issue for sox w/ lua api
- add doc for fix

tehn:
tape stop is async so this would need a callback to wait for the state to be known.

PR #1790, opened 2024-07-10
TAPE: stop recording on SLEEP

dndrks:
addresses [this TODO](https://github.com/monome/norns/blob/main/lua/core/menu/sleep.lua#L10) in service of closing https://github.com/monome/norns/issues/1436 (and just popping in a sox w/Lua API ticket [per this comment](https://github.com/monome/norns/issues/1436#issuecomment-936151213))

### DIFF
--- a/lua/core/menu/sleep.lua
+++ b/lua/core/menu/sleep.lua
@@ -5,9 +5,10 @@ m.key = function(n,z)
     _menu.set_page("HOME")
   elseif n==3 and z==1 then
     m.sleep = true
+    if _menu.m.TAPE.rec.sel == 3 then
+      audio.tape_record_stop()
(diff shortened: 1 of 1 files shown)


tehn (approved):

catfact:
Sorry @tehn @dndrks but I dont think this is sufficient, since stopping tape and closing the file is doubly asynchronous - first the OSC request is picked up by crone, then the tape module handles it on the next audio block boundary. So not only will the file-close operation not be finished when the Lua function call exits, it won't have begun.

This would need a mechanism for the main thread to block until tape is finished with the file handle.

tehn:
@catfact oh thank you for pointing this out

PR #1794, opened 2024-08-07
Better handle k1 short press

skibu:
Found that there were couple issues with the handling of k1 short presses.

1. Somewhat hard to understand, so added comments
2. When in menu mode one had to do a short press, under 0.25 sec, to get back to script. But sometimes would press down bit too long and then nothing would happen, which was really confusing. So changed so that when in menu mode that a longer key press would get the user back to the script.
3. Wanted ability to create a script where user can hit k1 and then go directly to the scripts PARAMS Edit page. This is really nice for apps where user will want to change params. Won't have to do the kind of magical sequence of a short key1 press to get to Menus, then turn encoder1 to get to PARAMS menu, then turn encoder2 to select Edit>, and then click on key3 to finally get to the script's param page. That is way too difficult. But to give my script the ability to jump straight to the parameters page I needed a way of handling short key1 presses myself.
4. Took out unreachable code by the "elseif z == 0 then" because z couldn't ever be 0 there.

Also, completely unrelated, improved a comment in paramset.lua

### DIFF
--- a/lua/core/menu.lua
+++ b/lua/core/menu.lua
@@ -125,33 +125,55 @@ _menu.enc = function(n, delta)
 end
 
 
+-- Handles all button presses both for the script and for the menus.
+-- It is how short presses on key1 are detected and handled.
 _norns.key = function(n, z)
   -- key 1 detect for short press
   if n == 1 then
     if z == 1 then
+      -- key 1 pressed so start timer
       _menu.alt = true
       pending = true
       t:start()
-    elseif z == 0 and pending == true then
+    elseif pending == true then
+      -- Key 1 released within the timer's allowed time so was short press.
       _menu.alt = false
+
+      -- Toggle menu mode. If was in menu mode will go to application mode,
+      -- and visa versa.
       if _menu.mode == true and _menu.locked == false then
         _menu.set_mode(false)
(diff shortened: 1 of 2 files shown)


tehn:
hello! thanks for thinking out these issues.

re #2: the param system uses K1 as an alt key for changing acceleration of encoders, so having a long press switch modes would interfere with this existing functionality.

re #3: this prompted a good discussion--- we should make a small API (really a single function) to jump the menu screen to the PARAMs page at a specific parameter. i can see this being useful for script execution. we'll add this in a separate PR soon.

i'm going to close this, but we can discuss further.

skibu:
I finally had a chance to look into this more. From reading the code and trying it out, it appears that k3 is the alt key for fine encoder adjustments, not k1. So perhaps your point on # 2 is not truly an issue.

But to move things forward I thought I would create a library where my proposed changes can be tried out. This includes providing the ability for key1 press, long or short, can be made to directly go to the params edit page and select the first settable option. And fine encoder adjustments still work. Nice thing about the library is that script writers can try out the functionality easily, before the source code is changed. I'm pretty close to getting this small library ready and will let people know so they can try it out if they would like.

Issue #1747, opened 2023-12-11
Support `screen.draw_to()` with separate Cairo thread


Dewb:
Looks like the guts of `_image_context_focus` and `_image_context_defocus` need to become Cairo thread events.

PR #1795, opened 2024-08-08
Fixed issue with image buffers not being displayed properly

skibu:
Video is worth a thousand words so here is what it is supposed to look like:  https://youtu.be/81JUmu6EaWg

And here is what it looks like using an image buffer. Quite disappointing. https://youtu.be/W6i9Ffs2YzA

Please provide feedback if you have additional suggestions.

I tried to implement the dynamic scrolling of some pretty pictures in my Norns script that is under development. In order to do some interesting things graphically I tried reading in a PNG file into an image buffer, and then just writing the image buffer to the screen using screen.display_image(). The images showed up, but maybe a quarter of the time I just got a blank screen or just the lower part of the image. It was quite disappointing and not very pretty at all.

I did lots of exploring and testing and found that there is a timing issue. Screen commands are not executed fully immediately. Instead, they are queued and then processed by Cairo. It means that after doing a screen.clear() and then immediately afterwards a screen.display_image(png_buffer, x, y) that the clearing of the screen hadn't yet been completed. This resulted sometimes in the image buffer being overwritten by the clearing of the screen that was still in process. Likely, this is also occurring for other graphics operations, but just not yet noticed by anyone.

The solution was to change weaver.c _screen_clear() so that it doesn't return until the clear has been processed. I found that putting in a delay after the screen_clear() fixed the problem, but a delay is of course not a desirable solution. Since don't have good access to the workings of the screen queue, nor to Cairo, I found that needed to use a screen command that waits for results. This of course makes sure that the previous clear event is done. I chose the simplest screen function that returns a value, which is _screen_current_point(). 

The difference is like night and day. With the simple change to weaver.c the image buffer is always drawn correctly.

I created two scripts to illustrate the problem and the solution. Scripts are at https://github.com/skibu/nornsFun and are [scrollImageFile.lua](https://github.com/skibu/nornsFun/blob/main/scrollImageFile.lua) and [scrollImageBuffer.lua](https://github.com/skibu/nornsFun/blob/main/scrollImageBuffer.lua) . 

scrollImageFile.lua shows what it is supposed to look like and doesn't actually use an image buffer. But scrollImageBuffer.lua will show the problem. And you can try the simple change to weaver.c to see how scrollImageBuffer.lua works with the change.

And lastly, I made screen.c consistent and added `surface_may_have_color = true;` to screen_surface_display(). I think it was previously just an oversight from when other screen changes were made a few months ago.


dndrks:
i'm not confident enough to speak on the screen redraw backend, but wanted to add that @skibu 's changes test well for me -- seems like a very sensible solution!

i _was_ able to get `scrollImageBuffer.lua` drawing appropriately without those changes, through @skibu's [suggestion](https://github.com/skibu/nornsFun/blob/ae92ff8a07b5fa89efcd7d386da154b6daaa1338/scrollImageBuffer.lua#L29) of just calling `screen.current_point()` after the `screen.clear()` in the Lua layer -- @tlubke, do you remember if the necessity of the `screen.current_point()` call is by design? if not, then it seems like the proposed changes are a good fix!

tlubke:
The solution might be to give all screen functions in weaver an event so that screen operations happen sequentially. Specifically `screen_load_png()`, `screen_display_image()`, and `screen_display_image_region()` for this example. I'll try that out quick.

tlubke:
[2024-08-08_norns_pr_1795_d.patch](https://github.com/user-attachments/files/16553565/2024-08-08_norns_pr_1795_d.patch)

Here's a patch that adds two more screen events, one for `display_surface()` ("display_image" in Lua) and the other for `display_surface_region()`. This seems to fix https://github.com/skibu/nornsFun/blob/main/scrollImageBuffer.lua for me.

skibu:
I’ll try out @tlubke ’s patch in a bit. I’m pretty confident that adding the new screen events is the proper solution. I can see how making functions like `screen_display_image()` event driven because the event queue would sequentially process the events as desired. Thanks for coming up with a solid solution!

Btw, I had also tried just using screen_results_wait(). It deadlocking meant that I needed a command that retrieved data in order to sync the commands.

matron/src/hardware/screen.c
```diff
@@ -636,6 +636,9 @@ bool screen_surface_get_extents(screen_surface_t *s, screen_surface_extents_t *e
 
 void screen_surface_display(screen_surface_t *s, double x, double y) {
     cairo_surface_t *image = (cairo_surface_t *)s;
+
+    surface_may_have_color = true;
```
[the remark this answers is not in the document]
skibu: I think would be great to do this in a separate PR.

skibu:
I tested out @tlubke 's change and took out my original change. It works great! Therefore I checked in @tlubke 's changes into the branch for everyone to easily see.

I think that this PR can now be re-reviewed by folks and accepted if others agree.

catfact:
Thanks awesome


Yes, queing everything is better than blocking the caller.

catfact:
Thanks awesome


Yes, queing everything is better than blocking the caller.

tlubke:
Probably a good time to cross off #1747 too.  I got started on it and modified scrollImageBuffer.lua to use `screen.draw_to()` on `png_buffer`. I occasionally see a flicker on the scrolling image, so not sure if I implemented everything correctly, or if I went too far with how many functions I made into events.

I'll push a branch and see what people think. https://github.com/tlubke/norns/commit/a995249f0f5b0b620a6204f79855112123b9f3ee

catfact:
@tlubke 

so, i really haven't been able to get dug in to norns code for a while, i think you are the expert on the screen modules at this point. you now have write access; feel free to approve and merge if you see fit.

tlubke:
Sounds good, I'll go over these changes again soon and see if I can remember where my train of thought was on my last edit.

If there is an upcoming norns version release, it probably makes sense to get this fix merged in there.

matron/src/screen_events.c
```diff
@@ -191,6 +191,14 @@ void handle_screen_event(struct screen_event_data *ev) {
     case SCREEN_EVENT_DISPLAY_PNG:
         screen_display_png(ev->buf, ev->payload.bd.d1, ev->payload.bd.d2);
         break;
+    case SCREEN_EVENT_DISPLAY_SURFACE:
+        screen_surface_display(ev->buf, ev->payload.bd.d1, ev->payload.bd.d2);
+        ev->buf = NULL; // Didn't make a copy of the pointer, nullify to avoid double free.
+        break;
+    case SCREEN_EVENT_DISPLAY_SURFACE_REGION:
+        screen_surface_display_region(ev->buf, ev->payload.d.d1, ev->payload.d.d2, ev->payload.d.d3, ev->payload.d.d4, ev->payload.d.d5, ev->payload.d.d6);
```
tlubke, line 199: minor: [coding style (c)](https://github.com/monome/norns/wiki/coding-style-(c)#norns), column length.

matron/src/screen_events.c
```diff
@@ -496,6 +504,30 @@ void screen_event_display_png(const char *s, double x, double y) {
...
+    struct screen_event_data ev;
+    screen_event_data_init(&ev);
+    ev.type = SCREEN_EVENT_DISPLAY_SURFACE;
+    ev.buf = surface;
+    ev.payload.bd.d1 = x;
+    ev.payload.bd.d2 = y;
+    screen_event_data_push(&ev);
+}
+
+void screen_event_display_surface_region(void* surface, double left, double top, double width, double height, double x, double y) {
```
tlubke, line 517: minor: [coding style (c)](https://github.com/monome/norns/wiki/coding-style-(c)#norns), column length + '*' alignment.

matron/src/screen_events.h
```diff
@@ -33,6 +33,8 @@ extern void screen_event_close_path(void);
 extern void screen_event_text_extents(const char *s);
 extern void screen_event_export_png(const char *s);
 extern void screen_event_display_png(const char *filename, double x, double y);
+extern void screen_event_display_surface(void* surface, double x, double y);
+extern void screen_event_display_surface_region(void* surface, double left, double top, double width, double height, double x, double y);
```
tlubke, line 37: minor: [coding style (c)](https://github.com/monome/norns/wiki/coding-style-(c)#norns), column length + '*' alignment.

tlubke (approved):
I spent a little while tracing back through this code and some of that paths it effects. I think this is good to merge on a functional basis.

I realize these are mostly my changes and I had forgotten about the coding style in the wiki here. Looking at some more recent commits in matron though, they don't seem to bet strictly adhered to.

@skibu I'll leave it up to you if you want to adjust the formatting to match the wiki.

I've approved the PR, and will merge it in a week or two if there is no opposition from other maintainers on formatting.

tehn:
for what it's worth, the entire repo might benefit from having a clang-format pass (there is a rule, i think it should match the code style)

skibu:
Since I was asked, I don't see formatting issues as critical since they are currently only loosely adhered to. If folks want formatting to be consistent I think it is far better to enforce it automatically every time a file is checked into github. 

Therefore I agree that would be good to go ahead and merge.

I don't think an official release should be done at this time though. After this PR I found another screen function that needs to be a screen event in order to prevent problems. Plus I found some other problematic bugs as well. I'll try to create issues for these soon so that they an be dealt with.

skibu:
I finally was able to create a separate issue for two other functions that also need to be queue functions. The issue is #1808 . This PR should be merged though instead of waiting for the new issue to also be resolved.

tlubke:
@skibu Thanks for opening those other issues. I'm going to use #1615 to track/group progress as a whole for all of the remaining screen fixes.

I think you're correct about merging this one instead of waiting.

PR #1799, opened 2024-10-06
Lua API doc improvements

niksilver:
* Allow norns lua docs to use markdown.
* Link Lua API documentation back to [script references](https://monome.org/docs/norns/reference/#modules). For example, the API doc for the grid module now links back to https://monome.org/docs/norns/reference/grid, etc.
* Remove ldoc warnings. It runs cleanly now.
* Tidy up config.ld.
* Correct module names in Lua API index.
* Correct/Create @alias tags where needed.
* List in docs for Screen.font_face() displays as a proper list.
* List in docs fpr lib.lattice.new_socket() displays as a proper list.
* Add link in lib.intonation.
* Update readme-setup.md to explain how to install the docs, and that it should run without errors or warnings.

Generally this is an attempt to tidy up the Lua API docs. You should see no code changes here at all. Only comments have changed and `config.ld` has been updated.

An explanation about the "Correct module names" and "@alias" bullet points above. In the current [API index page](https://monome.org/docs/norns/api/index.html) some of the module names display incorrectly. For example we see `lib.BeatClock` instead of `lib.beatclock`, and `Lattice` instead of `lib.lattice`. So these have now been corrected by applying `@module` and `@alias` tags correctly.

I notice that `config.ld` is set to process SuperCollider classes. I'm not aware that it runs over any SuperCollider classes, so those lines could be removed. But I've not removed them because I fear I'm missing something. Please do let me know if I've missed something here.

I realise these changes will not actually change [the API docs on the website](https://monome.org/docs/norns/api/index.html). That's because this needs to be [copied manually into another repo](https://llllllll.co/t/norns-development/14073/558). I will do that as a separate piece of work - perhaps with some further improvements there. However, I want to make sure these changes are accepted first, before I think about copying them over.


tehn:
looks great! will do a full review very soon. thank you.

tehn (approved):
Thank you, these are all great improvements!

I can copy the html over to the main site.

PR #1800, opened 2024-10-07
Add return values for Audio.file_info

jbonhag:
This PR documents the return values for `Audio.file_info`. These return values were validated against a known tape file:

```
> ch, samples, rate = audio.file_info("/home/we/dust/audio/tape/0001.wav")
> ch
2
> samples
72782080
> rate
48000
```

I wasn't sure whether I was supposed to regenerate the output html or not, so happy to remove that if it should be regenerated some other way.

### DIFF
--- a/doc/modules/audio.html
+++ b/doc/modules/audio.html
@@ -969,6 +969,18 @@
         </li>
     </ul>
 
+    <h3>Returns:</h3>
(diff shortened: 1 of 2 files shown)


catfact (approved):
thanks!

PR #1247, opened 2020-11-24
Cairo blending modes

tlubke:
motivation came from a conversation at the weekly discord meetup.

someone brought up that the *cairo* C API had support for different blending modes/operators as shown here: https://www.cairographics.org/operators/

I wrote a short norns script for testing all of the operators
```
local a = {}
a[0]  = "CLEAR"
a[1]  = "SOURCE"
a[2]  = "OVER"
a[3]  = "IN"
a[4]  = "OUT"
a[5]  = "ATOP"
a[6]  = "DEST"
a[7]  = "DEST_OVER"
a[8]  = "DEST_IN"
a[9]  = "DEST_ATOP"
a[10] = "XOR"
a[11] = "ADD"
a[12] = "SATURATE"
a[13] = "MULTIPLY"
a[14] = "SCREEN"
a[15] = "OVERLAY"
a[16] = "DARKEN"
a[17] = "LIGHTEN"
a[18] = "COLOR_DODGE"
a[19] = "COLOR_BURN"
a[20] = "HARD_LIGHT"
a[21] = "SOFT_LIGHT"
a[22] = "DIFFERENCE"
a[23] = "EXCLUSION"

local i = 0;

function init()
  redraw()
end

function enc(n, d)
  i = util.clamp(i+d, 0, 23)
  redraw()
end

function redraw()
  screen.clear()
  screen.level(15)
  screen.blend_mode(2) -- default
  screen.move(64, 8)
  screen.text_center(a[i])
  screen.level(8)
  screen.rect(32, 10, 64,32)
  screen.fill()
  screen.blend_mode(i)
  screen.level(4)
  screen.rect(48, 26, 64, 32)
  screen.fill()
  screen.update()
end
```

some of them are less friendly/useable than others, and I expect the list should be curated a bit.

the added C bits were kinda rushed, so I'm sure there's plenty to clean up.

PR #1809, opened 2025-01-27
Fixed ability to use all the drawing modes 

skibu:
Turns out that BLEND_MODES was missing `['SATURATE'] = 3` which was preventing one from using that draw mode and any of the draw modes after that one. Quite limiting! I'm certain this was just a typo.

Since added a BLEND_MODE also had to fix screen_set_operator().


tlubke:
(https://github.com/monome/norns/pull/1247#issuecomment-735302319) Based on this, I think the index limit and omission of the SATURATE mode were purposeful at the time.

Here's all of the reasoning that I can remember.

* **ADD** output is identical to **SATURATE** (I can't remember if this is because of being grayscale).
* **DEST_ATOP** should be identical to **OVER** (which is the default operator mode).
* **HSL_HUE**, **HSL_SATURATION**, **HSL_COLOR**, and **HSL_LUMINOSITY** don't provide much value in grayscale.

---

Since `screen.set_operator()` can take either a `string` or a `number`, changing the index order could break existing scripts that might use it with an index directly.

---

Having said all that, the range-check change in `screen.c` looks correct, and it may be worth considering removing the limit for the purpose of interacting with non-grayscale colorspaces through mods.

skibu:
Ah, SATURATE duplicating the functionality of ADD explains why it was removed in screen.lua . But there is still a definite bug that causes the operators beyond ADD to not work correctly because the indexes are off by one (even if one uses the name of the operator instead of the number).  I tried both using the index and the name of the operator and I got incredibly confusing results since the wrong operator was being used.

**The root of the problem is a mismatch between the operator list in screen.lua and in screen.c .** 

In screen.c the list of operators is:
```
static cairo_operator_t ops[NUM_OPS] = {
    CAIRO_OPERATOR_OVER,
    CAIRO_OPERATOR_XOR,
    CAIRO_OPERATOR_ADD,
    CAIRO_OPERATOR_SATURATE,
    CAIRO_OPERATOR_MULTIPLY,
    CAIRO_OPERATOR_SCREEN,
   ...
```

If SATURATE should indeed not exist **(will need you to confirm this)** then proper solution would be to also update the operator list in screen.c so it matches the list in screen.lua .

tlubke:
Ah, I see the issue now. Yes, I think the indices in `screen.lua` are the intended ones, and `screen.c` was not updated appropriately. It doesn't look this was a regression either (I'm surprised it took nearly 5 years for the issue to be detected!).

tlubke:
Was doing some testing with this commit and noticed **saturate** is actually in the ops list twice. Could you move `CAIRO_OPERATOR_SATURATE` to index-24 in `screen.c` to match the table in `screen.lua`? That way index limits won't need to be updated. Though the documentation fix in #1810 can be kept and changed to 28.

tlubke:
I ran ripgrep on all the scripts published in the norns community repo. I counted 16 scripts that use `screen.blend_mode()` with an argument that would be broken by this fix.

I know in the past there may have been PRs made to community scripts that are made along any updates with breaking changes. I don't however know if there is a threshold on number of effected scripts that influence the approach to take with fix. For instance the `screen.blend_mode()` documentation could be changed to reflect the actual indices, but I wouldn't want to make anyone feel like they are being gaslit by changing docs.

@monome/core 
Any input on which direction to take given this number of community scripts?

tehn:
16 isn't too many, but the more important question might be how many authors are there for those 16 scripts, and how active are they (as in, would a PR be seen).

I think fixing the docs and the code would be sensible. Won't any "breakage" be a rendering problem, rather than a complete fail?

tlubke:
@tehn 
I ran the query again. I guess I had originally double-counted some lua files that were part of the same script. I think there's a good chance most of the potential PRs would be seen.

## Active (9 of 14)
* infinitedigits
  * Acrostic
  * amenbreak
  * graintopia
  * l_ll__l_
  * zxcbvn
* imminent_gloom
  * støy
* justmat
  * d
  * larc
* sixolet
  * phenotype

## Inactive (5 of 14)
* ambalek (⌛⌛)
  * raindrops
* mendrzec (⌛⌛⌛)
  * impact	 
* Quixotic7 (⌛⌛⌛)
  * groovecats
* klindley (⌛⌛⌛)
  * demoncore
* skibu (⌛)
  * taweet

---

> I think fixing the docs and the code would be sensible. Won't any "breakage" be a rendering problem, rather than a complete fail?

That's right, and I think since XOR comes at an index before the change I don't think any of the above scripts would become illegible.

tehn:
Thank you the additional information. My opinion is that this change should be fine. We can PR the affected projects or at least alert the authors.

tlubke (approved):

Issue #435, opened 2018-06-25
investigate sc-supernova


catfact:
its worth putting a little time into checking the current state of sc-supernova on ARM.

available information and issues online seem a little stale.

simonvanderveldt:
Finally managed to finish sc 3.9.1 compilation on norns, apparently some combination of CMake flags causes compilation to fail :|
Managed to compile it using the following:
```
$ cmake -L -DCMAKE_BUILD_TYPE="Release" -DBUILD_TESTING=OFF -DSUPERNOVA=ON -DNATIVE=ON -DSC_WII=OFF -DSC_IDE=OFF -DSC_QT=ON -DSC_ED=OFF -DSC_EL=OFF -DSC_VIM=OFF ..
$ make
```
At least starting supernova works fine:
```
we@norns:~/supercollider/build/server/supernova $ ./supernova 
Supernova booting
samplerate mismatch between command line argument and jack
forcing samplerate of 48000Hz
Supernova ready
```
Next up: Try to get it to replace scsynth and run some scripts

simonvanderveldt:
So I had to recompile with `-DSC_QT=OFF` otherwise `sclang` failed to start because of some missing Qt dependency.

Also had to do a `make install` otherwise `sclang` would complain about missing library items, haven't been able to find a way to run it in a "development" setup using the dependencies from within the project/directory.

Anyway, the basics seem to work:
```
we@norns:~/supercollider/build $ ./lang/sclang 
compiling class library...
	Found 594 primitives.
	Compiling directory '/usr/local/share/SuperCollider/SCClassLibrary'
	Compiling directory '/home/we/.local/share/SuperCollider/Extensions'
ERROR: Class extension for nonexistent class 'HistoryGui'
     In file:'deprecated/3.9/HistoryGui.sc'
	numentries = 665683 / 7120940 = 0.093
	3845 method selectors, 1852 classes
	method table size 4556556 bytes, big table size 28483760
	Number of Symbols 9118
	Byte Code Size 263984
	compiled 228 files in 0.51 seconds

Info: 4 methods are currently overwritten by extensions. To see which, execute:
MethodOverride.printAll

compile done
localhost : setting clientID to 0.
internal : setting clientID to 0.
Class tree inited in 0.04 seconds


*** Welcome to SuperCollider 3.9.3. *** For help type ctrl-c ctrl-h (Emacs) or :SChelp (vim) or ctrl-U (sced/gedit).
sc3> Server.supernova
-> Server
[...]

PR #1815, opened 2025-02-07
Exclude directories from SELECT menu

niksilver:
This is bug fix and an enhancement.
1. It would be good to put test scripts in their own folder, out of the way of the main code, and to not have them show up in the SELECT menu. This change allows that by excluding a `test` folder from the SELECT menu.
2. The documentation at https://monome.org/docs/norns/reference/#folder-structure says anything in a `docs` folder is excluded from the SELECT menu. That's not true (if it's a Lua script). This change also ensures that exclusion happens.

This change is documented in `monome/docs` PR [#435](https://github.com/monome/docs/pull/435).


tehn (approved):

tehn:
this is a good idea, thanks!

PR #1818, opened 2025-02-23
Generic USB serial device

### LINKED ISSUE #1816: Playdate integration
What do folks think about adding [Playdate](https://play.date) integration to Norns? I have a feature complete implementation running on my Norns, but I'd like to get some feedback on the big picture before I ask anyone to look at the code.

## Motivation
From the perspective of Norns, Playdate is basically a retro game controller with two notable exceptions: it has a crank and it runs Lua. I'm primarily interested in using the crank to interact with Norns and Crow. There are lots of possibilities in the other direction too, such as controlling Playdate music apps with a midi keyboard, a grid or control voltage signals.

## Implementation
The Playdate supports serial connections over USB. Official documentation for the serial connection can be found in the [Playdate Lua serial API docs](https://sdk.play.date/2.6.2/Inside%20Playdate.html#M-wired-networking) and [Playdate C serial API docs](https://sdk.play.date/2.6.2/Inside%20Playdate%20with%20C.html#f-system.setSerialMessageCallback). There is also unofficial documentation of the USB serial capabilities by the [Playdate Reverse Engineering project](https://github.com/cranksters/playdate-reverse-engineering/blob/main/usb/usb.md).

The serial connection can be used to enable a controller mode that streams the button, crank and accelerometer state over serial. It does not (to my knowledge) support the HID protocol, so it could not be used with the Norns `gamecontroller` API without modifications.

### Proposed API
<details>

#### Methods
`playdate.connected()`: true if Playdate is connected
- The Playdate serial connection is only available if the Playdate is connected to USB and unlocked.

`playdate.send( line )`: send a string message to Playdate using the `msg` serial command

`playdate.run( path )`: launch the PDX at the given path on Playdate
- This could be convenient if there is a custom PDX associated with the Norns script. The user would be able to start the Norns script and automatically have the Playdate in the expected state

`playdate.controller_start()`: start controller mode
`playdate.controller_stop()`: stop controller mode
- These methods allow Norns scripts to react to Playdate hardware state changes by overriding certain callback functions.
- While controller mode is enabled, the Playdate screen stops updating and no longer responds to button presses except for the lock button. The lock button will stop controller mode on the Playdate side. There is no indication over the serial connection that controller mode has stopped other than no longer receiving updates.
- Controller mode might ultimately be a shortcut to getting date from Playdate into Norns without having to write a custom PDX. Users could also replicate or extend the behavior from controller mode in their custom PDX using `print()` in Lua or `playdate->system->logToConsole()` in C with the expected message format.

#### Device Management Callbacks
`function playdate.add( id, name, dev )`
`function playdate.remove( id )`
- These could be used to reactivate controller mode if the device auto-locks and is later unlocked.

#### Controller Event Callbacks
`function playdate.accel( x, y, z )`
- received once per frame with the current accelerometer vector
- x, y, and z are floats between -1.0 and 1.0

`function playdate.button( b, s )`
- received on the frame that any button except the lock button is depressed or released
- b is one of `A`, `B`, `U`, `D`, `L`, `R` or `M`, and s is 1 if depressed and 0 if released

`function playdate.crank( d )`
- received each frame that the crank position is updated
- d is the current angle of the crank in degrees (0.0 to 360.0)

`function playdate.crankdock( s )`
- received on the frame that the crank is docked or undocked
- s is 1 if docked and 0 if undocked

#### Serial Event Callbacks
`function playdate.event( line )`
- called for all non-controller-mode serial messages
- line is a string sent from the PDX using print or logToConsole
- used to implement custom serial connection between a Playdate app and a Norns script

## Other interesting serial APIs:
[`buttons` ](https://github.com/cranksters/playdate-reverse-engineering/blob/main/usb/usb.md#buttons)
- This enables the button testing mode of the playdate. It could be used to implement Norns event callbacks with slightly different behavior from the controller mode implementation:
	- It sends the current state of all buttons and the crank every frame
	- It does not send accelerometer data
	- It sends the state of the lock button and prevents the device from being locked
- This might some things easier like measuring how the crank position changes over time, but I don't think it's a net positive over using controller mode or implementing the behavior you need in a custom PDX.

[`bitmap`](https://github.com/cranksters/playdate-reverse-engineering/blob/main/usb/usb.md#bitmap)
- Send a 400x240 1-bit bitmap to the Playdate's screen
- I could see this being nice when controller mode is enabled since the Playdate screen is non-interactive at this point, but probably not worth it.

[`screen`](https://github.com/cranksters/playdate-reverse-engineering/blob/main/usb/usb.md#screen)
- Take a screenshot of the Playdate
- Might be nice for documentation, debugging or issue reporting

[`eval`](https://github.com/cranksters/playdate-reverse-engineering/blob/main/usb/usb.md#eval)
- Send a compiled Lua function to playdate and execute it
- Possibly a shortcut to implementing serial message handling on the Playdate side, but most apps would probably be better off responding to `msg` serial events.
- Could also be a way to interact with apps that don't explicitly support serial communication, but that may require reverse engineering the PDX.

[`stream`](https://github.com/cranksters/playdate-reverse-engineering/blob/main/usb/stream.md)
- This continually streams the playdates screen, audio and buttons over the serial connection. It is used to implement the [Mirror](https://play.date/mirror/) app which mirrors the Playdate on a connected computer.
- The audio stream might be the most interesting thing here, but you could also just connect the Playdate headphone output to the Norns audio input. 

`btn` / `changecrank` / `dockcrank`  / `accel`
- Simulate button presses on Playdate
- Could be an interesting way to let Norns interact with Playdate, but in most cases users would probably want to implement the serial message handlers.

`autolock`
- Change the autolock setting on Playdate
- It might be desirable to change Playdate to only autolock on battery when connected to Norns. I could also see users preferring to set this themselves and not wanting Norns script mess with this setting on their device.
</details>

## Open Questions / Other Considerations
1. The implementation in my Norns fork is mostly a copy of the crow implementation with some additional C APIs exposed to Lua. While I was writing it I was wondering if it might be worth having a more generic serial device interface that can read/write messages and defer to Lua for more device-specific logic. That could eliminate the need to implement this sort of integration in the Norns codebase and instead move it to a user script or library. This might not be possible without changes to device monitoring and initialization though.
2. The `playdate.event( line )` callback would be called for all messages logged from the Playdate. I debated if it should only be called for messages that use some sort of prefix like `~msg: ` , similar to how the controller protocol works (`~ctl: `). Ultimately I'm leaning towards leaving this up to the script/PDX author. 
3. There are official APIs supporting the serial connection, so that feels fairly stable to me. As for the serial commands other than `msg`, these are not officially documented outside of using the `help` command. I'm not sure if commands like `controller start` and `controller stop` can be considered stable.

Thanks for considering my proposal!

midouest:
resolves #1816

Adds a new device type, `DEV_TYPE_SERIAL`, which can be configured from Lua, as discussed in #1816. FYI I am far from being a C/Linux/serial expert.

An example for the Playdate by Panic can be found here: [midouest/playdate-norns](https://github.com/midouest/playdate-norns/blob/main/lib/mod.lua)

Design
---

Added a new `system_pre_device_scan` hook that gets called after startup but before the device scan and event loop. This is intended to allow mods to configure new device handlers before the devices are scanned. I didn't want to mess around with re-scanning devices after new handlers are added or allowing handlers to be removed, but that would be another possible design.

When an unmatched TTY is found, the device path, name, vendor and model are passed to weaver which checks each serial device handler for a matching vendor/model. This would allow a single handler to match multiple device models. If the handler matches the device attributes, another configuration callback gets called to configure the serial connection.

### DIFF
--- a/lua/core/hook.lua
+++ b/lua/core/hook.lua
@@ -36,7 +36,6 @@ local hooks = {
   script_pre_init = Hook.new(),
   script_post_init = Hook.new(),
   script_post_cleanup = Hook.new(),
-  system_pre_device_scan = Hook.new(),
   system_post_startup = Hook.new(),
   system_pre_shutdown = Hook.new(),
 }
--- a/lua/core/hook.lua
+++ b/lua/core/hook.lua
@@ -36,7 +36,7 @@ local hooks = {
   script_pre_init = Hook.new(),
   script_post_init = Hook.new(),
   script_post_cleanup = Hook.new(),
-  system_startup = Hook.new(),
+  system_pre_device_scan = Hook.new(),
   system_post_startup = Hook.new(),
   system_pre_shutdown = Hook.new(),
 }
--- a/lua/core/hook.lua
+++ b/lua/core/hook.lua
@@ -36,6 +36,7 @@ local hooks = {
   script_pre_init = Hook.new(),
   script_post_init = Hook.new(),
   script_post_cleanup = Hook.new(),
+  system_startup = Hook.new(),
   system_post_startup = Hook.new(),
   system_pre_shutdown = Hook.new(),
 }
--- a/lua/core/serial.lua
+++ b/lua/core/serial.lua
@@ -120,7 +120,7 @@ end
 -- @field VSTOP STOP character
 -- @field VSUSP SUSP character
 -- @field VEOL EOL character
-serial.cc = tab.readonly(table = {
+serial.cc = tab.readonly{table = {
   VINTR = 0,
   VQUIT = 1,
   VERASE = 2,
@@ -138,7 +138,7 @@ serial.cc = tab.readonly(table = {
   VWERASE = 14,
   VLNEXT = 15,
   VEOL2 = 16,
-})
+}}
 
 --- iflag constants
 -- @table iflag
@@ -157,7 +157,7 @@ serial.cc = tab.readonly(table = {
 -- @field IXOFF Enable start/stop input control.
 -- @field IMAXBEL Enable start/stop input control.
 -- @field IUTF8 Input is UTF8 (not in POSIX).
-serial.iflag = tab.readonly(table = {
+serial.iflag = tab.readonly{table = {
   IGNBRK = 1,
   BRKINT = 2,
   IGNPAR = 4,
@@ -173,7 +173,7 @@ serial.iflag = tab.readonly(table = {
   IXOFF = 4096,
   IMAXBEL = 8192,
   IUTF8 = 16384,
-})
+}}
 
 --- oflag constants
 -- @table oflag
@@ -188,7 +188,7 @@ serial.iflag = tab.readonly(table = {
 -- @field VTDLY Select vertical-tab delays:
 -- @field VT0 Vertical-tab delay type 0.
 -- @field VT1 Vertical-tab delay type 1.
-serial.oflag = tab.readonly(table = {
+serial.oflag = tab.readonly{table = {
   OPOST = 1,
   OLCUC = 2,
   ONLCR = 4,
@@ -200,7 +200,7 @@ serial.oflag = tab.readonly(table = {
   VTDLY = 16384,
   VT0 = 0,
   VT1 = 16384,
-})
+}}
 
 
 --- cflag constants
@@ -216,7 +216,7 @@ serial.oflag = tab.readonly(table = {
 -- @field PARODD Odd parity, else even.
 -- @field HUPCL Hang up on last close.
 -- @field CLOCAL Ignore modem status lines.
-serial.cflag = tab.readonly(table = {
+serial.cflag = tab.readonly{table = {
   CSIZE = 48,
   CS5 = 0,
   CS6 = 16,
@@ -228,7 +228,7 @@ serial.cflag = tab.readonly(table = {
   PARODD = 512,
   HUPCL = 1024,
   CLOCAL = 2048,
-}
+}}
 
 --- lflag constants
 -- @table lflag
@@ -241,7 +241,7 @@ serial.cflag = tab.readonly(table = {
 -- @field NOFLSH Disable flush after interrupt or quit.
 -- @field TOSTOP Send SIGTTOU for background output.
 -- @field IEXTEN Enable implementation-defined input processing.
-serial.lflag = tab.readonly(table = {
+serial.lflag = tab.readonly{table = {
   ISIG = 1,
   ICANON = 2,
   ECHO = 8,
@@ -251,7 +251,7 @@ serial.lflag = tab.readonly(table = {
   NOFLSH = 128,
   TOSTOP = 256,
   IEXTEN = 32768,
-})
+}}
 
 --- speed constants
 -- @table speed
@@ -285,7 +285,7 @@ serial.lflag = tab.readonly(table = {
 -- @field B3000000 3000000 baud
 -- @field B3500000 3500000 baud
 -- @field B4000000 4000000 baud
-serial.speed = tab.readonly(table = {
+serial.speed = tab.readonly{table = {
   B50 = 1,
   B75 = 2,
   B110 = 3,
@@ -316,6 +316,6 @@ serial.speed = tab.readonly(table = {
   B3000000 = 4109,
   B3500000 = 4110,
   B4000000 = 4111,
-}
+}}
 
 return serial
--- a/lua/core/serial.lua
+++ b/lua/core/serial.lua
@@ -3,46 +3,51 @@
 local tab = require 'tabutil'
 
 local serial = {
-    _handlers={},
+  _handlers={},
 }
 
 --- add a new serial device handler
 function serial.handler(args)
-    assert(serial._handlers[args.id] == nil, "duplicate serial handler id: " .. args.id)
-    serial._handlers[args.id] = args
+  assert(serial._handlers[args.id] == nil, "duplicate serial handler id: " .. args.id)
+  serial._handlers[args.id] = args
 end
 
 --- send a message to a serial device
 function serial.send(dev, line)
-    _norns.serial_send(dev, line)
+  _norns.serial_send(dev, line)
 end
 
 _norns.serial = {}
 
 function _norns.serial.match(vendor, model, serial_num, interface_num)
-    local attrs = tab.readonly{table={vendor=vendor, model=model, serial=serial_num, interface=interface_num}}
-    for id, handler in pairs(serial._handlers) do
-        if handler.match(attrs) then
-            return id
-        end
+  local attrs = tab.readonly{table={
+    vendor=vendor,
+    model=model,
+    serial=serial_num,
+    interface=interface_num
+  }}
+  for id, handler in pairs(serial._handlers) do
+    if handler.match(attrs) then
+      return id
     end
-    return nil
+  end
+  return nil
 end
 
 function _norns.serial.configure(handler_id, tio)
-    return serial._handlers[handler_id].configure(tio)
+  return serial._handlers[handler_id].configure(tio)
 end
 
 function _norns.serial.add(handler_id, id, name, dev)
-    serial._handlers[handler_id].add(id, name, dev)
+  serial._handlers[handler_id].add(id, name, dev)
 end
 
 function _norns.serial.remove(handler_id, id)
-    serial._handlers[handler_id].remove(id)
+  serial._handlers[handler_id].remove(id)
 end
 
 function _norns.serial.event(handler_id, id, line)
-    serial._handlers[handler_id].event(id, line)
+  serial._handlers[handler_id].event(id, line)
 end
 
 --- cc constants
--- a/lua/core/serial.lua
+++ b/lua/core/serial.lua
@@ -1,5 +1,6 @@
 --- USB serial device interface
 -- @module serial
+local tab = require 'tabutil'
 
 local serial = {
     _handlers={},
@@ -19,8 +20,9 @@ end
 _norns.serial = {}
 
 function _norns.serial.match(vendor, model, serial_num, interface_num)
+    local attrs = tab.readonly{table={vendor=vendor, model=model, serial=serial_num, interface=interface_num}}
     for id, handler in pairs(serial._handlers) do
-        if handler.match(vendor, model, serial_num, interface_num) then
+        if handler.match(attrs) then
             return id
         end
     end
--- a/lua/core/serial.lua
+++ b/lua/core/serial.lua
@@ -7,12 +7,21 @@ local serial = {
 }
 
 --- add a new serial device handler
-function serial.handler(args)
+-- @param args
+-- @param args.id unique identifier for the handler
+-- @param args.match function(attrs) returning true if the handler should accept the device with the given attributes, otherwise false
+-- @param args.configure function(term) receiving the device's initial terminal settings and returning the new terminal settings. The callback should modify the input table using bitwise operations and then return the input table
+-- @param args.add function(id, name, dev) called when a new device for this handler has been connected and initialized
+-- @param args.remove function(id) called when a connected device for this  handler has been disconnected
+-- @param args.event function(id, line) called when a message is received from a connected device for this handler
+function serial.add_handler(args)
   assert(serial._handlers[args.id] == nil, "duplicate serial handler id: " .. args.id)
   serial._handlers[args.id] = args
 end
 
 --- send a message to a serial device
+-- @param dev opaque device pointer
+-- @param line string message to send
 function serial.send(dev, line)
   _norns.serial_send(dev, line)
 end
@@ -24,7 +33,7 @@ function _norns.serial.match(vendor, model, serial_num, interface_num)
     vendor=vendor,
     model=model,
     serial=serial_num,
-    interface=interface_num
+    interface=interface_num,
   }}
   for id, handler in pairs(serial._handlers) do
     if handler.match(attrs) then
--- a/lua/core/serial.lua
+++ b/lua/core/serial.lua
@@ -9,15 +9,20 @@ local serial = {
 --- add a new serial device handler
 -- @param args
 -- @param args.id unique identifier for the handler
--- @param args.match function(attrs) returning true if the handler should accept the device with the given attributes, otherwise false
--- @param args.configure function(term) receiving the device's initial terminal settings and returning the new terminal settings. The callback should modify the input table using bitwise operations and then return the input table
+-- @param args.match function(attrs) returning true if the handler should accept the device with the given attributes, 
+--   otherwise false
+-- @param args.configure function(term) receiving the device's initial terminal settings and returning the new terminal
+--   settings. The callback should modify the input table using bitwise operations and then return the input table
 -- @param args.add function(id, name, dev) called when a new device for this handler has been connected and initialized
 -- @param args.remove function(id) called when a connected device for this  handler has been disconnected
--- @param args.event function(id, line) called when a message is received from a connected device for this handler
+-- @param args.event function(id, data) called when a message is received from a connected device for this handler. The 
+--   data argument will be no more than 255 bytes in length. Messages longer than 255 bytes will be spread out over 
+--   multiple event callbacks. It's possible for multiple messages shorter than 255 bytes to appear in a single event 
+--   callback.
 -- @usage
 -- friends = {}
 --
--- handler = serial.add_handler({
+-- serial.add_handler({
 --   id = "example_id", -- can be number or string
 --
 --   match = function(attrs)
@@ -58,8 +63,8 @@ local serial = {
 --     tab.remove(friends, id)
 --   end,
 --
---   event = function(id, line)
---     print(friends[id] .. " says:", line)
+--   event = function(id, data)
+--     print(friends[id] .. " says:", data)
 --   end
 -- })
 function serial.add_handler(args)
@@ -69,9 +74,9 @@ end
 
 --- send a message to a serial device
 -- @param dev opaque device pointer
--- @param line string message to send
-function serial.send(dev, line)
-  _norns.serial_send(dev, line)
+-- @param data string message to send
+function serial.send(dev, data)
+  _norns.serial_send(dev, data)
 end
 
 _norns.serial = {}
@@ -103,8 +108,8 @@ function _norns.serial.remove(handler_id, id)
   serial._handlers[handler_id].remove(id)
 end
 
-function _norns.serial.event(handler_id, id, line)
-  serial._handlers[handler_id].event(id, line)
+function _norns.serial.event(handler_id, id, data)
+  serial._handlers[handler_id].event(id, data)
 end
 
 --- CC (special character) constants
--- a/lua/core/serial.lua
+++ b/lua/core/serial.lua
@@ -133,16 +133,10 @@ serial.cc = tab.readonly{table = {
   VEOF = 4,
   VTIME = 5,
   VMIN = 6,
-  VSWTC = 7,
   VSTART = 8,
   VSTOP = 9,
   VSUSP = 10,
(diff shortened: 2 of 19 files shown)


tehn:
wow! i'll be able to review this code soon, thank you for investigating!

Dewb:
Awesome. I’ll try to adapt my serial PTZ script to this API sometime this week.

matron/src/device/device_list.c
```diff
@@ -50,6 +51,7 @@ void dev_list_init(void) {
     dq.size = 0;
     dq.head = NULL;
     dq.tail = NULL;
+    pthread_mutex_init(&dq.lock, NULL);
```
tlubke, line 54: Might consider using a static initializer for the mutex, or check the result of `pthread_mutex_init()` for success.

matron/src/device/device_list.c
```diff
@@ -149,6 +161,7 @@ static void dev_list_remove_node(struct dev_node *dn, union event_data *event_re
 }
 
 void dev_list_remove(device_t type, const char *node) {
+    pthread_mutex_lock(&dq.lock);
     struct dev_node *dn = dev_lookup_path(node, NULL);
     if (dn == NULL) {
         return;
```
tlubke, line 167: Mutex deadlock condition? Lock acquired on line-164, if potential return before free if `dn` is null. `dq` isn't referenced in this function directly , so the lock/unlock should probably go inside `dev_list_remove_node()`.

tlubke: More potential deadlocks in the switch statement for `DEV_TYPE_MIDI` and default.

matron/src/device/device_monitor.c
```diff
@@ -276,15 +277,25 @@ void add_dev_tty(struct udev_device *dev) {
...
+        dev_list_add(DEV_TYPE_MONOME, node, name, NULL);
     } else if (is_dev_crow(dev)) {
         fprintf(stderr, "tty is a crow\n");
-        dev_list_add(DEV_TYPE_CROW, node, name);
+        dev_list_add(DEV_TYPE_CROW, node, name, NULL);
     } else {
         fprintf(stderr, "dev_monitor: unmatched TTY device %s at %s\n", name, node);
+        const char *vendor, *model;
+        vendor = udev_device_get_property_value(dev, "ID_VENDOR");
+        model = udev_device_get_property_value(dev, "ID_MODEL");
```
tlubke, line 291: Reiterating serial-number comment from previously, and remembering some serial devices have multiple channels. Can't remember if this has to be selected via a _channel_ or _index_ with udev.

matron/src/device/device_serial.c
```diff
@@ -0,0 +1,134 @@
...
+        lua_gettable(l, -2);
+        if (!lua_isnil(l, -1)) {
+            d->newtio.c_cc[i] = lua_tointeger(l, -1);
+        }
+        lua_pop(l, 1);
+    }
+    lua_pop(l, 1);
+
+    tcflush(d->fd, TCIFLUSH);
+    tcsetattr(d->fd, TCSANOW, &d->newtio);
```
tlubke, line 90: I'm not familiar with these two functions, but looks like they have return codes. Might want to handle the errors, and close `d->fd` before returning if failing.

matron/src/device/device_serial.c
```diff
@@ -0,0 +1,134 @@
...
+            len = 0;
+        }
+        usleep(1000); // 1ms
+    }
+    return NULL;
+}
+
+void dev_serial_deinit(void *self) {
+    struct dev_serial *di = (struct dev_serial *)self;
+    tcsetattr(di->fd, TCSANOW, &di->oldtio);
```
tlubke, line 127: Same return code comment as previous.

I'm not sure I understand what's actually happening when switching between `oldtio` and `newtio`. Perhaps a good candidate for a code comment.

 does the `fd` need to be closed?

matron/src/device/device_serial.c
```diff
@@ -0,0 +1,134 @@
...
+        len = read(di->fd, di->line, 255);
+        if (len > 0) {
+            di->line[len] = 0; // add null to end of string
+            if (len > 1) {
+                // fprintf(stderr,"serial> %s", di->line);
+                handle_event(self, base->id);
+            }
+            len = 0;
+        }
+        usleep(1000); // 1ms
```
tlubke, line 120: This sleep may need to shorter depending on baud rate.

matron/src/device/device_serial.c
```diff
@@ -0,0 +1,134 @@
...
+void *dev_serial_start(void *self) {
+    struct dev_serial *di = (struct dev_serial *)self;
+    struct dev_common *base = (struct dev_common *)self;
+
+    uint8_t len;
+
+    while (1) {
+        len = read(di->fd, di->line, 255);
+        if (len > 0) {
+            di->line[len] = 0; // add null to end of string
```
tlubke, line 113: If `len` is maxed out because `read()` read 255 bytes (I _think_ this is way `read() behaves, but it's late), accessing the 255th index on a `char[255]` would be out-of-bounds. Making line `char[256]` and leaving this limit probably makes the most sense.

Also, double check the return type of `read()`, I think it may be a signed type (`ssize_t`).

matron/src/device/device_serial.h
```diff
@@ -0,0 +1,25 @@
...
+
+struct dev_serial {
+    struct dev_common base;
+    int fd;
+    struct termios oldtio, newtio;
+    char line[255];
+    char* handler_id;
+};
+
+extern int dev_serial_init(void *self, lua_State *l);
```
tlubke, line 20: Do these need to be marked `extern`?

matron/src/weaver.c
```diff
@@ -2289,6 +2316,57 @@ void w_handle_crow_event(void *dev, int id) {
...
+    if (!lua_isstring(lvm, -2)) {
+        fprintf(stderr, "serial handler id expected\n");
+        return;
+    }
+    if (!lua_istable(lvm, -1)) {
+        fprintf(stderr, "serial config table expected\n");
+        return;
+    }
+
+    dev_list_add(DEV_TYPE_SERIAL, path, strdup(name), lvm);
```
tlubke, line 2337: Why is only `name` duplicated here? Where does it get freed?

lua/core/serial.lua
```diff
@@ -0,0 +1,243 @@
...
+end
+
+--- send a message to a serial device
+function serial.send(dev, line)
+    _norns.serial_send(dev, line)
+end
+
+_norns.serial = {}
+
+function _norns.serial.config(vendor, model)
```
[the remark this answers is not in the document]
midouest: I'm not opposed to including the serial number here, but wouldn't they already be able to connect multiple of the same device in the current implementation? `_norns.serial.config` would match both devices that have identical vendor/model, and then `_norns.serial.add` would get called once for each. The mod author could then handle the duplicate devices however they choose. I haven't tested this so I could be missing something! Without the serial number there might not be a great way to distinguish between them beyond Lua object identity.

matron/src/device/device_list.c
midouest: Good catch! Thanks

matron/src/device/device_serial.c
midouest: Closed the `fd` here (I wonder if this should be done for crow also?). Still thinking about your other comment

matron/src/device/device_serial.h
midouest: They do not. Removed!

lua/core/serial.lua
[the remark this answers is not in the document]
midouest: Added serial number to the config callback!

matron/src/device/device_monitor.c
midouest: Hmm, it would be helpful to have an example of this. I added the interface number from `ID_USB_INTERFACE_NUM` to the serial config args. Based on some googling it seems like devices with multiple serial ports might use that to distinguish between ports? Let me know if that seems right to you!

matron/src/weaver.c
midouest: The instance of `name` being duped is owned by the event and gets freed in `event_data_free`. 

`dev_new` dupes `path` but not `name`. This works for other devices because they immediately take ownership of `name` when they are recognized. The serial devices are configured asynchronously, so I had to do things a bit differently.

midouest: Another option would be to not free the name in `event_data_free` and leave it to `w_handle_serial_config` to either free the name on failure or pass ownership to `dev_new`. Maybe that is a bit cleaner?

matron/src/device/device_serial.c
midouest: Updated this to automatically determine an appropriate sleep duration using the baud rate, character size, stop bits and parity flags. Depending on the flags used I think this could be as fast as 382 microseconds (B4000000, CS5, 1 stop bit, no parity bits) or as slow as 56 seconds (B50, CS8, 2 stop bits, 1 parity bit).

matron/src/weaver.c
tlubke: I think it's good as-is. Was originally asking as a genuine question more so than a rhetorical one -- so thanks for clearing up how it works to me :)

matron/src/device/device_monitor.c
tlubke: Note sure! I have a 4-channel FTDI module that I can test with to verify.

tlubke (commented):
Additions look good! I'll have to try this out on my norns next.

midouest:
@tlubke Thanks for the feedback! I changed up how the device is configured from Lua based on some [best practices](https://www.gnu.org/software/libc/manual/html_node/Setting-Modes.html) that I found.

matron/src/device/device_serial.c
```diff
@@ -0,0 +1,307 @@
...
+    unsigned int char_size = get_character_size(cflag);
+    unsigned int stop_bits = get_stop_bits(cflag);
+    unsigned int parity_bits =  get_parity_bits(cflag);
+
+    unsigned int bits_per_char = char_size + stop_bits + parity_bits;
+    double chars_per_second = baud_rate / bits_per_char;
+
+    unsigned int us = floor(SECOND * max_read / chars_per_second);
+    // latency becomes noticeable above 1ms
+    if (us > MILLISECOND) {
```
tlubke, line 303: This (and the thread loop) actually might need a bit of review. I think this probably should return rate of microseconds per byte.

i.e.
```
    ...

    double bits_per_symbol = char_size + stop_bits + parity_bits;
    double bits_per_second = baud_rate * bits_per_symbol;
    double bytes_per_second = bits_per_second / 8.0;

    return ceil(1e6 / bytes_per_second);
}
```

Then, when `usleep()` is called, you can either pass in that value or `(max_read * microseconds_per_byte)`. I'm not sure which is correct because the former may take up too much execution time, but would catch a one byte transfer immediately.

Haven't had a chance to test yet, but I think there may also be an implicit block happening in the `read()` call. Maybe you could add a print statement and see what the clock reading is before and after that call. Maybe there is no need for `usleep()` in the first place if the thread can suspend while waiting for a complete line from the serial device.

Another option is to just hardcode the sleep to 256 μs which I _think_ should be cover the maximum byte rate and catch each byte.

midouest: Good point, there is a block in `read()`. I put some timers in there and it looks like it blocks for half a second if there's no data, and otherwise reads as fast as data comes in (between 40us - 40ms testing at 115200 baud, CS8 + 1 stop bit). After testing thread cpu time with and without the usleep, it seems like it does slightly better without the sleep, so I opted to remove the sleep!

lua/core/serial.lua
```diff
@@ -0,0 +1,250 @@
...
+--- send a message to a serial device
+function serial.send(dev, line)
+    _norns.serial_send(dev, line)
+end
+
+_norns.serial = {}
+
+function _norns.serial.match(vendor, model, serial_num, interface_num)
+    for id, handler in pairs(serial._handlers) do
+        if handler.match(vendor, model, serial_num, interface_num) then
```
midouest, line 23: I'm thinking about passing these attributes to `handler.match` in a readonly table instead of as positional arguments. That might make this a little easier to manage if attributes need to be added or removed in the future.

lua/core/serial.lua
```diff
@@ -0,0 +1,250 @@
...
+            return id
+        end
+    end
+    return nil
+end
+
+function _norns.serial.configure(handler_id, tio)
+    return serial._handlers[handler_id].configure(tio)
```
midouest, line 31: When I was writing `handler.match` I was thinking it would be useful for it to be able to handle a family of devices that might have different attributes but generally speak the same language. However, `handler.configure` currently doesn't have the same flexibility because all it receives is the termios config. I've been wondering if I need to pass in the same attributes that `handler.match` receives, or give `handler.match` the ability to return some sort of user data that gets passed through to `handler.configure`.

Maybe it's not a big deal though. I think users could work around this limitation by defining multiple serial handlers, but it might be a bit of a headache to wrap that up in a nice interface for scripts.

lua/core/hook.lua
```diff
@@ -36,6 +36,7 @@ local hooks = {
   script_pre_init = Hook.new(),
   script_post_init = Hook.new(),
   script_post_cleanup = Hook.new(),
+  system_pre_device_scan = Hook.new(),
```
midouest, line 39: In hindsight maybe this hook isn’t necessary? I could just call serial.handler in the global scope of the mod since mods are loaded as the last step of startup script.

tlubke:
Found a few things in testing.

1. I was able to detect my 2-channel FTDI device after the last commit! I don't think https://github.com/monome/norns/pull/1818/commits/472fbd3f6226716d3604e96d348eadfe48c25b35 can stay though. I know some of the early grids are FTDI-based, so you will have to check with @tehn for what the EEPROM is set to for those vendor/model values to not break compatibility.
2. If a Lua exception takes place during `"system_pre_device_scan"` it appears to crash Matron. _(I didn't see it in `journalctl`, I always forget where to go for logs after being away from norns dev for a long time...)_
3. In `device_monitor.c:add_dev_tty()`, the `vendor`, `model`, `serial`, and `interface` strings could create a new empty string `nil` when that is empty. I could see an argument for leaving it as-is, but it feels like creating more error-handling work for the user.

Next step for me is probably to create a simple loopback mod and hook it up to a logic analyzer.

tehn:
thanks for such incredible reviews!

so yes, until 2021 all monome grids (and arcs) use the FTDI driver. vendor is "monome" and i think the dev can get passed to `is_dev_monome_grid()` for device type id

happy to test with hardware, can dig out an older one

midouest:
@tlubke Updated with your suggestion for defaulting `vendor`, `model`, `serial` and `interface` to empty strings instead of nil. However, I wasn't able to reproduce matron crashing due to Lua exceptions in `system_pre_device_scan`. Here's a snippet of the logs if I call a function that doesn't exist:

```
running startup...
error loading keyboard layout, using old value: us
start_audio(): 
loading mod: playdate (playdate/lib/mod)
loading mod: playdate-arc (playdate-arc/lib/mod)
calling: init playdate arc handlers
calling: playdate device handler
hook: playdate device handler failed, error: /home/we/dust/code/playdate/lib/mod.lua:143: attempt to call a nil value (field 'handler')
scanning devices...
dev_monitor: vendor=monome model=monome
dev_monitor: TTY appears to be a grid
```

I also made a couple small changes to the API:
- `serial.handler(args)` is now `serial.add_handler(args)`
- The `match(vendor, model, serial, interface)` callback is now `handler.match(attrs)`, where `attrs` is a table with `vendor`, `model`, `serial` and `interface` keys

midouest:
@tehn I think I have a 2015 edition grid and it is still detected after the grid detection change

tehn:
@midouest 2015 is FTDI, yes. i also just tested your branch with some FTDI devices and everything seems great!

i just merged https://github.com/monome/norns/pull/1820 which likely created a very trivial merge conflict, sorry!

looking to do a norns update (for new arc users, which need that one line update for CDC detection)--- would love to include this if it's feeling ready. seems like we're close, if not there?

tlubke:
I'll do a final review if you're to wanting to expedite this PR.

One final thought on grid detection, is the neonome / DIY grid dependent on the removed logic? Trying to read through https://llllllll.co/t/diy-monome-compatible-grid-w-adafruit-neotrellis thread to find evidence of what it enumerates as. Will head to the source code of those projects if I don't find it here.

lua/core/hook.lua
tlubke: This might be worth discussing again before merge. On the one hand it works as it is now. But on the other, I am of the opinion that fewer hooks is better. Another point is that the device scan stage isn't referenced very often from what I know of so it might come across as a bit mysterious since it looks different from the other hooks.

A `system_pre_startup` hook would fit the existing hooks a little more closely, but would require moving the mod/hook initialization outside of `startup.lua`.

tlubke:
Aside from https://github.com/monome/norns/pull/1818#discussion_r2024002893, I don't have anything else to add. Might submit a patch with some LDoc suggestions in `serial.lua` if you're open to it @midouest .

lua/core/hook.lua
midouest: Good points. I'm in agreement and inclined to remove it. I was thinking maybe it makes sense to have an official hook to decouple serial registration from mod loading, even if they happen at the same time right now.

A good future addition might be to rescan devices when a new handler is added. Then it wouldn't matter if the handlers are registered in a mod or a script.

matron/src/device/device_serial.c
```diff
@@ -0,0 +1,188 @@
...
+    lua_pushinteger(l, d->oldtio.c_lflag);
+    lua_setfield(l, -2, LFLAG);
+    lua_pushinteger(l, d->oldtio.c_line);
+    lua_setfield(l, -2, LINE);
+
+    lua_createtable(l, NCCS, 0);
+    for (int i = 0; i < NCCS; i++) {
+        lua_pushinteger(l, d->oldtio.c_cc[i]);
+        lua_rawseti(l, -2, i+1);
+    }
```
tlubke, line 75: I'm not super familiar with manipulating Lua state, `lua_createtable()` is being prepped for 8 records, but it looks like there might be 9 -- is `NCCS` is part of that table too or does it end with `CC`?

EDIT: Nevermind, appears `NCCS` is a nested table inside `CC`.

tlubke (approved):
After a final review I think this is good to merge, haven't done any more testing since my last comment though.

lua/core/serial.lua
```diff
@@ -0,0 +1,326 @@
...
+  VKILL = 3,
+  VEOF = 4,
+  VTIME = 5,
+  VMIN = 6,
+  VSWTC = 7,
+  VSTART = 8,
+  VSTOP = 9,
+  VSUSP = 10,
+  VEOL = 11,
+  VREPRINT = 12,
```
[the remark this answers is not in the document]
midouest: The values in these tables came from the termios headers on my Norns device. I was also able to find documentation for the iflags, oflags, cflags and lflags in the local termios headers. I pulled the CC documentation from https://pubs.opengroup.org/onlinepubs/9699919799/. I found further CC documentation at https://www.man7.org/linux/man-pages/man3/termios.3.html, but left out some values that were not in posix and/or not supported on Linux.

I think I'm going to leave the undocumented ones out for now.

matron/src/device/device_serial.c
```diff
@@ -0,0 +1,188 @@
...
+#define OFLAG "oflag"
+#define CFLAG "cflag"
+#define LFLAG "lflag"
+#define LINE "line"
+#define CC "cc"
+
+int dev_serial_init(void *self, lua_State *l) {
+    struct dev_serial *d = (struct dev_serial *)self;
+    struct dev_common *base = (struct dev_common *)self;
+
```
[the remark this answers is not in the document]
midouest: Done

PR #1821, opened 2025-04-05
Add more Pi variant platforms

okyeron:
PR to expand `platform.c` to include other Pi variants including `PLATFORM_PI4`, `PLATFORM_CM4`, and `PLATFORM_CM4S`

Rationale: Addresses cases of norns shield users with a Compute Module 4 in a Waveshare carrier board where the user will get the display rotated by 180 because the platform is incorrectly interpreted.

Previously tested as working with CM4 in a Waveshare carrier board. Does not change original enum indexing, so it should not affect CM3 norns.

### DIFF
--- a/matron/src/hardware/platform.c
+++ b/matron/src/hardware/platform.c
@@ -2,16 +2,36 @@
 #include <stdlib.h>
 
 #include "platform.h"
+#include <string.h>
+#include <stdio.h>
 
 platform_t p;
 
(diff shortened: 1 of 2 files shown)


tehn (changes requested):
looks good! could you remove the commented out code?

tehn (approved):

okyeron:
Comments removed. Thank you!

PR #1823, opened 2025-04-06
fix crash on tty device connection when no serial mod is enabled

midouest:
This was being set in a mod so I didn't notice that it was missing from the standard globals. 🤦‍♀️ 

Without it, Norns was crashing as soon as a tty device was connected.


tlubke (approved):

tlubke:
@tehn Gonna merge this so [250406](https://github.com/monome/norns/releases/tag/v2.9.0) can be hot fixed.

tehn:
thanks!! will update the update :)

PR #1825, opened 2025-04-08
feat: add `clock.link.get_number_of_peers()`

robbielyman:
this PR adds `_norns.clock_link_get_number_of_peers()` a new Lua function which will return the number of peers connected to this Norns over Ableton Link.

[related discussion on Lines:](https://llllllll.co/t/ableton-link-and-norns-question/39335/11?u=alanza)

~~this PR does *not* make a user-facing change to display the number of peers anywhere on norns, but it enables that change to be made solely in Lua. i'd happily help brainstorm or contribute that Lua change as well, if there's interest.~~

### DIFF
--- a/matron/src/clocks/clock_link.c
+++ b/matron/src/clocks/clock_link.c
@@ -16,7 +16,6 @@ static pthread_t clock_link_thread;
 static struct clock_link_shared_data_t {
     double quantum;
     double requested_tempo;
-    uint64_t number_of_peers;
     bool playing;
     bool enabled;
     bool start_stop_sync;
--- a/matron/src/clocks/clock_link.c
+++ b/matron/src/clocks/clock_link.c
@@ -100,7 +100,6 @@ void clock_link_start() {
     pthread_attr_t attr;
     pthread_attr_init(&attr);
 
-    clock_link_shared_data.number_of_peers = 0;
     clock_link_shared_data.quantum = 4;
     clock_link_shared_data.requested_tempo = 0;
     clock_link_shared_data.enabled = false;
--- a/matron/src/clocks/clock_link.c
+++ b/matron/src/clocks/clock_link.c
@@ -25,15 +25,15 @@ static struct clock_link_shared_data_t {
     pthread_mutex_t lock;
 } clock_link_shared_data;
 
+static abl_link link;
+
 static clock_reference_t clock_link_reference;
 
 static void *clock_link_run(void *p) {
     (void)p;
 
-    abl_link link;
     abl_link_session_state state;
 
-    link = abl_link_create(120);
     state = abl_link_create_session_state();
 
     while (true) {
@@ -81,11 +81,6 @@ static void *clock_link_run(void *p) {
             }
 
             abl_link_enable(link, clock_link_shared_data.enabled);
-            if (!clock_link_shared_data.enabled) {
-                clock_link_shared_data.number_of_peers = 0;
-            } else {
-                clock_link_shared_data.number_of_peers = abl_link_num_peers(link);
-            }
             abl_link_enable_start_stop_sync(link, clock_link_shared_data.start_stop_sync);
 
             pthread_mutex_unlock(&clock_link_shared_data.lock);
@@ -98,6 +93,7 @@ static void *clock_link_run(void *p) {
 }
 
 void clock_link_init() {
+    link = abl_link_create(120);
     clock_reference_init(&clock_link_reference);
 }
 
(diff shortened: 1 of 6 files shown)


artfwo:
thanks for the contribution!

i recall there's no need to pass returned number of peers through protected `link_shared_data` struct, as `numPeers()/abl_link_num_peers()` is thread-safe, so the lua function could just call abl_link_num_peers() directly every time (saving the need to update the field in the scheduler callback, which gets called very frequently).

could you update the PR accordingly? thx!

robbielyman:
yes, i’d be happy to make that change, but unfortunately i need more guidance: since the `link` pointer is only present in the clock thread, the ableton function can currently only feasibly be called from that thread. changing that would require placing the link pointer somewhere accessible from the main thread, which seems to me like a can of worms better left unopened.

in case this is useful information, given that the function is marked as realtime safe in the header by Ableton, it sounds like it’s not a problem to call very frequently.

artfwo:
> yes, i’d be happy to make that change, but unfortunately i need more guidance: since the `link` pointer is only present in the clock thread, the ableton function can currently only feasibly be called from that thread. changing that would require placing the link pointer somewhere accessible from the main thread, which seems to me like a can of worms better left unopened.
> 
> in case this is useful information, given that the function is marked as realtime safe in the header by Ableton, it sounds like it’s not a problem to call very frequently.

yeah, link object itself could be moved out of the thread, for instance declared as `static` in `clock_link.c`, and initialized in `clock_link_init()`.

the wrapper in `clock.c` is also superfluous, `clock_link.h` is included in weaver, so the function can be called from there directly too. hope this helps!

robbielyman:
i made the change! it might now not be the case that the number of peers is zero when norns has Link disabled—i'm not likely to test what the behavior is for several days, but it would be a pretty easy check for anyone able to both build norns and connect to it via Link.

artfwo:
the number of peers should be displayed only when clock source is set to `link`. i'd leave that logic to the scripts.

robbielyman:
good point. well ok, i think this is ready for review :)

artfwo:
> good point. well ok, i think this is ready for review :)

`number_of_peers` field in the shared struct is unnecessary.

matron/src/clocks/clock_link.c
```diff
@@ -91,13 +92,15 @@ static void *clock_link_run(void *p) {
...
 void clock_link_init() {
+    link = abl_link_create(120);
     clock_reference_init(&clock_link_reference);
 }
 
 void clock_link_start() {
     pthread_attr_t attr;
     pthread_attr_init(&attr);
 
+    clock_link_shared_data.number_of_peers = 0;
```
artfwo, line 103: it needs to be removed here as well :) please make sure the code builds/runs so it doesn't cause any issues after merging

robbielyman: Sure. I'll mark it as a draft for now, since building norns is relatively difficult.

robbielyman:
just got a chance to sync up with my norns; everything seems a-ok! here's a little test script to demonstrate:

```lua
function init()
  clock.run(function()
      while true do
        clock.sleep(1)
        redraw()
      end
  end)
end

function redraw()
  screen.font_size(15)
  screen.move(1,50)
  screen.clear()
  if params:get("clock_source") ~= 3 then
    screen.text("source != link")
  else
    screen.text("peers: " .. _norns.clock_link_get_number_of_peers())
  end
  screen.update()
end
```

to verify, make sure your norns is on the same network as another link-enabled device, set the `clock_source` param to `"link"` and join the link session with another device. the script's screen should say `peers: 1` (or some other positive number matching your expectations).

artfwo:
Generally the `_norns` API is sort of reserved for libraries. I'd suggest providing a wrapper in `core/clock.lua`.

`clock.get_num_link_peers` or similar.

matron/src/clocks/clock_link.c
```diff
@@ -24,7 +23,7 @@ static struct clock_link_shared_data_t {
     pthread_mutex_t lock;
 } clock_link_shared_data;
 
-static abl_link link;
+static abl_link link_instance;
```
artfwo, line 26: a better name for this could be `clock_link` for consistency with the rest of variables.

robbielyman:
i’m well aware of the distinction between the `_norns` interface and user-facing Lua changes like `clock.lua`. (i mean, i also know that you know that i’m aware.) as i mentioned in the PR description, making those changes is beyond the scope i set for myself. 

now, i’m happy to increase the scope, since reading between the lines of your comment it sounds to me like you want me to. i think it would be helpful for me if you could communicate similar asks more directly in the future.

artfwo:
> now, i’m happy to increase the scope, since reading between the lines of your comment it sounds to me like you want me to. i think it would be helpful for me if you could communicate similar asks more directly in the future.

yeah, it's a tiny change that improves the usability of the feature, so why not :) thanks for adding it!

PR #1826, opened 2025-04-08
serial handlers can transmit and receive messages with embedded zeroes

midouest:
Switch to using `lua_pushlstring` so that messages containing null characters can reach the Lua handler. This makes it possible to interpret the serial data as something other than ascii characters.

I also changed the serial event struct to embed a heap-allocated copy of the serial data. The serial read thread does not sleep other than waiting for IO. It's possible for it to read multiple times in a row before the event loop has a chance to process the messages if the serial device is sending a lot of data. This could lead to messages being overwritten in the buffer and effectively "dropped".

EDIT: Oops, just realized this affects sending messages also.

### DIFF
--- a/matron/src/events.c
+++ b/matron/src/events.c
@@ -145,6 +145,9 @@ MATRON_API void event_data_free(union event_data *ev) {
     case EVENT_SERIAL_REMOVE:
         free(ev->serial_remove.handler_id);
         break;
+    case EVENT_SERIAL_EVENT:
+        free(ev->serial_event.data);
+        break;
     case EVENT_CUSTOM:
         if (ev->custom.ops->free) {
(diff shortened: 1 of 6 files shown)


tehn:
looks good to me!

@tlubke want to have a quick look?

matron/src/events.c
```diff
@@ -145,6 +145,9 @@ MATRON_API void event_data_free(union event_data *ev) {
     case EVENT_SERIAL_REMOVE:
         free(ev->serial_remove.handler_id);
         break;
+    case EVENT_SERIAL_EVENT:
+        free(ev->serial_event.data);
```
tlubke, line 149: Set `serial_event.data` back to `NULL` to avoid potential dangling pointer issues.

tlubke (approved):

PR #1827, opened 2025-04-09
Print serial terminal settings

midouest:
- Sets a TerminalIO metatable on the `term` table passed to `handler.configure(term)` with a `:print()` method for printing the current settings. This functions similarly to executing `stty -F /path/to/tty -a` on the command line.
- Fixes a small typo in the cc table docs
- Fixes an off-by-one error when loading the CC values into Lua

### DIFF
--- a/lua/core/serial.lua
+++ b/lua/core/serial.lua
@@ -12,7 +12,8 @@ local serial = {
 -- @param args.match function(attrs) returning true if the handler should accept the device with the given attributes, 
 --   otherwise false
 -- @param args.configure function(term) receiving the device's initial terminal settings and returning the new terminal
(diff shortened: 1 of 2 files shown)


tlubke:
Feels a bit heavy for what it's doing. A couple of notes, but will think about this a little bit more too.

* `TerminalIO` isn't named very Lua-like.
* Probably don't need `TerminalIO:print()` since `__tostring()` is already set in the metatable.
* In the `__tostring()` method, I wonder if there is a clever way to make use of the existing tables instead of creating more.
* I think making output closer to the tables/variables in the `serial` module makes more sense than making output match `stty`.
* If you want to make a separate PR for the two fixes, I don't see a problem with those being merged right away.

PR #1831, opened 2025-04-16
Fixed tab.readonly() not endowing pairs/ipairs.

tlubke:
* Made all of `proxy`'s meta-table methods local functions defined inside `tab.readonly()`.
* For `pairs` implementation, use `next()` as an iterator but only return a value via `proxy`'s `__index`.
* For `ipairs` implementation, use a for-loop over the range of `tab.count(t)`, but only return indices that pass `proxy`'s `__index`.

### DIFF
--- a/lua/lib/tabutil.lua
+++ b/lua/lib/tabutil.lua
@@ -218,22 +218,52 @@ function tab.readonly(params)
   local t = params.table
   local exceptions = params.except or {}
   local proxy = {}
-  local mt = {
-    __index = function(_, k)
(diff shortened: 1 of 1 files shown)


tlubke:
NOTE: I have only tested this on my local Lua v5.4.7 interpreter, need to test with norns and check all core modules that use `tab.readonly()`.

tehn (approved):

tlubke:
I can patch this quick to fix the lint fails.

PR #1832, opened 2025-04-17
issue screen error on audio system fail

### LINKED ISSUE #1830: black screen is bad
~~Previously matron would run even if jack wasn't running. Now that jack is a dependency, matron quits if jack isn't running.~~ _Edit: this was always the case since we added jack as a dependency for the clock module. What's changed is screen handling._

If there's a circuit problem or image problem or systemctl problem and for some reason jack isn't running, matron quits and we get no information on the screen.

Matron should at least turn the screen on, display and error, _then_ quit.

tehn:
fixes #1830

Shows "audio system fail" when jack does not start.

Potential improvement might be "any key to shut down" --- currently shutdown requires remote login or white button. We don't want to shut down with a timer, to allow debugging.

### DIFF
--- a/matron/src/jack_client.c
+++ b/matron/src/jack_client.c
@@ -4,6 +4,7 @@
 #include <jack/jack.h>
 
 #include "screen.h"
+#include "hardware/screen/ssd1322.h"
 
 static jack_client_t *jack_client;
 double jack_sample_rate;
@@ -35,9 +36,10 @@ int jack_client_init() {
 fail:
   screen_clear();
   screen_level(15);
-  screen_move(0, 60);
-  screen_text("jack fail.");
+  screen_move(10, 40);
+  screen_text("audio system fail.");
   screen_update();
+  ssd1322_refresh();
   return 1;
 }
 
--- a/matron/src/jack_client.c
+++ b/matron/src/jack_client.c
@@ -3,8 +3,6 @@
 
 #include <jack/jack.h>
 
-#include "screen.h"
-#include "hardware/screen/ssd1322.h"
 
 static jack_client_t *jack_client;
 double jack_sample_rate;
@@ -34,12 +32,6 @@ int jack_client_init() {
   jack_sample_rate = (float)jack_get_sample_rate(jack_client);
   return 0;
 fail:
-  screen_clear();
-  screen_level(15);
-  screen_move(10, 40);
-  screen_text("audio system fail.");
-  screen_update();
-  ssd1322_refresh();
   return 1;
 }
 
--- a/matron/src/main.c
+++ b/matron/src/main.c
@@ -74,8 +74,12 @@ int main(int argc, char **argv) {
     battery_init();
     stat_init();
     osc_init();
-    jack_client_init();
     ssd1322_init();
+    if(jack_client_init()) {
+    	usleep(1000000);
+    	ssd1322_deinit();
+    	return -1;
+    }
     clock_init();
     clock_internal_init();
     clock_midi_init();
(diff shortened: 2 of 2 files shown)


tlubke:
I think this is fine as a fix, but would be nice to keep an extra dependency away from `jack_client.c`. I might try moving the screen results + events inits right after `config_init()` in `main()`. I'm curious if that also fixes the problem without breaking anything.

```
    screen_results_init();
    screen_events_init();
```

tlubke:
I got an alternative solution working if you would consider that as well. Basically goes like this:
* Fix the SEGFAULT caused by accessing nullptr in `jack_client.c` when client fails initialization.
* Check status of critical dependencies right before `w_startup()`.
* Exit if dependencies failed, triggering `cleanup()`.
* Add `screen_event_deinit()` which blocks until remaining screen events are processed.
* Move `ssd1322_deinit()` to be the last function called in `cleanup()`.

tehn:
I think I discovered that calling `sd1322_deinit()` clears the screen, which isn't great for maintaining indication of an error.

Happy to consider alternatives, will wait for your suggestion.

tlubke:
Here's a little work I just committed to my branch: https://github.com/monome/norns/commit/436f50db45fd2e925e413d7bf1465eb393b14548

This turned into more a refactor than I had initially thought, so I'll do my best to document the paths I've taken and where it might go next.

> I think I discovered that calling sd1322_deinit() clears the screen, which isn't great for maintaining indication of an error.

Because of this fact, I initially went down the path of wrapping `exit()` so that `ssd1322_deinit()` could leave the RST pin pulled high before closing its resources. This worked to keep the screen on for displaying errors, but resulted in a couple of new issues:
 1. The norns menu "SLEEP" option which does an OS shutdown, might not cause Matron to even exit, let alone exit with `0`.
 2. After waking up the next time when RST was left high, the screen basically shows a white noise pattern.
 
 Next, I removed the `screen.h` dependency in `jack_client.c` and moved the error handling and display into `main()`. I had the idea to poll for user input intervention to complete the shutdown so that the screen can be shot off. I couldn't find a clean way to interact with key/enc since a lot of that code is private scope. So, I decided to try it via the event loop. Unmodified, event_handle_pending() has the effect of throwing a bunch of Lua VM errors since the jack initialization error was being handled before a lot of weaver startup code is called. So, I added some functions to "arm" an exit based on a specific event type, and then added a while loop to check the condition. To avoid the Lua errors, I just have the actual event handling ignored if the condition exit is armed.
 
This works to shutdown matron and the screen, but doesn't leave the user a way to restart aside from the button on the back of norns. From what I can tell, that is the same place the user would be if the jack client failed to init even without all these changes though.

All this to say... I think a slight refactor of `main()` is probably worth discussing. There are quite a few resources that don't get de-initialized, and I even found some pthread locks that aren't being initialized.

For the sake of just fixing the black screen though, this PR is definitely lighter weight.

tehn:
@tlubke Oy I am sorry it led to such a wormhole.

For "normal use" de-initialization isn't as critical since the system is usually shut down.

For on-device development and testing it'd be reasonable to have good cleanup, since the application is repeatedly run and re-run.

For the fail state I can only see this PR as an improvement--- indicates critical problem and white button is fine for shutdown (power cut during likely minimal disk activity).

One cleanup I can do is to remove the screen.h/etc from jack_client as you mention, and have all the screen error happen in main. I think that'd feel cleaner.

But beyond that I'd leave a more substantial refactor for a later day.

Thanks for taking a hard look at this!

Issue #1741, opened 2023-11-24
screen q size / indicator


catfact:
like an idiot, i entirely forgot i had added a visual "warning" for when the screen event queue is filled.

in a way it's good that we know some scripts are hitting the Q capacity. but it makes me wonder if the Q shouldn't be bigger / grow as needed. (we don't say that you can't queue up thousands of drawing requests at once.)

tlubke:
This conversation on lines gave me an idea.

> Apologies if I’m misunderstanding the actual issue, but: if the queue is long enough would it make sense to skip queue entries rather than flush the queue entirely? e.g., every other or every 4th entry or whatever. Basically rather than having the script author reduce their framerate, reduce it for them at the system level :thinking:

> no, it’s not a queue of frames

The idea is pretty simple -- anytime the queue needs to skip an entry, we can clear everything up until the next `SCREEN_UPDATE` event. Though it's not a queue of frames _exactly_, it is a queue of events that make up frames.

That probably wouldn't fix the case where there are enough drawing requests to fill up the queue before a screen update event. I think your idea for a dynamically resizing queue makes a lot of sense for that reason.

xmacex:
For documentation, the flashes look like this; a single-frame half-screen yellow rectangle.

![VID_20231123_230030_exported_7697.jpg](https://github.com/monome/norns/assets/879300/0fc403c5-e384-4ab9-8549-7cca7600de1b)

toneburst:
> For documentation, the flashes look like this; a single-frame half-screen yellow rectangle.
> 
> ![VID_20231123_230030_exported_7697.jpg](https://github.com/monome/norns/assets/879300/0fc403c5-e384-4ab9-8549-7cca7600de1b)

Incidentally, that's my "bline" script. I rewrote the display module in the forthcoming 2.0 version, dumping a lot of the non-changing GUI elements into a fullscreen background image, which seems to have fixed the overflow issue, even with older Norns versions.

I'll probably be doing something similar in future scripts.

PR #1833, opened 2025-04-21
Added resizing to screen_events.c (#1741)

tlubke:


tlubke:
Might make a separate PR later for dropping inferred "frames" instead of clearing the whole queue.

tehn (approved):
Code looks good, and curious about the use case but happy to merge.

tlubke:
Originally closed because I had totally disregarded the queue was a circular buffer... So, I fixed the resize logic and reworked the reset function a little.

For anyone reviewing, please keep a skeptical eye open for corner cases (I feel like I might be missing something).

tehn:
Code looks good with my quick review, did not test on device.

Given the amount of RAM this system has, would it be bad to just keep the queue always really big?

tlubke:
> Given the amount of RAM this system has, would it be bad to just keep the queue always really big?

Certainly could take this approach, each `screen_event_data` should be 64-bytes assuming worst case alignment. Or 0.5MiB if `SCREEN_Q_SIZE` is changed to 8192.

tehn:
Given the processor has 1GB of RAM, I feel the tradeoff for simply allocating the relatively small amount of memory would be a good tradeoff to the extra complexity for changing the buffer size.

tlubke:
> Given the processor has 1GB of RAM, I feel the tradeoff for simply allocating the relatively small amount of memory would be a good tradeoff to the extra complexity for changing the buffer size.

Sounds good, how about 16K for the new size? Enough to call `screen.pixel()` 128x64 times without overflowing the queue and works out to a nice 1MiB of static allocation.

tehn:
Fantastic!

PR #1835, opened 2025-05-24
fix cm4s platform detection and screen

ngwese:
this probably impacts a very tiny number of norns users... possibly only one

* adds `platform_name()` function for obtaining a string representation of which platform was detected
* fixes "CM4S" detection, previously always reported as "CM4" due to substring matching
* chooses normal screen orientation for CM3, CM4, and CM4S devices (CM4 case is debatable)
* reports the platform name on startup along with platform enum number

warning: i haven't tested this on any other devices other than a unicorn og norns + cm4s

### DIFF
--- a/matron/src/hardware/platform.c
+++ b/matron/src/hardware/platform.c
@@ -10,27 +10,48 @@ platform_t p;
 void init_platform() {
   if(access("/sys/firmware/devicetree/base/model", F_OK ) != -1) {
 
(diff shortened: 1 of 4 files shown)


ngwese:
added one additional change where the platform will always be defaulted to UNKNOWN which better matches the prior logic which falls back if the `/sys/...` file can't be read.

tehn (approved):

PR #1838, opened 2025-06-02
Add Dockerfile for CI

Dewb:
This Dockerfile supports building the norns stack in an ARMv8 container with an OS environment that attempts to be as similar as possible to the norns disk image, from any host environment/architecture that can run Docker with the Docker BuildKit virtualization support (included in Docker Desktop, extra configuration required for Linux, [see here](https://docs.docker.com/build/building/multi-platform/#install-qemu-manually).)

Some additional details in the Dockerfile comments.

To run locally, the following two commands are sufficient to prepare the image and run the norns waf build process:
```zsh
% docker build -t norns-ci . 
% docker run -v $(PWD):/norns-build -t norns-ci 
```

The next step is to leverage this for CI using GitHub Actions. The simple strategy would be to have a single action that does both; unfortunately, the build step can take one or more hours on virtualized armv8. 

A better strategy for CI would be to have one GitHub Actions workflow that runs only when the Dockerfile changes, builds the image with `docker build`, and stores it in the GitHub Container Registry. Then a second Action could run on PRs/merges/etc. and just invoke the `docker run` step, pulling the image from the GHCR.

Some references for further Actions work:
- How to configure an Action to use virtualization: https://docs.docker.com/build/ci/github-actions/multi-platform/
- How to configure an Action to publish to the GHCR: https://dev.to/willvelida/pushing-container-images-to-github-container-registry-with-github-actions-1m6b

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,133 @@
+FROM --platform=linux/arm/v8 dtcooper/raspberrypi-os:bullseye AS build_armv8
+
+# norns-ci container definition 
+#
+# This container is focused on creating an accurate build environment for CI. It does 
+# not attempt to add any extra bits to make the container a usable way for a desktop
+# user to interact with a running virtual norns environment in realtime. Emulating armv8 
+# can be much slower (10-20x) than running on the host architecture.
+#
+# To build norns in a local armv8 container using a macOS or Windows machine with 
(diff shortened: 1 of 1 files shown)


tehn:
wow, amazing to see this! it's outside my expertise so any github integrations you'd like to try out, please pursue

tehn (approved):

Issue #1749, opened 2023-12-11
synchronizing with screen updates


Dewb:
Consider the following use cases:

**Scenario A**: a mod wants to know when the screen has been updated, whether by script, menu, or otherwise, and also wants to read an up-to-date Cairo surface or pixel buffer with the latest contents. _Examples: [ndi-mod](https://github.com/Dewb/ndi-mod) sending the screen over the network, a video recorder mod, other hypothetical screen output mods._

**Scenario B:** a script or mod wants to read the latest contents of an offscreen image after drawing to it. The script knows when it has drawn to the offscreen image, so notification is not strictly necessary like in Scenario A, but the result still needs to be read. _Examples: scripts using ndi-mod to send offscreen images over the network, scripts or mods to interact with hardware devices with small screens like the Ableton Push or Elgato Streamdeck._

Challenges: 
- Cairo drawing happens in a separate thread, and there is not currently a way for Lua script/mod code or C code, all of which run in the main thread, to know when the buffer is complete and consistent.
- Hooking `refresh()` gives you the update notification necessary in Scenario A, but isn't relevant for Scenario B.
- The methods to focus drawing on an offscreen image are not yet routed to the Cairo thread. (#1747)

Some possible solutions:
1. Double buffer: Allocate a second copy of one of the buffers in `ssd1322.c` (either the local copy of the Cairo ARGB buffer, or the 8-bit SPI buffer) and swap between the two every frame. Expose the most recently drawn one to the C API. Mods could then hook `refresh()` and then get the latest frame from a C function. This is probably the smallest and most performant fix to Scenario A, but does not address Scenario B.
2. Allow mods to register a C function to be run synchronously on the Cairo thread during `screen_update()`.  `screen_update()` only calls `ssd1322_update()` if the primary context is current. This would address Scenario A and Scenario B (assuming #1747 is fixed), but only for mods. 
3. Add an optional argument to the Lua `screen.update()` providing a Lua function to be run synchronously on the Cairo thread after the update is finished and the Cairo surface is valid. Solves Scenario A and B for both mods and scripts, but requires invoking Lua from the Cairo thread, opening several cans of worms.
4. Add an optional argument to the Lua `screen.update()` that provides a Lua function to be run on the main thread, using the `screen_results_wait()` system that supports `screen.peek()` etc.
[...]

PR #1839, opened 2025-06-17
Add Pico-LCD-1.14 support and I2S audio configuration

TarasShu:


ngwese:
others more directly involved can offer more comprehensive feedback but a few things i noted from a quick read through
- generally speaking the project has avoided accepting changes that the maintainers can't themselves test
- it seems like the expected display resolution is no longer 128x64?
- having much more detail around the change itself would help a reviewer. what is motivating the change? why was this design/abstraction chosen?  why can't configuration of the audio hardware be handled as it is on the existing devices (i.e. the device tree and referring to an alsa device by name)?

tehn:
i appreciate your inclination to contribute.

in an effort to keep the the codebase focused on the primarily supported hardware, i'm going to decline this addition. please maintain a fork, the way the the fates project does.

Dewb:
The solution I have in mind for #1749 should make it possible to support additional screen output strategies (whether network-based or local hardware) with a mod rather than changes in core. I don't have time to go into more detail right now but I wanted to flag that possibility.

PR #1844, opened 2025-08-05
Add Lua code linting with luacheck to project GitHub Actions

colinmcardell:
_what?_

This PR introduces automated Lua code quality checking using [luacheck](https://github.com/lunarmodules/luacheck) as a new parallel job in the norns  GitHub Actions workflow. When a PR is submitted or there is a push to main, the CI will now lint the Lua code.

_why?_

Luacheck catches common Lua programming errors early:
- Typos in variable names (e.g. accessing `prams` instead of `params`)
- Using undefined globals (forgetting to require a library)
- Unused variables and dead code
- Inconsistent coding patterns

_how?_

This PR takes a deliberately conservative approach to avoid disrupting development. Configuration of the linting is done through the `.luacheckrc` file, where I have added norns Lua globals, ignores (which tell luacheck to ignore certain violations), and exclude paths.

- **Limited scope**: Currently only lints `lua/engine/` (2 files) with zero warnings
- **Ignores**: Temporarily ignoring common patterns, like unused callback parameters
- **Exclusions**: Skips `lua/core/`, `lua/lib/`, and `lua/extn/sky/` for now

_next_

This PR opens up a path for gradual improvements:
1. Expand luacheck to cover core norns modules (`lua/core/`).
2. Expand luacheck to cover `lua/lib/`.
3. Address `lua/extn/sky` framework separately?

Getting luacheck to pass on the entirety of the `lua/` codebase will require touching almost every file, addressing a lot of white space issues, but plenty of violations related to the ignore pattern I've put in place within the `.luacheckrc` file. The plan is to follow up with a separate PR for each of these phases.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -2,9 +2,9 @@ name: compile
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
   workflow_dispatch:
 
 jobs:
@@ -26,12 +26,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: install luacheck
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y lua5.3 liblua5.3-dev luarocks
+          sudo luarocks --lua-version=5.3 install luacheck
       - name: lint lua code
-        uses: addnab/docker-run-action@v3
-        with:
-          image: robbielyman/norns-ci:latest
-          options: -v ${{ github.workspace }}:/norns-build
-          run: |
-            apt-get update -qq && apt-get install -y luarocks liblua5.3-dev lua5.3
-            luarocks --lua-version=5.3 install luacheck
-            luacheck lua/ --codes
+        run: luacheck lua/ --codes


Dewb:
Also, `lua5.3` and `liblua5.3-dev` are already part of the base Docker image, so I don't think it's necessary to install those in the running container.

colinmcardell:
Thanks for the detailed feedback!

_`runs-on:`_
I think `ubuntu-latest` might be the right choice for the linting job since we're analyzing Lua source code, and not compiling for the ARM target platform. _I believe_ the issue was that the robbielyman/norns-ci:latest Docker image is specific to ARM, which was causing an architecture mismatch.

I've switched the lint job to use native Ubuntu package installation instead of Docker, which eliminates the architecture constraint. What do you think?

_expansion approach_
The per-path overrides idea for `.luacheckrc` is excellent.
It looks like there is support for inline warning override support as well! - https://luacheck.readthedocs.io/en/stable/inline.html


_`luacheck` install method_
I tried getting `luacheck` to install using `apt-get` and had no success without luarock.

Dewb:
Yes, the docker image (defined in https://github.com/monome/norns/blob/main/Dockerfile) is built for ARM64, so it mimics the on-device RPiOS environment as closely as possible. There are some advantages to using the full image (using the same lua version as the device, for example) but I agree it's not nearly as important for the linting step as the compilation step, and it appears to be slightly faster (though not as much faster as I would have thought.) Either solution seems fine to me!

Dewb (approved):

PR #1847, opened 2025-08-08
Fixes .clang-format syntax

colinmcardell:
_what?_

Fixed incorrect syntax in `.clang-format` configuration file on line 94.

_why?_

The `RawStringFormats` section had invalid syntax (`Delimiter: pb` instead of `Delimiters: [pb]`), preventing clang-format from running properly.

_therefore_

clang-format now works correctly with the `clang-format.sh` script for formatting C/C++ code in the project.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -91,7 +91,7 @@ PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
 PointerAlignment: Right
 RawStringFormats:
-  - Delimiter:       pb
+  - Delimiters:       [pb]
     Language:        TextProto
     BasedOnStyle:    google
 ReflowComments:  true
@@ -111,4 +111,3 @@ SpacesInSquareBrackets: false
 Standard:        Cpp11
 TabWidth:        8
 UseTab:          Never
-...


Dewb:
Presumably the existing .clang-format was last used with an older version. Which version were you using that required this syntax change? The norns image doesn't appear to include clang-format by default so we have no documented "official" clang-format version, it might be good to add that to the README. The GitHub Actions runner for 24.04 includes 16.0.6, 17.0.6, and 18.1.3 so it should be one of those.

Dewb:
(it's also possible this was always wrong!)

colinmcardell:
Sorry for the quick fix with no explanation. I've updated the PR message.

It's possible this syntax error would never have worked properly with any clang-format version.
It's also possible newer clang-format versions became stricter.

- Tested with clang 20.1.7 - failed without fix, worked with fix
- Tested with clang 18.1.8 - failed without fix, worked with fix

I can test for earlier versions of clang-format if we think it's necessary.
If the current build Action is using 18.1.3, I can try getting that working locally and test it with the fix, if needed.

It looks like llvm-toolchain-18 for Ubuntu 24.04 (Noble) version 18.1.8 [is the latest](https://launchpad.net/ubuntu/+source/llvm-toolchain-18).

Dewb (approved):

Dewb:
Confirmation of 18.1.x is close enough IMO!

PR #1850, opened 2025-08-09
Run clang-format on all C/C++ source

Dewb:
60% of the initial changes were just breaking long lines, as it appears many files have enforced a column limit lower than the one in .clang-format. So I (temporarily?) disabled line length enforcement to make it easier to review the other changes. 

I can confirm this builds in the norns-ci container, and I reviewed every change, but I have not tested on the hardware yet.

Recommend viewing the changes with Hide Whitespace enabled, there are only about 500 changed lines in that mode.


tehn (approved):
looks great!

the space after if statements made me blink but i will survive!

Dewb:
Yeah, I raised an eyebrow at some of the choices too, but that's the beauty of deferring style to a config file, whatever it says goes!

Builds and runs on hardware with a quick awake test, so I'll merge this to unblock #1851.

Issue #1824, opened 2025-04-06
GitHub Action for 'pages-build-deployment' pointing to wrong directory.


tlubke:
```
...
/usr/local/bundle/gems/jekyll-sass-converter-1.5.2/lib/jekyll/converters/scss.rb:86:in `chdir': No such file or directory @ dir_chdir0 - /github/workspace/docs (Errno::ENOENT)
```

A few lines before it shows the repo mounted to `/github/workspace/`:
```
...
"/home/runner/work/norns/norns":"/github/workspace" ghcr.io/actions/jekyll-build-pages:v1.0.13
```

But, the doc directory is actually at `norns/doc` and **not** `norns/docs` which would appear to be the reason for this failure. The action itself looks new and possibly still under construction?

Dewb:
This is a built-in Action triggered by the Pages settings in the repo; unfortunately the only options for this action are `/` and `/docs`. https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

Dewb:
I changed the source in Settings > Pages from "main: /docs" to "None" to disable the build failure indicators for now.

PR #1854, opened 2025-08-12
Cleanup vestiges of Travis

Dewb:
Maybe we want to wait on removing these breadcrumbs on how ldoc used to be invoked until there's an adequate replacement? As I said in #1824, I'm not up to speed on what the current Pages/ldoc strategy is.


Dewb:
Maybe we want to wait on removing these breadcrumbs on how ldoc used to be invoked until there's anadequate replacement? As I said in #1824, I'm not up to speed on what the current Pages/ldoc strategy is.

PR #1855, opened 2025-08-17
Feature/testing framework lua

colinmcardell:
_what_

This PR proposes a LuaUnit-based, test running setup for automated test discovery, execution, and GitHub Actions integration.

_why_

norns already has LuaUnit-based tests in `/lua/lib/container/` that demonstrate effective Lua testing patterns. This builds upon that foundation to create a cohesive testing framework that can scale across the broader Lua codebase, without introducing additional/external dependencies. The idea is this change should enable more confident refactoring and ensure code quality for the Lua code that powers norns core functionality and scripting.

_how_

A testing framework with the following changes:
- **Updates LuaUnit**: Updates to LuaUnit 3.4 and associated updates to `.luacheckrc` ignores.
- **Mock Library**: Adds `lua/lib/test/mock.lua` providing stub, spy, and table mocking capabilities for isolated testing.
- **Test Runner**: Implements `lua/test/test_runner.lua` with automatic test discovery, self-validation, and reporting.
- **Test Script**: Adds `test-lua.sh` which will run the full suite of Lua tests using the test runner (includes `TestRunner` self-test and mock library tests).
- **Test Organization**: Mirrors codebase structure (`lua/test/lib/container/`) rather than co-locating tests with implementation.
- **GitHub Actions Integration**: Adds `test.yml` GitHub Actions workflow that runs the full Lua test suite on every push/PR.

_therefore_

This provides a foundation for Lua testing in norns, though the specific implementation choices are definitely open for discussion.

Looking forward to feedback, and suggestions for improving this foundation.


colinmcardell:
Thanks. I'm happy with how it's coming together.

A test is handled by LuaUnit by scanning all variables that start with `test` or `Test`. If they are functions, or if they are tables that contain functions that start with `test` or `Test`, they are run as part of the test suite.

Mirror vs. co-location was the main topic of discussion from my perspective, so I'm glad you've raised the questions.

Both approaches work.

Lua and LuaRocks projects usually use a separate test tree that mirrors source. busted and the Olivine style guide expect this by default.

Advantages of mirroring:
- Clear separation of source and tests.
- Keeps runtime paths clean.
- Aligns closely with busted defaults (`*_spec.lua` under `spec/`).

Advantages of co-location:
- Tests live next to the code they cover.
- Easier to find and keep tests in sync with moving or renaming code.

I'm currently working on test coverage for `lua/core/params/`. `lua/test/core/params/binary_test.lua` follows the mirroring pattern and is picked up by the `test_runner.lua` script when `./test-lua.sh` is run.

Note on `require`/`include` paths:

`/lua/core/config.lua` adds `lua/core`, `lua/lib`, … to `package.path`, not `lua/test`. With mirroring, tests stay outside runtime paths. `require` will never pick up `*_test.lua` unless explicitly named, but keeping tests off the path makes it harder to import them accidentally (for example via autocomplete). Mirroring minimizes this class of mistakes.

I would suggest we stick with this mirroring approach as it keeps the tests out of `package.path` so they interfere less with `require`/`include` during runtime/scripting. As mentioned, it also aligns more closely with conventions.

Dewb:
> keeping tests off the path makes it harder to import them accidentally (for example via autocomplete). Mirroring minimizes this class of mistakes.

Since user-scripters dwarf core-lua devs I think this argument is a very strong one.

Dewb (approved):

PR #1856, opened 2025-08-30
fix factory + cm4s headphone and battery

ngwese:
- adds `platform_{factory,shield}` functions to determine hw category
- changes screen flip logic to use `platform_factory`
- allows i2c and battery init if platform_factory == true
- allows i2c_hp if platform_factory == true

Overall this change appears bring a factory norns with cm4s up to functional parity with a cm3/cm3+ device. I would recommend testing on a cm3 based device before merging.


ngwese:
looks like the battery charge/discharge display might not be working on the cm4s still.

Issue #1789, opened 2024-07-08
allow tape looping behavior to be changed


catfact:
this behavior is actually controlled by a runtime flag called `loopFile` within the `Tape` module, but it does not have public setters. i believe adding setter and corresponding OSC handler would be sufficient to toggle looping. 

https://github.com/monome/norns/blob/main/crone/src/Tape.h#L346

might be good to expose this to OSC / lua.

tehn:
this would be good.

it doesn't seem like there's a good place in the PARAMETER menu that makes sense to control this (short of adding a new SUB for tape, which this would be the only option).

alternatively, i propose a minor rework of the TAPE UI:

- E2 (scroll) switches between PLAY and REC (which follows existing UI for up/down)
- K2 cancels an armed PLAY or REC (currently there's no way to cancel, and this would also follow existing conventions.) presently this toggles between PLAY and REC (i don't know why i made this decision)
- E3 in PLAY toggles loop mode

if this seems like a good idea, i can work on this next week

dndrks:
combing through old tickets surfaced: https://github.com/monome/norns/issues/766
@tehn , re: K2 -- could 'canceling' an in-progress PLAY inherit the current STOP functionality, leaving K3 free for PAUSE?

tehn:
the UI opportunity is clear, but the tape module presently doesn't have a PAUSE state so that's an additional feature

colinmcardell:
Fixed by #1857

Dewb:
I haven't used this yet but I'm excited to try it! Thanks @colinmcardell!

Issue #766, opened 2019-03-23
TAPE clears current file on Stop


okyeron:
Just discovering TAPE today and I noticed the following:

- load a file and it cues the file to it's start point
- as soon as I hit "stop" the file is cleared and I need to re-select the file from the menu again

I would find it very useful if you could start/stop/start/stop/etc. the same file without having to reload and jump back to the start each time - oh wow... I just realized I'm asking for PAUSE instead of STOP

Use case: Load a long spoken word sample, play a section, stop (or PAUSE), then play another section, Pause, etc.

Even better would be the capability to scrub thru a Tape file by time index with the encoders.

giovannilami:
I don’t know if to not make a loop is a precise choice, but it's strange since the old Norns' firmware looped tape' samples as default. It would be great to have this again in 2.0

catfact:
no, it's not a deliberate choice, it's from the fact that new tape is made from scratch instead of using supercollider. i'll add the loop feature - it's easy but not trivial. 

(you're right to point out this was old default, since i'd forgotten - thanks!)

okyeron:
@catfact is pause a doable thing as well? 

I like the idea that whatever might be loaded in Tape could stay there "paused" until a new file is loaded.

catfact:
yes, i'll presently add loop (by default), pause (as an option? i'm not sure what the consensus is), and possibly seek (depending on how hard that proves. i think it will be fine.)

giovannilami:
@catfact thanks a lot! ^^
sure, adding pause as an option would be something extremely useful.
ciao!

catfact:
sorry about this one, has just been slipping through the cracks.

i started an ambitious rewrite of tape system with these and other features, but it has been languishing for a while.

PR #1857, opened 2025-09-02
tape: add pause/resume and loop control with threading improvements

colinmcardell:
_what_:
Adds pause/resume functionality for both record and playback. Also exposes loop control API. Includes reworking of tape state machine w/threading safety improvements and additional fixes:

- `pauseRecord(bool)`, `pausePlayback(bool)`
- `setLooping(bool)`, `isLooping()`, `isPaused()`

...

- `/tape/record/pause` → `tape_record_pause()` → `Audio.tape_record_pause()`
- `/tape/play/pause` → `tape_play_pause()` → `Audio.tape_play_pause()`
- `/tape/play/loop` → `tape_play_loop()` → `Audio.tape_play_loop()`

Mostly addresses #1789, #766, and partially address #1758 with an E3 debounce in `fileselect` on preview.

_why_:
Working on exposing tape loop control, I went down a bit of a rabbit hole, and ended up adding pause as well, by way of a command queue between the tape public API and the audio thread. I also addressed some threading race conditions.

_how_:
- Command queue thread-safe tape state transitions
- Safe ringbuffer cleanup and thread management
- Fixed buffer dimensions and underrun handling in reader
- Adds OSC and Lua bindings for new functionality

_notes_:
- This should be a non-breaking API change, fully wired through all three layers (OSC → C bindings → Lua API).
- Lua tape scripting has not been updated to use the new functionality, although the API is fully available. I can follow up with  this but would like to get the core changes reviewed and tested first, as well as discuss how best to update the Lua tape scripts.
- Tested on aluminum norns CM3 hardware and is stable. I will also test on a norns shield Pi3 and shield Pi4 when I get a chance.

### DIFF
--- a/lua/core/menu/tape.lua
+++ b/lua/core/menu/tape.lua
@@ -72,6 +72,53 @@ local function read_tape_index()
   end
 end
 
+--- draw a small loop icon aligned to a right-side x position
+-- @param x_right number rightmost x where icon should end
+-- @param y_center number vertical center of icon
+-- @note size tuned for status row in tape UI
+local function draw_loop_icon_right(x_right, y_center)
+  local w, h, ah   = 11, 7, 4
+  local xl, xr     = x_right - w, x_right
+  local yt, yb     = y_center - math.ceil(h / 2), y_center - math.ceil(h / 2) + h
+
+  local top_y      = yt + 2
+  local bottom_y   = yb - 1
+  local left_x     = xl + 1
+  local right_x    = xr
+  local top_span_x = xr - ah + 1
+  local bot_span_x = xl + ah - 1
+  local head_r_x   = xr - ah
+  local head_l_x   = xl + ah + 1
+
+  screen.aa(0)
+
+  -- frame: top, right, bottom, left
+  screen.move(left_x, top_y)
+  screen.line(top_span_x, top_y)
+  screen.move(right_x, top_y)
+  screen.line(right_x, yb - 2)
+  screen.move(xr - 1, bottom_y)
+  screen.line(bot_span_x, bottom_y)
+  screen.move(left_x, yb - 2)
+  screen.line(left_x, top_y)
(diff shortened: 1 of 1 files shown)


colinmcardell:
I worked out a patch to tape.lua last night. It takes into account the discussed UI/UX changes in #1789. I will amend it to this PR later tonight.

colinmcardell:
Didn't get the tape.lua changes pushed out last night. I will have time tonight!

tehn:
reviewed the code, looks great! still haven't had a chance to test on device.

agreed that it makes sense to maybe not unload the file on EOF. if complete, an issued PLAY command should just play it again, right?

colinmcardell:
> if complete, an issued PLAY command should just play it again, right?

Yes. `tape.lua` is reloading the file and parking at the top, waiting for the PLAY command. This UX is implemented in this PR, works well and feels right.

An optimization would be to prevent the underlying Tape class from unloading the file until explicitly told to. This is not blocking the merge of this work in my opinion.

colinmcardell:
I've add a debounce to `fileselect` preview as mentioned in #1758. It's true that this doesn't resolve the underlying race conditioning in the Tape state, but it does seem to address the issue of Tape getting into a broken state (on aluminum norns hardware) and doesn't hurt to have the debounce in place. Once the underlying race is addressed in Tape it's possible that the debounce time could be reduced.

colinmcardell:
> Playback PLAY/PAUSE/RESUME and LOOP indicators are a bit crowded...

I've updated the UI to address this. I will attach screenshots shortly.

colinmcardell:
01_idle

_idle (no loop)_

02_idle_loop

_idle (loop)_

03_play_loop

_play (loop)_

04_pause_loop

_pause (loop)_

05_rec_mode_loop

_rec mode (playback looping)_

PR #1861, opened 2025-09-21
[WIP] universal gamepad support

p3r7:
this PR provides (almost) universal gamepad support

in its current state, it works pretty well on my norns with the 3 controllers I have tested with.

## how?

this PR removes the use of custom gamepad profiles as lua files and instead relies on the [SDL controller profile database](https://github.com/mdqinc/SDL_GameControllerDB).

i didn't use those profiles in the first place as they were lacking the info of analog sensors resolution (`EV_ABS`, ABS for absolute),

but i discovered that those resolutions could be retrieved with a call to [`libevdev_get_abs_info`](https://www.freedesktop.org/software/libevdev/doc/latest/group__bits.html#gab6393c2d73997b052471a03c8fb05fec).

by combining the 2, we're 100% compatible with our old custom profile logic.

i didn't know this was possible, as SDL itself seems to not look up this info and hard-codes resolution in its code source for a selected number of controllers. e.g.:

image

i guess they do that do be as platform-agnostic as possible...

## what's next to do

everything works fine but i'm getting this error message in the matron logs when i plug a controller. everything seems to work though. idk if it was occurring before my patch or not but i may be doing something wrong.

```
   dev_monitor: skipping node-less entry in /dev/input
   failed to init libevdev (Invalid argument)
   dev_list_add: error allocating device data
```

also i just dropped the SDL gamepad db file under `lua/core/gamecontrollerdb.txt`. but it may be smarter to grab it as a git submodule. what do you think about it?

also i'd certainly merge the `gamepad_db.lua` and `gamepad.lua` files into one.


xmacex:
My dorky little Retrolink controller with only d-pad, select, start, A and B works fine, according to *gamepad-monitor*.



However it smells like chemicals, I should get rid of it.

xmacex:
The only script with gamepad support I published so far was [*serialism*](https://github.com/xmacex/serialism). It was [easy-peasy](https://github.com/xmacex/serialism/blob/main/serialism.lua#L180-L198).

Issue #1478, opened 2021-12-06
Tape breaks when used from script


catfact:
from https://llllllll.co/t/norns-scripting/14120/1032 :

```
function init()
  file = "/home/we/dust/audio/tape/0000.wav"
  audio.tape_play_stop()
  audio.tape_play_open(file) 
  audio.tape_play_start()
end
```

"Playback works initially; however if I then go to Tape and select Play or run audio.tape_play_stop() in matron I am no longer able to hear any audio from tape even after rerunning my script or playing a wav file with the Tape app."

i can reproduce this.  almost certainly it is a bug in the `crone::Tape` state machine. only restarting `crone` process

strongly suspect that adding delays would avoid the bug: such that, e.g. the stop envelope completes before the next file is opened, etc.

this is another reason to perform/complete a full rewrite of the Tape module.

eethann:
I just encountered this working on a tape operator for Orca, definitely interested in it to enable timed triggering of tapes for use during performance, etc. Not sure how useful I can be on debugging, but happy to help how I can.

catfact:
gave this a minute of thought. there isn't really an easy bandaid. what i wish i could do is make all the tape API functions just do nothing if the module is busy (instead of corrupting state) and return a bool to the caller indicating whether the action was taken or not. (of course this is not very convenient for API user but would at least provide some sort of non-borking workaround.) 

however even this kind of weak solution would be convoluted and tricky to implement right now because tape commands are just OSC messages and can't return anything (can only echo responses on a separate return channel.) so i don't think there is much point in addressing this in a useful way before we land the `converged` branch, whereupon lua and the Tape module will live in the same process.

PR #1862, opened 2025-09-21
fix: tape UI drift w/centralized tape state management

colinmcardell:
_what_

Centralizes tape state in the engine to fix UI drift issues. The engine becomes the single source of truth for play/record state, tape filenames, position, and loop status through periodic polling.

_why_

Ensure reliable tape menu UI updates and remove duplicated state tracking.

_how_

- Update `Tape.h` to expose additional getters for observed tape state, and better handling of internal state for observation
- Add tape status polling (`/poll/tape`) file open/close events (`/tape/play|record/open|close`)
- New `audio.tape` observable for centralized core Lua tape state management
- Refactor tape menu to consume centralized state, eliminating local timers and manual redraws
- Add test coverage for tape functionality in `tape.lua` and `audio.lua`

_notes_

- The gamepad functionality is maintained and unit tested, but I did not test this method of interaction with a gamepad. It should be feature parity.
- Cleaned up an unused function in `oracle.c` `handle_tape_play_state`. It was a no-op and nothing references it. I can’t find any registration or calls to `/tape/play/state`. Unused throughout, and should be safe.
- Removed previously added `unload & re-arm` in tape REC mode. This was producing odd logic/state conditions that were hard to reconcile at this time. Options are now `cancel` (go back) or `unload` when K2 is pressed in REC mode with file open.


colinmcardell:
This should fix #1478, and adds events that can be used to fix #1436 (wait until tape record is stopped before sleep).

tehn (approved):
Tested on device and reviewed code. Looks and feels great. New playhead rendering is excellent. Thank you!

tehn:
Also I think it's time for a new beta release so people can try this out. Will get one put together shortly.

PR #1866, opened 2025-11-17
Add screen method to get primary context

Dewb:
Allow mods to get a pointer to the main screen context.

Contrary to speculation in #1749, this is all that's necessary to enable ndi-mod to use the new screen system, though I still plan to eventually propose some things in that direction for performance/flexibility.


tehn (approved):

xmacex:
Apologies for commenting on a closed issue, but on startup, if I have ndi-mod enabled I am getting

```
hook: ndi-system-post-startup failed, error: error loading module 'ndi_mod' from file '/home/we/dust/code/ndi-mod/lib/ndi_mod.so':
	/home/we/dust/code/ndi-mod/lib/ndi_mod.so: undefined symbol: screen_context_get_primary
```

on norns 3.0.0.

colinmcardell:
I believe this is because of the move to C++ (required to unify the binary) which mangles the function names upon compilation. As far as I understand it, there is no mod ABI so any change to the underlying code means the mod will stop working. The solution I can think of is would be to wrap the "public" API(s) in `extern "C"` so the C++ binary exports the unmangled C symbols that mods resolve against. Probably worth tracking as an issue to wrap the matron functions in `external "C"` to maintain the mod interface point.

Dewb:
Agree with @colinmcardell’s analysis. There has never been a compatibility contract for C mods—they have mostly survived new releases so far—but the move to C++ is the biggest break yet. ndi-mod and others may continue working if we demangle all the C exports, but that’s basically all of the former-matron functions; nothing was off-limits. Other possibilities could include defining an official, limited C mod API surface that covers the existing C mods (there aren’t many), or documenting a set of steps for a C++ mod build that is as wide open as the 2.x mod exposure surface was.

PR #1868, opened 2025-11-22
add arc intensity

tehn:
note this requires a libmonome update (which I can bundle in the next norns update) and also arc firmware update (to be released formally).

there are no plans to add intensity to older edition arcs as they are already super dim.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -11,16 +11,23 @@ jobs:
   compile-norns:
     runs-on: ubuntu-24.04-arm
     steps:
-      - uses: actions/checkout@v4
-        with:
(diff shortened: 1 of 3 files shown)


tehn:
thanks for the tip! corrected the Dockerfile but it seems the CI container needs to be recompiled.

libmonome on the actual disk image is manually compiled from source. the entire process is very by-hand. https://github.com/monome/norns-image/blob/main/build-dev-image.md?plain=1#L88

for the next update I'll simply include the source tree and add that step to modify the disk image.

Dewb (approved):

tehn:
thank you @Dewb !

PR #1872, opened 2025-11-30
Allow stacking segments

### LINKED ISSUE #1852: Arc:segment() - skip blank leds?
https://github.com/monome/norns/blob/8e6ac7005aa4d014eef49e093124374c2524a9f2/lua/core/arc.lua#L139

Segment is nice, but it is very final - filling the ring and overwriting any leds you might have set as background. There can also only be one segment on each ring at a time.

If it were to skip blank leds, multiple segments could coexist, layer and have backgrounds.

I think this could be a way to do it:
`
if m[i] ~= 0 then self:led(ring, i, m[i]) end
`

There may be issues with already existing scripts, but segment leans to standalone use and drawing is usually initiated with a clear() - so it might be ok?

imminentgloom:
Fixes #1852 

By default Arc:segment() clears any background leds making it impossible to stack several segments.

Added a boolean flag, "stacking", which defaults to false to preserve original behaviour.
Setting it to true skips drawing any background/0 brightness leds.

### DIFF
--- a/lua/core/arc.lua
+++ b/lua/core/arc.lua
@@ -106,7 +106,8 @@ end
 -- @tparam number from : from angle in radians
 -- @tparam number to : to angle in radians
 -- @tparam integer level : LED brightness in [0, 15]
-function Arc:segment(ring, from, to, level)
+-- @tparam boolean : false (default) to clear background, true if drawing multiple segments
+function Arc:segment(ring, from, to, level, stacking)
   local tau = math.pi * 2
 
   local function overlap(a, b, c, d)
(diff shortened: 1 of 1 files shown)


tehn:
got started, tested the grid single led, works. will finish typing the rest tomorrow

https://github.com/monome/norns/compare/rel-leds?expand=1

imminentgloom:
this sounds great! like that it's a general solution as well

will we be able to do something like this after:
`a:segment(1, from, to, function() return math.random(5) + 1 end)`
which draws an noisy, dim segment without breaks?
i did this inside segment() in another script, and started testing last night without success. if it works or can be made to work it might be an interesting example for the reference docs

tehn:
closed in favor of #1873

PR #1876, opened 2026-02-09
fix: guard ARM-specific code behind NORNS_DESKTOP for x86 desktop builds

blainemotsinger:
The --desktop build mode was incomplete: ssd1322.c (NEON intrinsics), gpio.c, and -mfpu=neon were always compiled, breaking x86_64 builds. Properly exclude ARM hardware drivers and flags when building for desktop.

### DIFF
--- a/matron/src/hardware/input/inputs.h
+++ b/matron/src/hardware/input/inputs.h
@@ -2,6 +2,9 @@
 
 #include "hardware/io.h"
 
+#ifndef NORNS_DESKTOP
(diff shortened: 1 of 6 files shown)


tehn (commented):
Thanks for looking into this.

Would it not be simpler and cleaner to instead put all of the defines inside of ssd1322.c? So all the includes could stay, functions could get called but they'd just be returned.

Have you had success running the desktop build?

PR #1878, opened 2026-02-17
Add nng to dev/ci container to support work on converged branch

Dewb:
Changes:
- build nng from source in the dev/ci container
- make sure user "we" has apt/sudo permissions in the container for debugging
- regularize line endings in Dockerfile
- fix issue with the vs code dev container plugin building for the wrong architecture on ARM macOS

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,7 +5,10 @@
 	"build": {
 		"context": "..",
 		"dockerfile": "../Dockerfile",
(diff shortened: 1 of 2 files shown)


Dewb:
This change breaks one of the clock timing tests, which is curious. I will split the nng change and the devcontainer arch fixes to narrow down the cause.

colinmcardell:
I will take a look shortly, thanks.

I will say that the GH action environments have made these tests flaky in the past. I will let you know what I find.

PR #1886, opened 2026-03-02
softcut: add linear resampling on buffer import

### LINKED ISSUE #1884: softcut: add resampling on import
`buffer_read_stereo` and related methods do not attempt to take the source sample rate into account. however, the function arguments are given in seconds, somewhat implying that resampling will occur. the non-resampling behavior should be spelled out in the [API reference](https://monome.org/docs/norns/api/modules/softcut.html#buffer_read_stereo) at least. 

even better would be to just add resampling on file import, which is not difficult. options and considerations:

- link `libsamplerate` and use [that API](https://libsndfile.github.io/libsamplerate/api_full.html). (i am sure the runtime library is included on the norns image already, but not sure about the `-dev` package.)

- roll a simple resampler using linear or cubic interpolation. (main reason to consider linear, IMHO (besides simplicity,) is that polynomial  interpolation can cause small amplitude overshoots, which is a problem when importing 0db-normalized samples. so kind of necessitaties an optional downward-normalization step.)

jackby03:
Updates `crone/softcut` submodule to the commit that adds linear resampling when importing audio files whose sample rate differs from the engine rate (48 kHz).

See the submodule PR for full implementation details: monome/softcut-lib#72

Closes #1884


Dewb:
I think it would be great to get this merged. But I agree with @schollz's comment in https://github.com/monome/norns/issues/1884#issuecomment-3986261194 that the current implementation is a breaking change that comes with costs for script authors and users. 

To restart discussion, I'll make a straw man proposal to avoid breaking scripts:

1. The softcut functions that gain the resampling ability also get a new `resample` boolean argument controlling the new capability, which can default to true
2. The existing norns Lua API functions `softcut.buffer_read_mono` and `.buffer_read_stereo` (and any other affected API methods) pass `resample=false` in their C implementations
3. The API gains new functions `softcut.buffer_resample_mono` and `.buffer_resample_stereo` that are implemented with `resample=true`
4. The old methods `buffer_read_mono` and `buffer_read_stereo` are flagged as deprecated

This has some downsides:
- "resample" is not as clear or friendly a name as "read" to inexperienced users. 
- The softcut Lua API names diverge from the underlying native API.

So there's room for improvement. Another possibility is to keep the `buffer_read_x` names in the Lua API, but add the resample argument to the Lua layer and have it default to false to avoid breaking changes. However, I do think in the long run most users do want to resample, so making that the default for new scripts seems friendlier.

PR #1893, opened 2026-05-26
converged 2026

tehn:


tehn:
ok working again on a clean compile from a "stock" norns, intending to merge to main and mark this as 3.0.0

@colinmcardell i am not seeing libnng under third-party, so i added it, but i it seems the wscript doesn't pick it up. in the previous PR you mentioned it was set up to use either system installed nng or the third-party, but it seems the second option is failing for me still. perhaps i missed a step?

symptom is: even if compiled on a norns that has the converged branch working (likely due to system install of nng) the "stock" norns fails as ws-wrapper is looking for a shared lib and not finding it

```
May 26 18:03:21 norns ws-wrapper[1411]: /home/we/norns/build/ws-wrapper/ws-wrapper: error while loading shared libraries: libnng.so.1: cannot open shared object file: No su>
```

tehn:
side issue is that CI is failing because i don't think it knows about the new json lib

colinmcardell:
Oh nice. Progress!

Regarding libnng, we'll have to add it as a system dependency installed as a package through the update.sh. I backed out of the submodule approach because it wasn't working properly.

colinmcardell:
sudo apt-get update && sudo apt-get -y install libnng1 libnng-dev

This should work.

tehn:
thanks for the reminder, I lost track of that. no problem!

-------- Original Message --------

colinmcardell:
https://gist.github.com/colinmcardell/3b8982327c009c2c74f7e2a0dff22cab

This has all the things 🙂

colinmcardell:
> side issue is that CI is failing because i don't think it knows about the new json lib

I can whip up a patch for the lua linter warnings.

tehn:
ok, the beta installer works! it required a couple extra steps as the apt sources are so old! the update script is pretty patchwork at this point. thinking going forward this might be a cusp where update happens _to_ 3.0.0 and then future updates assume that the system has already been "fixed" to this point.

i'm thinking this should be good to merge to main, but i'd suggest we do a beta trial for a week or so for any needed hotfixes prior to listing the update on the stable channel.

also still pretty excited about 6 vu's on the mixer screen, wow

colinmcardell:
> > side issue is that CI is failing because i don't think it knows about the new json lib
> 
> I can whip up a patch for the lua linter warnings.

Ah, easy enough: https://gist.github.com/colinmcardell/c559a740a41374f32a8c3e83553ad9c5

tehn:
> > > side issue is that CI is failing because i don't think it knows about the new json lib
> > 
> > 
> > I can whip up a patch for the lua linter warnings.
> 
> Ah, easy enough: https://gist.github.com/colinmcardell/c559a740a41374f32a8c3e83553ad9c5

fix pushed!

tehn:
ok! we have a release available through SYSTEM > BETA now

tested on hardware.

xmacex:
Hello. I don't know if this is helpful, but on DIY norns shield running latest release, this beta update

- [x] Downloaded
- [x] Unpacked
- [x] Ran the install (a status indicator might be nice to have)
- [x] Restarted
- [x] One or two scripts which use a SuperCollider engine work ok
- [x] Some fx mods which use a SuperCollider engine work ok

Is there something I could help testing?

tehn:
@xmacex thanks for testing! we've done a lot of testing already but there's so much surface area that any additional "normal use" reports would be much appreciated.

xmacex:
Hmm did cranes start running not continuously but by seconds, or did it always do so? I am looking at both the timestamp numbers on the top left, and the flight animation and I think the numbers were running continuously with one decimal and the birds were flapping faster... or am I imagining it?

tehn:
can't comment on that script directly, i have no idea

xmacex:
Thinking aloud: something something do the softcut phase polls run at the speed set/requested by the Lua library, (in cranes, `softcut.phase_quant(i, 0.01)` on line 122) or once per second (crone/src/SoftcutClient.cpp `setPeriod(1)` on line 46). More thinking aloud later: I think that number is milliseconds, not seconds. Even more: [what friendly internet people are saying](https://llllllll.co/t/cranes-v2-3-241112/21207/309?u=xmacex) supports that something there might have changed, about the polling I am still guessing. I will try to find other scripts which use softcut polls (and learn what they even are). More: nmRain which is a softcut delay used to run continuously and now updates at 1Hz. I'll look if softcut studies also behave like this, after disabling all my mods...

sonocircuit:
thank you all so much for the continuous effort! this is a huge update and the vu meters are such a joy to look at.
I can confirm that the softcut phase poll is not working as intended as well as the waveform rendering. I installed the latest beta and ran mlre as well as concrete. In both cases the playhead visualisation is updated slowly (producing jumps) and the waveforms are only partially displayed.

tehn:
thank you for flagging this. i'll get some test cases together and confirm.

@colinmcardell are there any known limitations in the current implementation?

petercolombo:
hey all - new to posting feedback here, but I'm on the newest beta and enjoying it a lot -

noticed that there seems to be an alternative font buried in the new OS? i'm getting a whole new system font when loading certain scripts, but it shows up inconsistently.

it will load / reload the stock norns font when entering and exiting the PARAMS page, whether it's via my scrolling to the right via E1 or pressing F4 on my keypad. by "load / unload", what i mean is that the new font will appear for the first time while accessing PARAMS, carry across the other norns pages (tape, mixer etc), and will only reload to the stock font when i go to the currently running script screen (equivalent of F5 input), then going to the home screen (equivalent to F3 input). scrolling to the home screen from Params will keep the "new" font loaded across all screens.

my question is.....is there now a way to load custom fonts onto norns?? is there a way to have finer-grain control over this? i love the look of the new font, but would love to toggle between the two consistently. 

thank you all for your lovely work on this platform - the new VU meters are so snappy and great. what an incredible upgrade!

tehn:
i believe you've discovered a bug-- there was no new intended font selection feature.

do you have specific steps (and scripts?) to replicate this situation that we can try to reproduce?

thanks for testing!

petercolombo:
I first encountered this bug using Seeker II - had Norns and Grid connected to Crow via USB, sending note information to JF. went to PARAMS and was met with this: 

IMG_3531

I was able to revert to stock Norns font using F5 to go back to the script screen, then F3 to go to the Home Screen directly. When accessing Params directly via F4, this font reappears.

colinmcardell:
> thank you for flagging this. i'll get some test cases together and confirm.
> 
> 
> 
> @colinmcardell are there any known limitations in the current implementation?

I will have a look and see how things are wired up (or not 🙂). It's possible that softcut phase poll didn't make the migration.

colinmcardell:
> I first encountered this bug using Seeker II - had Norns and Grid connected to Crow via USB, sending note information to JF. went to PARAMS and was met with this: 
> 
> 
> 
> IMG_3531
> 
> 
> 
> I was able to revert to stock Norns font using F5 to go back to the script screen, then F3 to go to the Home Screen directly. When accessing Params directly via F4, this font reappears. 

Thanks for the details. This is very helpful. I can take a look into this as well. It seems possible that it's an issue with the font not getting set appropriately when toggling between the script screen and the menus screen.

tehn:
i'm actually not sure which code would've changed in the converged branch that would've impacted font/menu/screen at all? i will attempt to replicate.

petercolombo:
I found, when using the hardware Norns inputs over my keypad, that scrolling across from Params via E1 would have the font carry across all screens, not just Params: entering into the script screen, exiting back to home via K1 brought back the stock Norns font.

xmacex:
> I installed the latest beta and ran mlre […] the playhead visualisation is updated slowly

Yes happening on mlre. If I rise an octave ie. double the playback rate, also the UI update seems to run at double time.

colinmcardell:
Apologies for the delay. Think I found a few issues.

The phase poll issue is a missing entry in the new param dispatch map. The truncated waveforms seem to be because of a bug in the render callback. I'm still looking into the font issue... it does seem like a pretty odd one.

I will try to open a PR this weekend after some testing on hardware.

xmacex:
Well done spotting the cause. Don't hesitate to let me know @colinmcardell if I can help test something.

---

Meta: is this the good way for us to give feedback on beta norns, here in a thread, or should we do it otherwise e.g. by rising issues, or some way which would be best for devs/maintainers?

colinmcardell:
I've opened this PR https://github.com/monome/norns/pull/1896 which should address the softcut param dispatch map and render callback issues.

tehn:
fantastic, on my list for tomorrow and can post up a new beta. thanks tons!

colinmcardell:
I've been digging deeper into the menu font issue... and generally bouncing back and forth reading through the pre<->post converged seams and changes (it touches a lot 🙂). I think there are a few things worth addressing.

If it's easy enough to make a new beta and then later follow up with another, merging #1896 now and then following up with some additional fixes would be great. I will try to get some additional fixes in shortly.

- [x] `norns.system_cmd` and `util.os_capture` "regression" - errors silently drop the `system_cmd` callback and leaks, and makes `os_capture` return nil. As far as I understand it, system commands were inherently async by way of OSC, and now they block + don't maintain the same "contract" of returning empty string. I'm actively digging into this... it's a decent chunk of rewrite work to make sidecar async, but I have some work done in this area already in getting SoX functional and threaded. I should be able to peel it apart (or push it to the finish line if we want to add SoX now-ish).
- [ ] mods not working - converged switch to C++ which mangles the function names upon compilation causing `undefined symbol` error. This seems straight forward to wrap the matron function in `extern "C"`
- [ ] menu font issue -  I _think_ it's actually a latent bug that has been surfaced by the switch to the converged binary. It seems like a race condition. I'm trying to reproduce with a little script, and will fix/report back.

I'm open to feedback on priority and thoughts in general of these issues. I've got them listed the way that I'm ranking them at the moment.

In the meantime, let's keep testing things with normal usage, and let me know if you experience or notice any other issues.

tehn:
@colinmcardell thank you for looking into this! I'm building the new beta now.

@ngwese might have some insight around the sidecar issue, as he authored that aspect (though, it's been awhile I admit)

tehn:
@colinmcardell I've just discovered a new bug related to `system_cmd` which you point out above: since the SYSTEM > UPDATE mechanism uses `system_cmd` (wget etc) with a callback, somehow this all fails spectacularly (by spectacularly I mean it stalls (blinks forever) so maybe actually not so spectacular)

But, thanks tons for the softcut fixes!

There's a new beta posted up but now not a smooth way to update from the existing beta (it should work fine from stable).

For those people on current beta wanting to upgrade, you'll need to do it manually:

ssh/screen into norns via cli then:
```
cd update
wget https://github.com/monome/norns/releases/download/v3.0.1/norns260616.tgz
tar xzvf norns260616.tgz
cd 260616
./update.sh
sudo shutdown now
```

sonocircuit:
Installed norns260616 and the softcut polls and wave rendering are working as expected. thank you so much! please let me know if there is anything in specific I could test or help in any other way.

@colinmcardell regarding the vu meters I have a minor suggestion. would it be possible to scale the monitor vu with the monitor level? otherwise it simply mirrors the input vu.

colinmcardell:
> @colinmcardell I've just discovered a new bug related to `system_cmd` which you point out above: since the SYSTEM > UPDATE mechanism uses `system_cmd` (wget etc) with a callback, somehow this all fails spectacularly (by spectacularly I mean it stalls (blinks forever) so maybe actually not so spectacular)

This makes sense. Ha. Ok. There is a pretty natural boundary presenting itself between the norns application itself (now a single binary) and the system level stuff (a new binary hanging out to do system/environment related work). Honestly, the restart command is similar and has some pretty flaky stuff to get it to "work". It kind of all wants a service like the sidecar to delegate to. I need to think on this for a bit. 🤔


> But, thanks tons for the softcut fixes!

No problem! Excited things continue to progress.

colinmcardell:
I have a repro and fix for the menu font issue. Will open a PR soon.

xmacex:
I also did the dance above and installed the new beta. I did get this from the `./update.sh`, as the last output before it exited.

```none
rm: cannot remove '/home/we/matronrc.lua': No such file or directory
'/home/we/maiden/dist/sources/base.json' -> '/home/we/dust/data/sources/base.json'
'/home/we/maiden/dist/sources/community.json' -> '/home/we/dust/data/sources/community.json'
./update.sh: line 97: cd: ./package/libmonome: No such file or directory
./update.sh: line 98: ./waf: No such file or directory
sudo: ./waf: command not found
```

Restarted, and softcut scripts I tried earlier namely Cranes, nmRain and softcut study 9 query work fine and update in usual speed rather than what seemed like 1Hz earlier (although the flapping animation in Cranes looks kind of weird).

colinmcardell:
> I have a repro and fix for the menu font issue. Will open a PR soon.

I'm actually going to back out on this for the moment.

I'm able to reproduce the issue. And I can patch this and that... but it's deeper than simply setting the font when the menu is displayed. It's an issue of screen "ownership", and not a regression as far as I can tell. A test given the same conditions with Seeker II on the latest non-converged release might be informative?

Also, not a race condition because there is no threading, nothing to race. It's that a script can run a metro that sets the font (or any screen/Cairo surface param/state) and the menu display will be affected.

The fix would be a refactoring to decouple how the screen and the surface is drawn to. The script should get its own surface. The menu pages, screensaver, and system overlays (e.g. battery warning), should likely get a surface to draw to as well.

Transformations if imbalanced or not managed correctly, carry over to the menu as well (rotating the screen in the script, results in a rotated menu).

Overall a rather complex thread to pull on at the moment, and I would rather table it for now and focus on landing the work that has changed.

Deprioritizing the "menu font issue" for now.

colinmcardell:
> @colinmcardell regarding the vu meters I have a minor suggestion. would it be possible to scale the monitor vu with the monitor level? otherwise it simply mirrors the input vu.

I can look into it. You are basically suggesting pre/post fader metering capabilities?

tehn:
> Also, not a race condition because there is no threading, nothing to race. It's that a script can run a metro that sets the font (or any screen/Cairo surface param/state) and the menu display will be affected.

you're correct about this. as you state, there are various ways in which a script can mess up the menu by calling screen functions outside of the suggested/required `redraw` function. (i haven't confirmed that the seeker script is to blame, but this lines up with previous issues we saw back around launch time (8 years ago? wow)

i apologize that this was likely not a regression and an unnecessary sidequest. i agree that it should be shelved.

colinmcardell:
I'm working through solving the sidecar command/update/restart issue(s). The update portion is launch gating and I'm working through what might make that the most robust. There is a gap across the stack (Lua, C, script execution method) and I need to thread it all through. 

Asynchronous/streaming sidecar communication seems to be another pretty clear win to include with the v3 release. I'm bundling the work together.

I will spend some time this weekend and have tested code to PR with solutions and fixes. 🤞

sonocircuit:
> I can look into it. You are basically suggesting pre/post fader metering capabilities?

thanks for looking into it. what I meant was to have the monitor vu display the post fader value like all the other vu meters. currently it's mirroring the input signal.

colinmcardell:
A couple of updates:

- I have changes that solve the update issue. It's a little messy at the moment. But validated locally by adding support for overriding the release url, which has been helpful. I've also made the update fail/back-out if an error is encountered, instead of updating in place and leaving the system in a potentially broken state. I'm trying to find time to tidy up the changes over the next few days.
- system commands and the sidecar has been refactored to fix a few failure/edge cases, supporting all the things needed for the update, generally more resilient status and error handling, and supports asynchronous.
- I looked into the mods function mangling issue but haven't had a chance to build and validate. I will before pushing the update.sh work.
- I've read through the code around the monitor VU. This will be an easy enough patch... it's an outlier from the others, and the issue would be best served with a future refactor post converged release, to centralize the behavior into a shared "channel strip".
- as previously mentioned, not pursuing the screen ownership font issue.

I think that's all the things. Will continue to update. Just wanted to make sure to keep the comms open.

Dewb:
I can take a look at the menu screen context issue this week. Probably not a new issue, but perhaps the converged architecture makes losing the race condition more likely.

colinmcardell:
> I can take a look at the menu screen context issue this week.

Awesome. Sharing where I was at in the investigation.

The mega version of the fix seems to be to build `n` different screen contexts at the C level so that the script, menu and overlay system each has its own surface to write to,then a toggle that indicates which surface is writing to the actual display.

A possible lighter version might be to set all of the things that can be changed on the context (font, anti-aliasing, and transformations/rotations), within the function that toggles the menu into view, along with some method to reenforce in a script that calling screen methods outside redraw will produce unexpected results. The transform/rotate part requires exposing a C function to reset the context to identity. I'm not certain how we might reenforce the screen methods to keep them inside redraw, which is what lead me to the mega version.

Dewb:
Closing the loop on the menu font issue:

I think for a few different reasons it would be good to do as @colinmcardell describes and completely separate the menu context/buffer from the script context/buffer. Doing so would allow screen.peek to work correctly on the first frame after a menu toggle, would allow feedback-based effects to continue uninterrupted, and would allow things like ndi-mod or a SDL output to display the menu and script graphics on separate surfaces.

But that's not necessary here. There is no race or bug in the menu system font selection, it's all fine. The observed behavior comes because the seeker_ii specifically overwrites the core menu's drawing routine.

image

tehn:
> But that's not necessary here. There is no race or bug in the menu system font selection, it's all fine. The observed behavior comes because the seeker_ii specifically overwrites the core menu's drawing routine.

wow, thanks claude!

i agree that while a split screen context would be rad, it's a bit outside the immediate scope.

even further outside the scope, years ago i strongly considered implementing the entire menu/management system in C, just to define a clearer separation between scripts and the infrastructure. the introduction of the mod system made this pretty unrealistic without breakage. (sorry for the distraction, this is all a historical aside)

in other news i'm available for further testing!

colinmcardell:
> Closing the loop on the menu font issue:
> 
> I think for a few different reasons it would be good to do as @colinmcardell describes and completely separate the menu context/buffer from the script context/buffer. Doing so would allow screen.peek to work correctly on the first frame after a menu toggle, would allow feedback-based effects to continue uninterrupted, and would allow things like ndi-mod or a SDL output to display the menu and script graphics on separate surfaces.
> 
> But that's not necessary here. There is no race or bug in the menu system font selection, it's all fine. The observed behavior comes because the seeker_ii specifically overwrites the core menu's drawing routine.

Great sleuthing. All makes sense and lines up. Happy to contribute, once the dust settles with this line of work.

I am getting close to a PR. I have been chipping away each evening. The surface area of the system commands/sidecar is spread across a good amount of norns and so I've had to go through and patch a few things here and there. The PR will be a bit hefty in the sidecar area... all great discoveries and improvements throughout. Stay tuned.

> even further outside the scope, years ago i strongly considered implementing the entire menu/management system in C, just to define a clearer separation between scripts and the infrastructure. the introduction of the mod system made this pretty unrealistic without breakage. (sorry for the distraction, this is all a historical aside)

❤️ the distraction is great!

This converged work is definitely creating momentum in a direction that lends itself nicely to this idea of pushing whatever core Lua code down a layer... also building C interfaces that separate away the specifics of the underlying Linux system.

colinmcardell:
I am wrapping up the work on sidecar/system commands, and the update path.

Curious for some feedback as I'm working through an issue with `norns.system_cmd(cmd, callback)`, which never calls the callback when the command outputs nothing (e.g. `mkdir`, `rm`, an `ls` with no matches), so callbacks chained off silent commands don't run, and the Lua callback reference leaks each time.

The reference leak I'll fix. The real question is the callback contract... keep the behavior where the callback doesn't run on empty output (compatible, but the chained callbacks break with no error), or have it always fire once with an empty string (predictable and consistent with every other async callback, but a visible change to the scripting contract).

---

Update: I answered my own question. Pre-converged `norns.system_cmd(cmd, callback)` returned empty string, so the fix is patch the leaking Lua ref, and always fire with "", which will restore the pre-converged scripting contract. 🪄

tehn:
awesome, this is indeed the behavior that would preserve backwards compatibility. thank you!

tehn:
I have some availability this week for testing--- anything I can help with? Looking forward to putting out another beta build.

colinmcardell:
Yes. Thanks for the nudge. I have been making progress. I feel good about a PR to test this week.

Fair warning: it's a good amount of change to the sidecar and the update.sh. I've had to touch other system level functionality to gracefully delegate them through sidecar and get it out of the way of the converged binary which now holds the JACK thread and must not fork. It will need testing on the update process end to end, and testing to confirm that all of the system level function that has been touched is continuing to work correctly.

If it's look and feeling alright, we can put together a beta release which will require a manual update. Then we can follow up with the mods fix and the VU meter tweak as another beta release that can test the update process.

PR #1895, opened 2026-06-14
fix: add default encoder sensitivity and acceleration

n8-n:
Hi all. Here's a small fix to allow users to define the default knob sensitivity.

Details:
My Norns Shield has "jumpy" encoders, and on Lines I found a few threads that discussed ways of changing the knob sensitivity and acceleration, e.g [this thread](https://llllllll.co/t/norns-shield-jumpy-encoders/44292/2). Changing the values didn't seem to make any difference, so I dug into the code and saw that those initial values aren't being used as defaults because they are overwritten in other parts of the code.

The main issue is that the enc.accel and enc.sens script state values are being set in the [Script.clear() function](https://github.com/monome/norns/blob/main/lua/core/script.lua#L66), which is called whenever a script is loaded. Values in encoders.sens and encoders.accel are modified in menu.lua and mix.lua, so they can't be used as default values.

With the fix, I'm basically just adding explicit default values and setting the script state to use these values when Script.clear() is called.

### DIFF
--- a/lua/core/encoders.lua
+++ b/lua/core/encoders.lua
@@ -11,14 +11,22 @@ local util = require 'util'
 
 local now = util.time()
 
+local function default_accel_and_sens()
+  local default_sens = {2,2,2}
+  local default_accel = {true,true,true}
+
+  return default_accel, default_sens
+end
+
+
 encoders.tick = {0,0,0}
-encoders.accel = {true,true,true}
-encoders.sens = {1,1,1}
+encoders.accel, encoders.sens = default_accel_and_sens()
 encoders.time = {now,now,now}
 encoders.callback = norns.none
 
 --- set acceleration
 encoders.set_accel = function(n,z)
+  print("Modifying encs in [norns.encoders.set_accel]: n="..n..", z="..tostring(z))
   if n == 0 then
     for k=1,3 do
       encoders.accel[k] = z
@@ -32,6 +40,7 @@ end
 
 --- set sensitivity
 encoders.set_sens = function(n,s)
+  print("Modifying encs in [norns.encoders.set_sens]: n="..n..", s="..s)
   if n == 0 then
     for k=1,3 do
       encoders.sens[k] = util.clamp(s,1,16)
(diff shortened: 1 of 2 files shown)


lua/core/encoders.lua
```diff
@@ -82,11 +89,15 @@ encoders.process_with_accel = function(n,d)
 end
 
 
+encoders.print_state = function()
+  for k=1,3 do
+    print("encoders | sens="..encoders.sens[k]..", accel="..tostring(encoders.accel[k]))
+  end
+end
```
n8-n, line 96: I can remove these print functions if they're not needed

tehn:
reviewing the code, I cannot tell that this code does anything more than what the existing code does.

These existing lines:

```
norns.enc.accel(0,true)
norns.enc.sens(0,2)
```

seem to do the same thing as the proposed code. Index 0 means set all three indices.

If you wanted to globally change your sensitivity, just edit those lines above. For legibility we could add

```
local default_accel = true
local default_sens = 2

norns.enc.accel(0,default_accel)
norns.enc.sens(0,default_sens)
```

But beyond that the proposed change is just a lot of extra code.

n8-n:
Sure, fair enough. I wanted to keep the logic in encoders.lua so that it would be obvious for anyone else who wanted to modify the default sensitivity.

xmacex:
These kinds of things make me always wonder about the [*Settings* menu](https://monome.org/docs/norns/play/#system) in norns, and what is these and what could be there.

tehn:
I think more the problem is that right now the encoder sens/accel is set as an absolute value by scripts. For the desired functionality I think what's needed is not a default value, but a multiplier/scalar. Default would be 1, but increasing this would "slow" the encoders and require more accumulation which may result in some de-glitching.

Encoders are better fixed in hardware, I might add. No amount of software is going be more than a not-great bandaid.

xmacex:
Indeed – I won't forget the day I got my jumpy encoders replaced (in hardware) by a friendly stranger.

n8-n:
Yeah, a hardware fix seems like the way to go. I also saw on Lines that some people had success with adding extra capacitors to slow encoder jumps. I just need to learn how to solder

PR #1900, opened 2026-07-10
Remove -mpfu=neon flag for desktop build

RadicalTeapot:
This is an ARM-only flag which should be removed when targeting desktop (x86_64) builds. The current behavior is kept for non '--destkop' builds.


ngwese:
took a quick look. would recommend checking if DEST. _CPU in the waf build env is set to ‘arm’ or ‘aarch64’ to determine whether that flag is appropriate

desktop builds aren’t necessarily x86, then could easily be arm on mac or other desktop linux machines

PR #1903, opened 2026-08-10
sidecar: typed streaming transport + detached exec, and a fail-safe SYSTEM>UPDATE

colinmcardell:
_why_

with the current converged binary, updates and system commands such as restart are delegated to the sidecar. update fails completely and restart fails often... requiring manual reboot of the device.

update, restart, and shutdown all need privileged, detached execution. the existing update flow runs `tar` and `update.sh` as concurrent fire-and-forget shell calls, racing each other, then shows "done" before either has finished. when an update step stalls, the screen is stuck on its last message forever.

since the sidecar is norns' child process, any command that restarts norns also kills the sidecar and with it the command doing the restarting.

_what_

this PR reworks the sidecar around a typed, streaming transport with message framing (`norns/sidecar_msg`), line streaming (`norns/sidecar_lines`), and supervised execution (`norns/sidecar_shell`) as pure, separately tested pieces, and `sidecar.cpp` keeping the NNG glue. additionally, the sidecar gains the ability to launch detached system commands is added, which keep running after norns stops.

the SYSTEM>UPDATE flow is then reworked (using the sidecar transport improvements) into a fail-safe / fail-loud updater that replaces the norns binary and reboots only once the new binary proves launchable, rolling back otherwise. and importantly no longer stalls and fails spectacularly :).

sidecar's protocol can now launch work that outlives norns replacing/restarting itself, so we also got a much more reliable restart mechanism.

_highlights_

- sidecar - a stuck command can no longer tie up the sidecar (`norns/sidecar_shell`). commands run with limits. a command that goes quiet or floods output is cut off (each request sets its cap), stderr streams back alongside stdout, and stopping a command asks it to exit before forcing it.
- lua shell - lua's `os.execute` and `io.popen` are reimplemented over the sidecar, closing the last `fork()` opportunities in the JACK-holding converged binary process. `os.execute` keeps the exact lua contract. `io.popen` hands back a read-only handle (`read`/`lines`/`close`, see notes below). captured output is delivered through lua's `print`, so command output still lands in maiden.
- update - the SYSTEM>UPDATE menu guards against a failed or malformed `releases.txt` reply, which previously left the checking screen spinning forever.
- update - the SYSTEM>UPDATE menu shows the installer's current step on-screen as it runs, with a heartbeat that lets it tell a long quiet step from a stalled install.
- update - the new binary proves it launches using `norns --check` before the device reboots into it.
- update - the swap stages `.new` and keeps `.old`, so a failed check rolls back to the working install.
- update - a failed step writes status (`failed:<step>` is the format) used to present messaging on-screen, and returns to a clean pre-update state, rather than booting a broken build.
- update - `m.releases_url` override variable for local update testing (see _test running the update locally_ below).

_verification_

- tested on hardware end-to-end
    - download -> checksum -> verified swap -> reboot into the new version
    - failure paths (corrupt / checksum / missing-script / bad-binary -> correct `failed:<step>`, no reboot, on-screen failure)
    - an installer that dies without reporting
    - manual update flow (wget + tar + run `./update.sh`)
- new unit test coverage across the changes (c/c++ doctest, lua luaunit, and a bash-based dry-run test for update.sh).
- test-update.sh wired into a new `test-update` CI job.
- the test-update dry-run suite runs `update.sh`, aimed at a scratch tree through its `NORNS_UPDATE_ROOT` / `NORNS_UPDATE_SUDO` test-time overrides.

the rest of this pr is largely the cleanup that fell out of making this all possible:
- the system actions (shutdown, reset, update) are now a fixed allowlist at the C layer. each action name maps to a fixed command and a fixed transient systemd unit which is launched detached by the sidecar via `sudo -n systemd-run`. lua only ever supplies the action name, so no new shell execution capability is exposed to scripts.
- tidy: drops the dead `passdone` copy in `lua/core/menu/system.lua` (settings and wifi each own their own live copy).

_test running the update locally_

the SYSTEM>UPDATE menu page can now read its release manifest from an `m.releases_url` override variable, settable at runtime over the repl. otherwise it defaults to the monome `releases.txt` on github. with this override, the real download -> verify -> swap -> reboot path can be driven against a local http server for testing:

**1. serve a manifest + bundle.** `releases.txt` is lua that the menu `load()`s. it sets a global `releases`:

```lua
-- example releases.txt
releases = {}
releases.stable = {
  version = "999999", -- greater than the device's ~/version.txt so the menu shows it
  url = "http://<host>:8000/norns999999.tgz",
  sha = "http://<host>:8000/norns999999.sha256",
}
releases.beta = releases.stable -- the menu logic validates both channels
```

the bundle it points at is a release tarball which is a top dir named for the version, holding `update.sh`, the built `norns/` tree (`build/norns/norns` + `build/maiden-repl/maiden-repl`), `maiden/`, the `config/` systemd units, `version.txt`, and `changelog.txt`. `update/pack.sh <version>` tars that dir and writes the matching checksum:

```bash
update/pack.sh 999999    # tars ./999999/ -> norns999999.tgz + norns999999.sha256
```

the quickest valid bundle is a copy of an installed release tree with its version bumped. serve `releases.txt`, the `.tgz`, and the `.sha256` from one dir over the LAN — `python3 -m http.server 8000`.

**2. point the device at it** over the repl (any restart resets the override to upstream):

```lua
require("core/menu/update").releases_url = "http://<host>:8000/releases.txt"
```

note the menu still pings `github.com` (and wants 400M free on disk) before it reads the override, so the device needs working internet even though the release itself is served from the LAN.

**3. open SYSTEM>UPDATE and confirm.** the success bundle downloads, verifies, swaps, and reboots into the new version. for the fail-safe paths you don't even need a valid bundle — a corrupt `.tgz`, a wrong `.sha256`, or a bundle missing `update.sh` each surfaces its `failed:<step>` on screen with no reboot.

_notes_

- bug: lua `io.popen` is partially regressed... "w" mode intentionally fails with a loud error, and reads hand back the finished command's whole output rather than streaming it. this should be fixed in a follow-up PR. it requires additional changes to the sidecar to support it, and i'm intentionally not putting that work here, to keep this already massive amount of change as contained as possible at this point. the system cmd and sidecar work from this PR is already an excellent prefactoring for the changes required to get `io.popen` fully working.
- restart is now dependable as a mechanism, but jackd itself can still hang, across repeated restarts and around ssh sessions. this PR does not fix this issue. two potential causes found while working (ssh'ing and restarting **a lot**):
  - bug: `crone_cleanup()` → `jack_client_close` only runs when the event loop exits on its own. restart delivers SIGTERM, and with no handler installed the process dies before reaching it, so every restart is an unclean JACK client death. the fix _should_ be a SIGTERM handler that routes into the existing teardown path.
  - bug: closing the last ssh session makes systemd-logind purge the shared memory (`RemoveIPC=yes`), destroying jackd's shm while it runs. the next restart then fails until a power cycle. the fix here is likely a `RemoveIPC=no` logind addition. this belongs in norns-image for fresh flashes, but it should likely also ship through `update.sh`, so existing devices would pick it up on their next update without a reflash.
- async `norns.system_cmd` now has a ceiling upstream does not have. a command that stays silent for more than 4 minutes is killed and reported failed. the timeout bounds silence, not total runtime, so a command that keeps producing output can run as long as it needs. `os.execute` and `io.popen` pass no ceiling at all, keeping stdlib duration behavior for scripts that use them (this is on the script author).
- with this PR `update.sh` has indeed outgrown its bash beginnings at ~500 lines... with a rather silly (but kind of fun to write) hand-rolled test rig as an additional ~500 lines. this is fine, but the bash scripting seems to have reached its limit. it is fully functional and tested work, but further expansion should consider moving to python IMO.
- i'm really excited about this work. looking forward to review(s). thanks.

### DIFF
--- a/lua/core/menu/update.lua
+++ b/lua/core/menu/update.lua
@@ -1,136 +1,273 @@
 local m = {
   alt = false,
-  install = "stable"
+  install = "stable",
+  dots = 0,
+  -- polls with no new progress, cleared by progress and by each check
+  stall_ticks = 0,
+  status_path = "/tmp/norns-update-status",
+  progress_path = "/tmp/norns-update-progress",
+  -- named in system_cmd.cc
+  update_unit = "norns-update",
+  -- override, settable from the maiden repl for testing
+  releases_url = nil,
+  -- 0.5s ticks without installer progress before probing the unit (10 min)
+  install_timeout_ticks = 1200,
+  -- bumped on leaving the page, so a stale callback can be told from a live one
+  visit_id = 0,
+  constants = {
+    CHECKING = "checking",
+    MESSAGE = "message",
+    CONFIRM = "confirm",
+    UPDATE = "update",
+    CANCEL = "cancel",
+    INSTALLING = "installing",
+    INSTALL_FAILED = "install failed",
+  },
 }
 
 releases = {}
-local VER = { "stable", "beta" }
 
-local function checked() print("what??") end
-local function get_update_2() end
+local UPDATER = _path.home .. "/norns/update/update.sh"
+
+-- async callbacks, defined below their callers
+local compare_releases
+local handoff_update
+
+-- the menu timer is shared, so a callback from a previous visit must not stop it
+local function if_same_visit(fn)
+  local visit = m.visit_id
+  return function(...)
+    if m.visit_id ~= visit then return end
+    return fn(...)
+  end
+end
+
+-- the shared menu timer, cycling the dots until the page settles
+local function start_tick_timer()
+  _menu.timer.time = 0.5
+  _menu.timer.count = -1
+  _menu.timer.event = function()
+    m.dots = (m.dots % 3) + 1
+    _menu.redraw()
+  end
+  _menu.timer:start()
+end
 
 local function check_newest()
   print("checking for update")
   if m.alt then print("stable and beta") end
-  norns.system_cmd( [[curl -s \
-      https://raw.githubusercontent.com/monome/norns/main/releases.txt \
-      ]],
-      checked)
+  local url = m.releases_url or "https://raw.githubusercontent.com/monome/norns/main/releases.txt"
+  norns.system_cmd("curl -s " .. url, if_same_visit(compare_releases))
 end
 
-checked = function(result)
-  print(result)
-  load(result)()
+local function valid_release(r)
+  return type(r) == "table" and tonumber(r.version) ~= nil
+end
 
+compare_releases = function(result)
+  print(result)
+  releases = {}
+  local manifest = load(result or "")
+  -- an http error body arrives as an ordinary callback, so it must fail here
+  -- rather than error mid-callback
+  if not (manifest and pcall(manifest) and valid_release(releases.stable) and valid_release(releases.beta)) then
+    print("update: could not read the releases list")
+    _menu.timer:stop()
+    m.message = "check failed. try again."
+    m.stage = m.constants.MESSAGE
+    _menu.redraw()
+    return
+  end
+
+  _menu.timer:stop()
   if m.alt==false and tonumber(norns.version.update) >= tonumber(releases.stable.version) then
     m.message = "up to date."
+    m.stage = m.constants.MESSAGE
   else
-    m.stage = "confirm"
+    m.stage = m.constants.CONFIRM
   end
   _menu.redraw()
 end
 
 
 local function get_update()
-  m.message = "preparing..."
   _menu.redraw()
   pcall(cleanup) -- shut down script
   norns.script.clear()
   _menu.locked = true
   print("shutting down audio...")
-  --_norns.execute("sudo systemctl stop norns-jack.service") -- disable audio
   for i=1,6 do _norns.cut_enable(i,0) end -- disable softcut
   _norns.execute("sudo systemctl stop norns-sclang.service") -- disable audio
-  print("clearing old updates...")
-  _norns.execute("sudo rm -rf /home/we/update/*") -- clear old updates
-  m.message = "downloading..."
+  m.dots = 0
   _menu.redraw()
   print("starting download...")
-  local cmd = "wget -T 180 -q -P /home/we/update/ " .. releases[m.install].url .. " " .. releases[m.install].sha
+  local release = releases[m.install]
+  local cmd = table.concat({
+    "/bin/bash", UPDATER, "download", release.url, release.sha
+  }, " ")
   print("> "..cmd)
-  m.initial_disk = norns.disk
-  norns.system_cmd(cmd, get_update_2) --download
-  _menu.timer.time = 0.5
-  _menu.timer.count = -1
-  _menu.timer.event = function()
-    m.blink = m.blink == false
-    m.message = "downloading: " .. m.initial_disk-norns.disk
-    _menu.redraw()
-  end
-  _menu.timer:start()
+  norns.system_cmd(cmd, if_same_visit(handoff_update))
+  start_tick_timer()
 end
 
-get_update_2 = function()
+local function read_failed_step()
+  local f = io.open(m.status_path, "r")
+  if not f then return nil end
+  local line = f:read("l")
+  f:close()
+  return line and line:match("^failed:(.+)$") or nil
+end
+
+local function clear_failed_step()
+  os.remove(m.status_path)
+end
+
+local function enter_install_failed(step)
   _menu.timer:stop()
-  m.message = "unpacking update..."
+  m.failed_step = step
+  m.stage = m.constants.INSTALL_FAILED
   _menu.redraw()
-  print("checksum validation...")
-  m.message = "checksum validation..."
-  local checksum = util.os_capture("cd /home/we/update; sha256sum -c /home/we/update/*.sha256 | grep OK")
-  if checksum:match("OK") then
-    print("unpacking...")
-    _norns.execute("tar xzvf /home/we/update/*.tgz -C /home/we/update/")
-    m.message = "running update..."
-    _menu.redraw()
-    print("running update...")
-    _norns.execute("/home/we/update/"..releases[m.install].version.."/update.sh")
-    m.message = "done. any key to shut down."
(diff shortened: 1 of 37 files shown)


colinmcardell:
force pushed lua lint fixes to keep things tidy.

colinmcardell:
a few of the update screens which now stream each install stage as it runs and, on failure names the step that failed (including whether rollback worked if it had to rollback).
installing-2-verifying-checksum
installing-5-installing-packages
installing-7-updating-norns
failed-checksum
failed-binary-swap-rollback
installing-8-finishing

tehn:
wow!! thank you! i have a couple hectic days but review and testing coming shortly. looking forward to a rollout

lua/core/menu/update.lua
```diff
@@ -143,20 +280,32 @@ m.redraw = function()
...
+    screen.move(64,50)
+    screen.text_center("please do not power off")
+  elseif m.stage == m.constants.INSTALL_FAILED then
+    screen.move(64,30)
+    screen.text_center("update failed")
+    local pre, step = "error:", m.failed_step or "unknown"
+    local x = 64 - (screen.text_extents(pre) + 4 + screen.text_extents(step)) / 2
+    screen.level(10)
+    screen.move(x,40)
+    screen.text(pre)
```
tehn, line 299: likely easier to do string concatenation here ie
```
screen.text(pre .. " " .. step)
```
and then use text_center instead of pre-calculating width with `text_extents`.

there was some concern with `text_extents` some years ago, i need to refresh my memory on the problem.

norns/main.cpp
```diff
@@ -13,6 +13,13 @@
 #include "sidecar.h"
 
 int main(int argc, char **argv) {
+    // update.sh gates the final reboot on this. exiting 0 before any
+    // fork/jack/hardware init proves the new binary links
+    if (argc > 1 && strcmp(argv[1], "--check") == 0) {
```
tehn, line 18: curious if `--version` would be a micro feature that could double as this check

tehn (commented):
reviewed the code, looks good, skimmed the sidecar section.

will be able to test on hardware before the end of the week.

thank you for the thorough notes!

norns/main.cpp
colinmcardell: that's a good idea.

tehn:
functional test went well, tomorrow i'll proceed with testing the update system

looking good!

colinmcardell:
That's great! i'm starting to work on the bugs and regressions mentioned in the PR notes. It will be another round of work on the sidecar, given my later findings related to `os.execute` and `io.popen` forking. It should bring system commands in parity with pre-converged behavior with no forking.  🎉 

Hopefully we can have a few more beta releases to really put the update process through its paces!

tehn (approved):

tehn:
I'll get a new beta posted so we can continue on the remaining checklist items, thank you for the continued improvements!

tehn:
testing the process went well. i was (accidentally) able to test first a bad url and then a full disk, both worked wonderfully.

it's now up on beta!

note for anyone running an earlier 3.0 beta, you'll need to manually run the update via ssh:

```
cd update
wget https://github.com/monome/norns/releases/download/v3.0.2/norns260819.tgz
tar xzvf norns260819.tgz
cd 260819
./update.sh
```

xmacex:
I had an earlier beta of norns yesterday, things were fine, I went to sleep. Today, I did the manual install from above (sadly missing the new update procedure with more feedback, looks nice in the screenshots). Now my norns won't start until I disable the [*repl* mod](https://llllllll.co/t/repl/60157?u=xmacex) which is great mod by @p3r7 for live-coding, reading logs etc. Kind of like in-the-box maiden. Once I have disabled it, things seem fine again.

Why I am writing here in the thread is that I suspect it has to do with this sidecar thing, which I admit I don't understand what it is really. If I should I make an issue instead, just let me know.

Anyway, *repl* starts a process using os.execute, with a path to an executable as string argument with an ampersand suffix, and there is some io.popen-ing happening too. I followed with print statements until here: https://github.com/p3r7/repl/blob/main/lib%2Frepl_osc_gw.lua#L42-L47. Norns halts. I can see with ps the process has started. If I kill the process from the command line, norns resumes. Basically, if this is correct, the os.execute blocks even if the intention is that it forks and detaches. It did yesterday until the new beta some hours ago.

Repl is a niche mod (more people should know about it toot toot!) and I will be happy to try to migrate that to some new way to run subprocesses, if that is what this sidecar thing is.

There are however other scripts which might also fork processes and expect not to block. One small "family" of these are the USB audio Input/output things, which started with the Connect-OPZ I made, and further the autojack and Connecto. All of them (directly or indirectly) fork alsa_in using the ampersand from os.execute. My own Connect-OPZ seems to work fine, but I wonder if scripts which use os.execute with the ampersand will block norns or not, like the repl mod seems to.

Please let me know what I can help test.

And thank you for the good work on norns.

colinmcardell:
Hey! This is great feedback. Thanks. Yes, all mods are broken at the moment, and then secondary to that, mods that use os.execute or io.popen will be broken until they are well threaded through sidecar. This is a great list of mods to try because they seem to address both of the main things needing fixing. Thanks for the info. Stay tuned, io.popen, os.execute, and mods regressions are next on my list. 🙏

xmacex:
Super, just ping me if I can help testing something, otherwise I'll keep an eye for betas to try.

The [_tending_the_waves_ mod](https://github.com/markijzerman/tending_the_waves/), which is for listening and sampling and mangling and just enjoying Internet radio streams blocks norns once I choose a channel to listen to and the mod [os.executes to start mpv](https://github.com/markijzerman/tending_the_waves/blob/main/lib%2Fmod.lua#L74-L77). Norns starts up fine. Also I can listen to the stream just fine, norns is just inresponsive. Once I ssh in and kill mpv, norns seems to jolt back into action.

colinmcardell:
I will be pulling together the changes to knock out the remaining known issues for the converged binary work soon. I have most of it working (still need to e2e test the mods seam). I cannot put more time in this week, but will report back next week as things move along. 🍻

PR #1906, opened 2026-09-05
crone: tape reader: wait for pending stop in open(); don't reset state on no-op start() (fixes #1758)

### LINKED ISSUE #1758: tape playback unresponsive if preview is not stopped

jstefani:
Fixes #1758.

## Repro
In fileselect: preview a file with E3, then press K3 to select it **without** stopping the preview first. K3 press sends `/tape/play/stop`; K3 release sends `/tape/play/open`. The stop is still fading out (~50 ms) when `open()` arrives.

## Root cause (crone `Tape<N>::Reader`)
- `open()` saw `isRunning` and refused ("stream is running; no action was taken"). The disk thread then exited and cleared `hasFile`, so the following `start()` aborted with "no file pending". The selection is silently dropped.
- Worse: `Reader::start()` reset `isPrimed` / `transportState` *before* `SfStream::start()` checked `isRunning`. A `start()` while already running (which Lua issues right after a refused `open()`) starved the in-flight fade-out: `process()` bails on `!isPrimed`, the fade never completes, `shouldStop` is never set, and the disk thread waits on its CV forever. The reader is stuck until crone restarts.

This matches @catfact's note on the issue that the fix belongs in the tape state machine rather than in Lua debounce timers.

## Changes
- `SfStream::waitForDiskThreadExit()`: request a Stop if one isn't already in flight and block until the disk thread has exited. Lets the normal fade-out complete; forces the thread out after 150 ms if the audio thread isn't servicing the stream; gives up after the caller's timeout.
- `Reader::open()`: when running, wait (500 ms max) instead of refusing.
- `SfStream::start()` gains a `resetForStart()` hook, called under the lifecycle lock only after the running/`hasFile` checks pass. The Reader's per-run reset moves there, so a no-op `start()` no longer mutates state.
- `Reader::diskLoop()`: close a handle left open by the EOF "park" path before opening the next file, instead of leaking it.

No Lua changes. `Writer` behaviour is unchanged (it still refuses `open()` while recording, which seems intentional).

## Verification
Threaded harness (fake 128-frame audio thread + real wav, built against the homebrew jack/libsndfile headers) driving the reader with the issue's timing:

| scenario | before | after |
|---|---|---|
| preview | sound | sound |
| stop, open +100 ms, start +150 ms (the issue) | sound* | sound |
| stop/open/start back-to-back (fileselect scroll) | silent, reader stuck | sound |
| open while playing, no stop | silent, reader stuck | sound |
| stop, wait, open, start | silent (still stuck) | sound |
| start with no file | silent | silent |

\* passes in the harness only because a 100 ms gap happens to outrun the fade; a faster K3 press/release hits the refusal.

An equivalent fix against the pre-#1857 `Tape.h` was confirmed on hardware (norns shield) with the exact K2 → E3 preview → K3 → K2 sequence from the issue. This port to current `main` has been verified with the harness only; a hardware pass on `main` would be welcome.

### DIFF
--- a/crone/src/Tape.h
+++ b/crone/src/Tape.h
@@ -6,6 +6,7 @@
 #define CRONE_TAPE_H
 
 #include <atomic>
+#include <chrono>
 #include <condition_variable>
 #include <cstdint>
 #include <iostream>
@@ -161,6 +162,9 @@ class Tape {
             std::cout << "Tape::SfStream: starting..." << std::endl;
             shouldStop = false;
 
+            // derived-class state reset; only now that we know we're really starting
+            resetForStart();
+
             // bump command epoch so any pre-existing commands are ignored by consumer
             runEpoch.fetch_add(1, std::memory_order_acq_rel);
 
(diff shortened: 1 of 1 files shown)


jstefani:
Fable 5.1 generated

jstefani:
To clarify - I tested the same fix on hardware against the 2.9.4-era Tape.h - let me know if you need it tested against 3.x and I can likely do that.

PR #1908, opened 2026-09-06
fileselect: resolve preview/select paths from real filenames, not display labels

jstefani:
## Problem
In TAPE > PLAY, files with long names never preview on E3, but play fine once selected with K3.

`fileselect.getlist()` builds `display_list` by passing each name through `util.trim_string_to_width` (97 px for audio rows so the duration fits), so anything wider becomes `some-long-na...`. The E3 preview handler passed that truncated **label** to `tape_play_open`, so the path didn't exist and nothing played. Dashes aren't involved; it's purely name width.

The K3 handler had a workaround (if the label ends in `...`, fall back to `fs.list[pos]`), which is why selection worked. That workaround is itself wrong once a filter hides rows: `display_list` only holds visible entries while `fs.list` holds everything, so the indices drift and K3 selects a different file from the one highlighted.

## Fix
Keep a parallel `fs.names` table holding the real filename for every visible row, and resolve both preview and selection through it. Drop the `...` heuristic.

## Verification
Stubbed-norns lua test (fake 5 px/char text extents, four files incl. two long `.wav` names and a `notes.txt`):

| case | before | after |
|---|---|---|
| E3 preview long name | opens `/audio/a-really-long-fi...` | opens real path |
| K3 select long name | ok (heuristic) | ok |
| E3 preview, `.wav` filter hides `notes.txt` | truncated path | ok |
| K3 select, `.wav` filter hides `notes.txt` | selects **`notes.txt`** | ok |

Deployed to a norns running 2.9.4 (the file is identical there) for hands-on confirmation.

### DIFF
--- a/lua/lib/fileselect.lua
+++ b/lua/lib/fileselect.lua
@@ -17,6 +17,7 @@ function fs.enter(folder, callback, filter_string)
   fs.folders = {}
   fs.list = {}
   fs.display_list = {}
+  fs.names = {}
   fs.lengths = {}
   fs.pos = 0
   fs.depth = 0
@@ -104,6 +105,7 @@ fs.getlist = function()
   local dir = fs.getdir()
   fs.list = util.scandir(dir)
   fs.display_list = {}
+  fs.names = {}
   fs.lengths = {}
   fs.visible = {}
   fs.pos = 0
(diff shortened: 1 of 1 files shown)


jstefani:
This mod creates the long file names : https://norns.community/nice-tapes/

PR #1911, opened 2026-09-09
Build the dev container packages using Raspbian sources

Dewb:
Debian Bullseye is now EOL'd and an modified `/etc/apt/sources.list` is now required to update packages. There appear to be some issues with the transition to archive.debian.org, so switch to the Raspbian deb archive, which better matches the norns hardware image anyway.


Dewb:
Opened this from the pre-converged main branch by accident, leading to an ugly merge, will re-submit.

PR #1912, opened 2026-09-10
Dev container: use raspbian site for all packages

### LINKED ISSUE #1909: CI failing on apt-get 404s
Example:

https://github.com/monome/norns/actions/runs/34156533953/job/101849424300

One of these is probably the solution (substituting bullseye) 
- https://github.com/actions/runner-images/issues/12627#issuecomment-3213496270
- https://superuser.com/a/1830197

Dewb:
Debian Bullseye is now EOL'd and modifying `/etc/apt/sources.list` is now required to update packages. There appear to be some issues with the transition to archive.debian.org, so this commit switches the container's sources.list to the Raspbian deb archive, which better matches the norns hardware image anyway.

This should fix #1909.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,10 @@ ENV LANG=C.UTF-8 \
     SUPERCOLLIDER_PLUGINS_VERSION=3.13.0 \
     NNG_VERSION=1.11
 
+RUN \
+    rm /etc/apt/sources.list && \
+    echo 'deb [trusted=yes] http://raspbian.raspberrypi.org/raspbian/ bullseye main contrib non-free rpi' > /etc/apt/sources.list
+
 RUN apt-get update -yq && apt-get install -y \
     bc \
     build-essential \


Dewb:
@colinmcardell any idea why tests aren’t able to link the atomic family of functions? If the image change removed them (possibly by bringing in an older version of gcc?) the configure step should have flagged it, no? Where are CFLAGS set for tests?

colinmcardell:
Looks like the Raspbian build might default to `-march=armv6` instead of `-march=armv7-a`
This _should_ fix it...

https://github.com/monome/norns/blob/1d7209428841bc2b38619c8238ba0d2788bdbe68/tests/wscript#L736

becomes
```c
uselibs = ["LUA53", "NNG", "ATOMIC"]
```

---

Update. I ran tests locally with the above change successfully.

Dewb:
Excellent, thanks! Which config do on-device builds use, armv6 or v7a? Is there an easy way to make sure tests, the official build, and the container build are aligned on these choices? A completely separate waf script for tests makes me nervous.

colinmcardell:
Currently I believe the on device builds use armv6, however I have been taking a look at them and we should consider refining the arch and compile flags because there are gains to had. Out of scope for now.

I can take a look at deriving the tests build settings so that it's centralized as a follow up as well.

Dewb:
(Skipping review to unblock CI ASAP)

colinmcardell:
🚀

PR #1913, opened 2026-09-13
build: add -faligned-new to prevent over-aligned heap crashes in c++14

jstefani:
Found this while working on private mods to the tape (which I hope to share later). It does not crash stock norns but it is an issue that my mods were hitting. The flag fixes it and, I believe, is needed after PR #1857 added a 64-byte aligned ReaderWriterQueue to MixerClient.


colinmcardell:
Thanks for the PR. I looked through the mentioned code. Does seem like the correct flag to add so that it allocates appropriately. It would be nice to repro the issue you have run into. Do you have more details?

Also, it might be worth considering placing the addition in the root wscript C++ flags instead, so that all inheriting the root flags get this as well.

jstefani:
I was adding variable speed playback to the tape when I hit it. Crone started dying with SIGBUS in MixerClient's which led me to look at the updates in PR 1857. Adding faligned fix it. That's about as far as I went.

If you don't mind some AI though, when asked: "Why did stock never crash? Stock has the same wrong belief but never a store that depends on it. All the existing members are atomics, mutexes, pointers, ints, floats, written one at a time with ordinary 4-byte stores that don't care about 16-byte alignment. The new 32-byte zero fill was the first place where GCC had both the opportunity to vectorise and a reason to attach a :128 hint. A double on its own would probably have been fine too; it was the small array that did it."

Regarding the root wscript, I moved it there per your suggestion. Indeed better than where I had it. Thanks!

Dewb:
Adding this flag is a good idea. It would be even better to move the codebase to C++17 or C++20, which would make the flag unnecessary, but that's a larger project.

Dewb:
@jstefani per https://github.com/monome/norns/blob/main/CONTRIBUTING.md, if you use chatbot assistance you should be verifying its claims and writing PRs and comments in your own words. 

The queue was added to `Tape`, not `MixerClient`, and the description you quote above is not useful as it depends on context in your chat session and/or in your code that we do not see, and it does not fully describe a reproducible test case.

jstefani:
Hello. I am, to my shame, the reason the contributing doc was updated. I was thus very careful to write everything in my own words. I only added the additional description when asked for more. Even then I used my own words for the first part of the reply but realized it might help if I also gave the models output - my words would have been less helpful "I added two new member vars and it started crashing. I don't know why." because even with 35 years of writing c and c++ I didn't realize that that compiler/optimizer might do what it does in this case. So in this case the model was very helpful to explain why adding two member vars (but not yet accessing them) was crashing things. 

I love norns, want to contribute and am trying to follow your rules.

If anything the miswording about MixerClient is very much because I did use my own words. I wrote what I (mis)remembered (MixerClient contains a tape - the mixerclient that I was using was still getting the result of changes in 1857).

A few days ago I removed all my PR's per your rules. Today I purposefully picked this new issue as it was only a compiler flag. I told myself that this would a good PR that can be clearly stated, no example needed, just write it and contribute to the project I love.

Dewb:
Hey, no shame, we're all figuring this out as we go. I just wanted to call attention to the fact that you didn't answer colin's request for steps to reproduce the error.

>  my words would have been less helpful "I added two new member vars and it started crashing. I don't know why."

I disagree, I think this [alongside a description of the vars you added] would have been much more helpful. "This is what I did and what I observed" is always useful, and "I don't know why" is always okay.

jstefani:
Here's the diff to reproduce:

    --- a/crone/src/Tape.h
    +++ b/crone/src/Tape.h
    @@ class Tape {
             std::atomic<bool> isPrimed{};
             bool needsData{};
             std::atomic<bool> loopFile{};
    +        double phase = 0.0;
    +        Sample hist[4][NumChannels] = {};

Dewb (approved):

Dewb:
Can confirm the above diff results in a SIGBUS crash at startup in the MixerClient constructor, and the aligned-new flag fixes it.

 GDB output: 
```
Thread 1 "norns" received signal SIGBUS, Bus error.
0x00089ac0 in crone::MixerClient::MixerClient() ()
```
I was curious if we had any other reports that could be traceable to this issue, but I don't think we've seen this class of error before.

PR #1914, opened 2026-09-13
Ignore late resumes for cancelled clocks

subvertnormality:
Hey! A native clock wakeup can remain queued after `clock.cancel()` removes its coroutine. When it arrives, `clock.resume()` passes `nil` to `coroutine.resume()` and raises `thread expected`.

This ignores only missing IDs that were previously allocated; unknown IDs still error.

Tested on physical norns: stock produced 27 late-resume errors and the fix produced zero, while both expected MIDI messages still arrived

--

Tested on physical Linux norns 5.10.92

The process:

1. started a coroutine that scheduled a short sleep
2. allowed its native wakeup to become queued
3. cancelled its Lua coroutine
4. checked whether the queued wakeup later tried to resume it

Results:

- stock clock: 27 `coroutine.resume(nil)` / `thread expected` errors
- patched clock: 0 errors
- normal immediate and delayed MIDI messages arrived in both
- an unknown future clock ID continued to raise an error
- the original stock clock was restored byte-for-byte afterward


Dewb:
Possibly naive question: is there no way to unqueue the scheduled native wakeup at the same time we cancel the coroutine? @artfwo I assume that if this was easy we'd already be doing it?

artfwo:
> Possibly naive question: is there no way to unqueue the scheduled native wakeup at the same time we cancel the coroutine? @artfwo I assume that if this was easy we'd already be doing it?

yeah, we are doing it [here](https://github.com/monome/norns/blob/main/matron/src/clocks/clock_scheduler.cc#L215)

but i've never seen the problem before and not sure why this happens though. some unforeseen race condition is the only reason i can think of, and in this case the guard looks valid.

Dewb:
if I'm following the code correctly, the clock code has to manage `clock.threads`, a table of coro/thread IDs to coro objects on the Lua side, and `clock_scheduler_events`, an array of structs on the C side that associates coro/thread IDs with state information. I also see that the implementation of clock.resume has a "todo: use C API instead" [comment here](https://github.com/monome/norns/blob/7c3ce460078ddb63075e099e5cd666597294d0dc/lua/core/clock.lua#L59). Maybe the race condition could be eliminated by combining these data structures so they can't get out of sync?

subvertnormality:
I think this small defensive fix is better than leaving the crash unresolved, but I’m not familiar enough with the deeper architecture to judge whether the race should be addressed there. Happy to defer to the maintainers on the preferred implementation.

Dewb:
Agreed that these safety checks are a good and appropriate bandaid until a complete understanding of what's happening can be reached.

artfwo:
> if I'm following the code correctly, the clock code has to manage `clock.threads`, a table of coro/thread IDs to coro objects on the Lua side, and `clock_scheduler_events`, an array of structs on the C side that associates coro/thread IDs with state information. I also see that the implementation of clock.resume has a "todo: use C API instead" [comment here](https://github.com/monome/norns/blob/7c3ce460078ddb63075e099e5cd666597294d0dc/lua/core/clock.lua#L59). Maybe the race condition could be eliminated by combining these data structures so they can't get out of sync?

iirc, the TODO was left in hope `clock.resume` can be a thin wrapper for a weaver function, but i don't think having it in C matters in the context of this issue.

a plausible explanation for the issue could be that clock scheduler thread is pushing coroutine wakeup events through the queue (evq, which is another source of nondeterminism), and when a resume is already queued to be processed in weaver, cancelling it won't work, since it's no longer in the scheduler.

Issue #1917, opened 2026-09-15
automate build release artifacts


tehn:
long overdue, a likely good candidate for automation would be building release artifacts ie `norns260915.tgz` with matching `norns260915.sha256`

right now i manually build these on-device in a folder `~/newupdate`

which has the previous update inside. steps then:

- rename folder to release date
- `cd norns`
- `git pull`
- `./waf configure`
- `./waf --release`
- `cp norns/update/*.txt .`
- `cd ..`
- `./pack.sh 260915`

see https://github.com/monome/norns/blob/main/update/pack.sh

at this point i scp the artifacts and then manually upload them to a release page.

then i modify https://github.com/monome/norns/blob/main/releases.txt to point at a new version

---

this is of course error-prone but i've also become accustomed to the process.

theoretically if someone wanted to take on automating this, we could have an always-available newest-beta from `main`

one main reason i'm bringing this up is that i frequently (almost always) forget `--release` when compiling, which means a debug build is getting published, which will have less optimizations!

according to `norns/wscript` those optimizations might be very impactful:

```

    if bld.env.NORNS_DEBUG:
        norns_cxxflags += ['-g']

    if bld.env.NORNS_RELEASE:
        norns_cxxflags += [
            '-mcpu=cortex-a53',
            '-mtune=cortex-a53',
            '-mfpu=neon-fp-armv8',
            '-mfloat-abi=hard',
            '-funconstrained-commons',
            '-ffast-math',
            '-fgcse-sm',
        ]
```

Dewb:
Should the `cp norns/update/*.txt .` step be `cp update/* .`? I see the update mechanism on device expects update.sh to be in the root of the unpacked update.

Dewb:
<img width="1149" height="739" alt="Image" src="https://github.com/user-attachments/assets/14ae89e5-26af-4517-99e9-7aa1039865bb" />
Looking at the contents of the last release, it seems like there are some additional steps required to put together the contents of ~/newupdate, like getting a maiden build, and perhaps producing the (optional?) config/lib/package folders?

Dewb:
The build artifacts for this run (must be logged into GH) https://github.com/monome/norns/actions/runs/35173892142, when copied into ~/update, seem to be processed correctly into an updated running system. 

Currently the packaging process pulls from the norns-image repo all the updated service unit files from the last update in order to satisfy the expectations of the current update.sh; we need to figure out a plan how those release-specific system patches are going to work in a more automated setup.

[...]

PR #1919, opened 2026-09-16
Draft GH Action to build update package

Dewb:
First draft of #1917. This action will not run automatically, but can be triggered from the Actions page once it is merged into the main branch, with a string field to enter the desired release name/datecode (e.g. 260916.)

If I got the process outlined in #1917 right, the workflow should build the update zip and post it as an artifact for testing. Future additions could create and post the files to a GitHub Release, etc.

The unknowns that need to be tested:
- [ ] Will a build inside the Raspbian container running on an armv8 runner produce a binary that's usable on hardware? So far, so good; [this update package](https://github.com/monome/norns/actions/runs/35173892142) is installable, starts up, and runs scripts.
- [ ] Are there any other small deltas between the Docker build environment and the on-device environment that we need to correct?
- [ ] Need a plan for how config updates are going to be managed (see [comment here](https://github.com/monome/norns/issues/1917#issuecomment-5707578241))
- [ ] Need to either build maiden or pull a release from the maiden repo. Current draft workflow grabs a version from 2021, the last release with a build and not just source.


colinmcardell:
Thoughts on a `release` branch to trigger release builds, and have `main` build beta?

Dewb:
I was imagining three stages to this:
1. Produce builds as job artifacts, and test them on-device to confirm building via container is going to be viable.
2. Merge to main and enable the manually triggered release workflow, while keeping the triggering and updating of the version info json manual. Do at least a beta release and maybe a full release this way.
3. Work on completely automating the process end-to-end.

Things like working out a branch release strategy would happen in stage 3. In general I think the arguments for trunk-based development are strong, and I try to avoid release branches. In the past I've used a tag strategy, which works okay, but the GitHub UI for tagging is not great; if we dial in the workflow_dispatch arguments nicely, that *could* be a more user-friendly way to automate the entire process. The thing that release branches allow you to do is to release a hotfix after unreleated changes have been landed in the trunk (main). However, if you can release from any branch with the right tag/workflow arguments, then you can create a hotfix branch whenever you need one without having to create a branch for every release.

In the long run, it might be a good idea to have three advertised streams: "nightly" that tracks main for the brave and savvy, "beta" for tags that we want to test on users who are less technical but still game to test, and "release" for everyone. However, updates currently include scripts with side effects, so without working out a more rigorous strategy for making updates undoable/replayable, I'm not sure we want to encourage too much stream-switching.

Dewb:
A test package and sha256 are now available at [the bottom of the page here](https://github.com/monome/norns/actions/runs/35167604446) (must be logged into GitHub)!
Screenshot_2026-09-16_20-53-03

Dewb:
Good news, the compiled binary seems to run ok on device. Still some things to work out with meeting update.sh's expectations for the package structure (edit: done.)

Dewb:
I brought over all the config files from the norns-image repo, with the commits from both the main and norns-converged branches. The contents of these files match the configuration on-device with the latest beta update applied, with a couple of exceptions:

- the norns-image repo version of `/etc/systemd/journald.conf` sets `Storage=persistent`, while my device sets `Storage=volatile`. Stuck with the repo version for now.
- the norns-image repo contains files `99-realtime.conf` and `jackdrc` which are not referenced in any scripts or present anywhere on device, so I've omitted them.

Dewb:
latest package for testing is here: https://github.com/monome/norns/actions/runs/35815437422

tehn:
phenomenal!!

> * the norns-image repo version of `/etc/systemd/journald.conf` sets `Storage=persistent`, while my device sets `Storage=volatile`. Stuck with the repo version for now.

over the years we've tried various ways to eliminate logs filling up the disk, curious if the `volatile` was one of the attempts

also, i'm wondering if all the accidental debug builds (where i forgot `./waf --release`) emit a ton more debug info hence fill logs?

generally i've been very bad about keeping the `norns-image` repo updated. i really have been just copying the previous update tarball, adding to it, then re-balling. so, i am very grateful for this effort you've put into automating and cleaning up the mess i've made. thank you.

tehn:
reviewed and everything looks good, i'll let @colinmcardell hit go as this touches the update system

colinmcardell:
Amazing! This is great! I will have a look today.

colinmcardell:
I've had a read through. Looking very good! Lots to wrangle. Sent me on a couple of investigations.

1) There are several bits that we should adjust before merging in regarding the `update.sh` changes (and maybe a few other areas). If you don't mind, I will take a moment this weekend to bang on it locally on my hardware here, and post modifications in the comments that you can consider folding in.

2) Some of the imported configs from norns-image are superseded by changes introduced by way of `update.sh`. Likely want to update them. Huge positive of bringing these configs closer to the update process!

---

these are from the 260819 tarball:
### `etc/systemd/system/norns-main.service`
```
[Unit]
After=norns-jack.service norns-sclang.service
Requires=norns-jack.service

[Service]
Type=simple
User=we
Group=we
LimitRTPRIO=95
LimitMEMLOCK=infinity
Restart=always
RestartSec=500ms
ExecStart=/home/we/norns/build/ws-wrapper/ws-wrapper ws://*:5555 /home/we/norns/build/norns/norns

[Install]
WantedBy=norns.target
```

### `etc/systemd/system/norns-sclang.service`
```
[Unit]
After=norns-jack.service
Requires=norns-jack.service

[Service]
Type=simple
User=we
Group=we
LimitRTPRIO=95
LimitMEMLOCK=infinity
ExecStart=/home/we/norns/build/ws-wrapper/ws-wrapper ws://*:5556 /usr/local/bin/sclang -i maiden

[Install]
WantedBy=norns.target
```

---

I think we can remove `image/config/etc/rsyslog.conf` and `image/config/etc/rsyslog.d/rsyslog` at this point...`update.sh` runs `apt -y remove rsyslog`.

It looks like journald isn't actually reading `etc/systemd/journald.conf` which means it's defaulting to auto.

```
$ sudo dmesg | grep journald.conf
systemd-journald[110]: /etc/systemd/journald.conf:1: Assignment outside of section. Ignoring.
systemd-journald[110]: /etc/systemd/journald.conf:2: Assignment outside of section. Ignoring.
```

It has no `[Journal]` section header, so systemd ignores both lines. Also `SystemMaxUse` only caps the disk journal. The RAM one is capped by `RuntimeMaxUse` ([journald.conf(5)](https://manpages.debian.org/bullseye/systemd/journald.conf.5.en.html))

### `etc/systemd/journald.conf`
This works (verified on device):
```
[Journal]
Storage=volatile
RuntimeMaxUse=16M
```

This also makes the `rm -rf /var/log/journal` in `update.sh` redundant, but also harmless.

Dewb:
Great, sure, feedback this weekend would be excellent. 

The goal I had in my head for bringing in the config history was that the first update released through this process should produce the same device state if it was applied to a freshly re-imaged device and one that has had all updates since the last image release applied individually. Inferring that config state from the norns-image repo was convenient (and re-applying customizations made to the image has benefits as well) but what needs to ve dine to guarantee the above outcome was to look at the contents of the config folders in all the updates since the last image release, in case there was anything shipped that didn’t get captured in either the main or norns-converged branch of norns-image, like the two service files you cited. I’ll do a pass over the release history tonight.

